PR #90, opened 2014-11-13
joystick module fixed with proper event handling, added easyFly2 USB + Optic6 TX

dipspb:
Hello, thank you for MAVProxy.

Would you pls to merge my fixes:
1) Proper joystick events handling has been implemented
2) Added/tested support for my IPACS easyFly2 USB dongle with Hitec Optic 6 TX

Regards,
Dmitry Prokhorov

### DIFF
--- a/MAVProxy/modules/mavproxy_joystick.py
+++ b/MAVProxy/modules/mavproxy_joystick.py
@@ -71,6 +71,18 @@ joymap = {
      None,
      None],
 
+    # Support for Optic 6 with the IPACS easyFly2 Interface USB Adapter from:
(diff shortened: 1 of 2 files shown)


tridge:
Hi Dmitry, this seems to include some changes to unrelated files, such as accel cal. Is that deliberate?

tridge:
I've pushed the addition of the IPACS easyFly2, but not the other changes. The changes to event handling break rudder on my interlink controller, and it isn't clear to me what advantage they bring. It also isn't clear to me why we need the rcsetup changes

PR #98, opened 2015-01-08
fence: allow a bit more time for a fence point to arrive

mday299:
This should help a little on poor links.


tridge:
applied, thanks!

PR #99, opened 2015-01-10
Windows: Added mavinit.scr to installation and fixed build scripts

stephendade:
-Added mavinit.scr to Windows installtion
-Fixed build scripts for windows
-Fixed batch files to run MAVProxy direct from source in Windows


tridge:
applied, thanks!

PR #100, opened 2015-01-22
udp options: Add a --nocrc flag which turns off CRC on inbound UDP

geeksville:
Useful for low end CPUs to not spend so much time in mavproxy.

Needs associated changes in pymavlink


geeksville:
The mavnative stuff is going well (PR coming soon).  That gives a far bettter speedup because we don't use any python code for low level packet rx.  So no need for this half measure.

PR #102, opened 2015-02-04
mavnative: A set of independent changes to improve performance

geeksville:
These changes include independent commits to either improve performance (by supporting --native and --nolog and a few refactorings in mavproxy_link) or to make it easier to profile MAVProxy performance.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -176,6 +176,8 @@ class MPState(object):
         self.continue_mode = False
         self.aliases = {}
         self.max_rx_packets = None
+        self.logqueue = None
+        self.logqueue_raw = None
 
     def module(self, name):
         '''Find a public module (most modules are private)'''
@@ -796,6 +798,7 @@ if __name__ == '__main__':
                       default=1, help='MAVLink target master component')
     parser.add_option("--logfile", dest="logfile", help="MAVLink master logfile",
                       default='mav.tlog')
+    parser.add_option("--nolog", action='store_true', help="suppress log file writing")
     parser.add_option("-a", "--append-log", dest="append_log", help="Append to log files",
                       action='store_true', default=False)
     parser.add_option("--quadcopter", dest="quadcopter", help="use quadcopter controls",
@@ -895,7 +898,8 @@ if __name__ == '__main__':
           mpstate.module('link').link_add(serial_list[0].device)
 
     # log all packets from the master, for later replay
-    open_logs()
+    if not opts.nolog:
+        open_logs()
 
     # open any mavlink output ports
     for port in opts.output:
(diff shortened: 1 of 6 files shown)


geeksville:
Some numbers from Edison:

# Edison baseline
run 'old' non-optimized mavproxy
* 22% cpu usage - mavproxy.py --master=/dev/ttyMFD1,921600 --rtscts --daemon
* 9% cpu usage - add --streamrate=1 
* 9% cpu usage - add --streamrate=-1 same CPU usage (almost entirely due to idle processing overhead)

# Edison improved
new native version and other optimizations
* 10% with --master=/dev/ttyMFD1,921600 --daemon --native (vs 22% with old version)
* 5% with --master=/dev/ttyMFD1,921600 --daemon --native --nolog
* 4% by removing unused modules
* 2% by using --streamrate=1

tridge:
looks like this needs mavnative, so needs to wait for that to go in

MAVProxy/mavproxy.py
```diff
@@ -280,6 +304,8 @@ def unload_module(modname):
             if hasattr(m, 'unload'):
                 m.unload()
             mpstate.modules.remove((m,pm))
+            mpstate.modules_idle.remove((m,pm))
```
tridge: won't this raise ValueError if not in the list?

MAVProxy/mavproxy.py
```diff
@@ -697,7 +738,7 @@ def main_loop():
         for m in mpstate.mav_outputs:
             rin.append(m.fd)
         if rin == []:
-            time.sleep(0.0001)
+            time.sleep(0.1) # We have no ports, check back again in 100ms
```
tridge: we'll need to cope with joystick here which needs rapid idle calls

MAVProxy/mavproxy.py
mrpollo: would a silent failure be best in here, or should we explicitly check for existance before calling remove?

MAVProxy/mavproxy.py
mrpollo: is there a way to identify joystick so we can add a conditional or any other type of check?

mrpollo:
@tcr3dr I think we should do something about this PR, maybe re-submit a rebased version addressing @tridge concerns?, what do you think?

tcr3dr:
@mrpollo I can definitely handle the improvements to polling / event loops. Let me catch up on all the changes made in this PR over the next day or two and I'll update with a plan.

amilcarlucas:
Will this be picked up, or should it be closed ?

khancyr:
it is really old. 
It looks like have some part here https://github.com/ArduPilot/MAVProxy/pull/556

tridge:
closing as too much has changed, sorry!

PR #103, opened 2015-02-08
Joystick: Added support for Hobbyking AeroFly FMS

stephendade:


tridge:
looks like Grant already did this one!

PR #104, opened 2015-02-12
modules: added a new joystick

gmorph:
Added support for this joystick simulator cable
http://www.hobbyking.com/hobbyking/store/__13597__USB_Simulator_Cable_XTR_AeroFly_FMS.html
Support had already been added but it looks like the hardware has changed internally thus requiring this change.


tridge:
applied, thanks!

PR #106, opened 2015-02-18
Module Handling: Ensure the user gets a useful debug trace if loading fails

stephendade:
If a not all dependencies is not installed (such as python-opencv), module loading will fail without a detailed error message. This patch forces MAVProxy to print out the debug trace of why module loading failed, so the user can quickly see if it's because of a dependency. Thanks for @jamiereid and @devdsp for figuring out this issue.


stephendade:
Made a better patch in #108

PR #109, opened 2015-02-23
misseditor: It was breaking when trying to delete rows from an empty table

gmorph:
Added a check to ensure there are rows in the table before we try to delete them.


tridge:
applied, thanks!

PR #117, opened 2015-04-03
Update documentation

stephendade:
General update to documentation, based on current MAVProxy.
Also includes a short tutorial on developing MAVProxy modules.
Note I've included a Windows setup binary in here, with a link in the install instructions.


tridge:
applied, thanks!

PR #118, opened 2015-04-07
Optionally visualize loiter radii as cirlces on the map.

mday299:
Added a map setting "loitercircle" that will show all loiter waypoints in the mission as circles with radius WP_LOITER_RAD rather than points.

Depends on:
https://github.com/mavlink/mavlink/pull/365


tridge:
applied, thanks!

PR #120, opened 2015-04-07
misseditor: Prevent mission editor lockup on missing table input.

mday299:
Little bug fix to prevent the Mission Editor from locking up when somebody forgets to populate a cell.


tridge:
applied, thanks!

PR #122, opened 2015-04-09
misseditor: improved input validation. Prevent lockup due to scoping.

mday299:
Improved input validation.  I discovered that if an early variable in the try block caused the exception then all subsequent variable would never be declared.  This caused an error at the self..event_queue.put call.

This patch ensure every variable gets created and initialized by breaking up the try/excpet block.


tridge:
applied, thanks!

PR #123, opened 2015-04-10
map: Bug fix.  Loiter circles now clear when mission is cleared/changed.

mday299:
Loiter circles did not disappear when the map was clear or when the mission was changed.

This was due to an error in how the circles were being added to a layer on the map.


tridge:
applied, thanks!

PR #127, opened 2015-04-15
command: make watch command bidirectional

arthurbenemann:
So something like the prompt bellow is posssible
```
STABILIZE> watch *PARAM*
STABILIZE> Watching *PARAM*
STABILIZE> param set SYSID_THISMAV 1
>PARAM_SET {target_system : 1, target_component : 0, param_id : SYSID_THISMAV, param_value : 1.0, param_type : 9}
<PARAM_VALUE {param_id : SYSID_THISMAV, param_value : 1.0, param_type : 4, param_count : 487, param_index : 65535}
```

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -521,6 +521,9 @@ def process_mavlink(slave):
         return
     if mpstate.settings.mavfwd and not mpstate.status.setup_mode:
         for m in msgs:
+            if mpstate.status.watch is not None:
+                if fnmatch.fnmatch(m.get_type().upper(), mpstate.status.watch.upper()):
+                    mpstate.console.writeln('> '+ str(m))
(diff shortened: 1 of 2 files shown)


arthurbenemann:
The only problem is that it doesn't output the messages that are coming in from another mavlink connection (added with --out), which was what I set out to do in the first place.

arthurbenemann:
Commit https://github.com/arthurbenemann/MAVProxy/commit/ef116b5d65206eaa577d35c9cead24d3d0e3f0d1 fixed the problem I describe at https://github.com/tridge/MAVProxy/pull/127#issuecomment-93131557

This PR is now finished. Although a better implementation of output ports would be nice, like bundling them with the link manager module.

mrpollo:
:+1: this is great for debugging

tridge:
applied, thanks!

PR #130, opened 2015-04-19
Windows: Extra hidden modules included in the build script

stephendade:
Added hidden libraries to the build script so that joystick module (among others) now works.
Also raised the installer script version to keep in sync with setup.py


tridge:
applied, thanks!

PR #132, opened 2015-04-22
cmdlong: add velocity command

rmackay9:
This adds support to send a set-position-target-local-ned command to copter to request the vehicle to move (normally in guided mode) at the specified earth-frame velocity.

I intent to use this command from within the copter SITL test so that we can protect against the velocity controller being broken (this happened between AC3.2.1 and AC3.3 although the issue is now resolved)

By the way, if there is a way to create the set-position-target-local-ned command from within the arducopter.py script in SITL we could drop this PR.  I simply couldn't find a way to do it though.

I also realise that this command doesn't fit perfectly within the cmdlong module because it's technically not a command-long but it seemed like a bit of a waste to create a whole new module just for this one command.


tridge:
applied, thanks!

PR #133, opened 2015-04-22
default to local udp host if there is no 'master' arg or hardware connected

arthurbenemann:
This is useful if you have a 3DR Solo connected to your wifi network, and no serial ports connected.

I don't see any downsides.


tridge:
applied, thanks!

PR #134, opened 2015-04-24
DGPS: Fix for receiving UDP injection packets

njoubert:
This is a bugfix for listening to DGPS packets on UDP, and sending a gps_inject_data mavlink message to the quad.


tridge:
applied, thanks!

PR #138, opened 2015-05-05
Fixed issue when using guided with given coordinates

Yukinoshita-Yukino:
Fixed cannot find "latlon" issue when using guided with given latitude, longitude and altitude.


tridge:
applied, thanks!

PR #145, opened 2015-05-17
WXConsole: Fixed disappearing-menu problem when loading some modules

stephendade:
When modules-containing-menus are loaded, the wxconsole menu will disappear due to reassigning the self.menu object whilst it's still being used. Solution is to set the set the menu to "None" whilst updating the self.menu


tridge:
thanks Steve, applied!

PR #146, opened 2015-05-18
Add versioning information

stephendade:
Added help module. It also provides version info and checks for updates (both Linux and Windows builds).
I needed to make a few small changes to the MPMenuChildMessageDialog to stop it from instantly showing on creation. This in turn led to a minor change to the map module to account for the aforementioned MPMenuChildMessageDialog changes.


tridge:
applied, thanks!

PR #151, opened 2015-05-29
Find and replace references to Tridge's repo

stephendade:
Update to the Dronecode repo


mrpollo:
does the domain need really have to be in caps? eg **Dronecode.github.io** --> **dronecode.github.io**

stephendade:
I'm just following the naming convention that the repository uses :)

tridge:
applied, thanks!

PR #152, opened 2015-05-29
mavproxy.py: return False from module load ImportError handler

peterbarker:
If you tried to import a module which had also had an import error
in it, the MAVProxy console became unusable.


peterbarker:
This broke functionality.  More thinking needed.

PR #154, opened 2015-06-03
speech: stop talking when speech unloaded

peterbarker:
Also avoid appending the speech setting multiple times
as _keys grows


tridge:
applied, thanks!

Issue #141, opened 2015-05-06
GUI modules with wx are broken on OSX


mrpollo:
Trying to use MAVProxy with either console or map kills those module Threads and keeps MAVProxy is running.

Installed wx from homebrew `brew install wxmac` I can confirm wx is working.

Python Stack:

```
☁  flight_replay [tmp_flight_replay66] ⚡ mavproxy.py --master=127.0.0.1:14550 --console
Connect 127.0.0.1:14550 source_system=255
Logging to mav.tlog
Running script /Users/rroche/.mavinit.scr
-> module load droneapi.module.api
DroneAPI loaded
Loaded module droneapi.module.api
-> set moddebug 2
/usr/local/lib/python2.7/site-packages/MAVProxy/modules/lib/wxconsole.py:53: wxPyDeprecationWarning: Using deprecated class PySimpleApp.
  app = wx.PySimpleApp()
Loaded module console
MAV> Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/site-packages/MAVProxy/modules/lib/wxconsole.py", line 62, in watch_thread
    msg = self.parent_pipe_recv.recv()
EOFError
```

System Stack Trace: https://gist.github.com/mrpollo/7501b3a0a277498e98b0

What other details can I provide to help track down the root of this problems, I'm not sure where to look for a fix here, I can provide a PR if someone can help out understand the depth of the problem.

amilcarlucas:
Can you retry with master?

amilcarlucas:
Works in the latest release. Closing it now. Please reopen if the problem persists in the latest release.

PR #158, opened 2015-06-15
Implements process-safe loading for wx.

tcr3dr:
The wx Python module is not process-safe on OS X, which causes the loading API to abort whenever a wx-based widget to be loaded. This patch enables, at minimum, the `--map` and `--console` options to be used on OS X, as well as a simple solution that will prevent regressions when developing on other platforms.

This solution is straightforward, though it involves a lot of small changes:

* `wx` is never directly loaded with `import wx`. Instead it is loaded through a helper: `from wx_loader import wx`.
* On the main thread, importing `wx_loader` throws an error (actually only on OS X, see below). The line `import wx_processguard` before `wx_loader` disables the error; this is used explicitly whenever we've forked a subprocess and want to load `wx`.

The bulk of the changes require modules which extend `wx` classes to be split from the main module object and a separate `_ui.py` companion file. You'll see that this the bulk of the changes in this PR: deleted code from files split modules into separate main, `_ui.py`, and `_util.py` files.

Because Mavproxy is a largeish codebase, we can err on the side of caution and have `wx_util.py` file only throw errors on OS X when we load `wx` on the main thread. Ideally, this causes 0 regressions on other platforms, and allows any unfixed features on OS X to be corrected over time.

Because this PR is a bit monolithic, I think it can also be split into three commits:

* Refactoring of the `_ui.py` files into separate files.
* Implementation of `wx_loader`.
* Correction of smaller changes like `wx.SimplePyApp()` => `wx.App(False)`

That would take a chunk of additional work, but whatever it takes to get these changes upstreamed.


mrpollo:
@tridge I believe I had emailed you about this before, I passed this task to Tim (@tcr3dr) togheter with your notes and concerns, we are hoping we can get this merged upstream at some point

mrpollo:
fixes #141

stephendade:
Nice! We've been trying to figure out how to get the GUI elements running on OXS for some time now :)

mrpollo:
whats the process to get this merged?

tridge:
my apologies for not looking at this earlier, I haven't been looking at PRs for mavproxy for a while. I'll take a look

tridge:
I think you may have forgotten to add some files in the PR? I don't see a wx_loader.py for example

tcr3dr:
@tridge Files now added. I had a snafu with gitignore, how bizarre.

Thanks for looking into this!

tcauchois:
Fixed the graph module as well in https://github.com/tcauchois/MAVProxy/commit/d9189636a9e3eaacc45e13ef6c49b34402d9b75f ... once this is merged I'll submit a pull request.

mrpollo:
@tcr3dr maybe you want to pull @tcauchois changes into your branch so we get this merged faster

tcauchois:
#163 incorporates both commits.

tridge:
i'll close this one and comment in #163, which has the graph fixes

tcr3dr:
Sounds good!

PR #163, opened 2015-07-09
Implements process-safe loading for wx [enhanced]

tcauchois:
See https://github.com/Dronecode/MAVProxy/pull/158 for details.  I extended the fix to the graph module.


tridge:
Tom and Tim, looks like this is pretty close! A few things still broken:
 - "module load misseditor" is broken
 - settings menus are broken (try MAVProxy->Settings in console module)
 - on map right-click "Show Position" is broken
I'm also puzzled how this change actually works. I don't have an OSX box, but when I've previously looked into this I thought that OSX couldn't open a window from a forked process. If this is working on OSX then it must be something quite different. Any clues? A lot of the re-arrangement of code in these two patches doesn't look like it actually does anything, so I must be misunderstanding it. Can one of you enlighten me?

tcauchois:
I'll have a pull request with mentioned fixes up shortly; there was an OSX bug in multiprocessing.queue.  As for how this is working, I don't have much to say other than importing wx globally doesn't seem to work and importing it in child_task does, presumably because that's when module __init__ is called...  And the refactoring is so that one doesn't need to import wx before getting to the child_task method.

tcauchois:
https://github.com/tcauchois/MAVProxy/commit/5671b435fd5287c0210e08d6fa3a35369bdba219

This fixes everything I can find except that when you use the menu bar in console to open up the map, the process with the map segfaults.  On the bright side, this validates the containment strategy as the rest of mavproxy works fine!  I think this might have something to do with us calling process_stdin from the gui watcher thread, instead of from the main thread?  Anyway, bad news.

Unrelatedly to all the GUI stuff, but probably relatedly to OSX, when I start MAVProxy too early in the boot sequence of my pixhawk, I get the following:

  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/serial/serialposix.py", line 485, in read
    raise SerialException('read failed: %s' % (e,))
SerialException: read failed: [Errno 6] Device not configured

Which I *think* is ENXIO coming from a read() call.  Weirdly enough, if we had forgotten to call open() we'd get a portNotOpen exception instead.

I'll maybe dig around those two issues tomorrow, but in the meantime all the gui bits seem to work.

tridge:
hi Tom, it is looking good now on Linux.

On windows there are still some issues. I've fixed the easy ones in this branch:
https://github.com/tridge/MAVProxy/commits/tcauchois-tcr-fixwx

AttributeError: 'mpQueue' object has no attribute 'qsizeCounter'
  File "build\bdist.win32\egg\MAVProxy\modules\mavproxy_map\mp_slipmap_ui.py", line
  File "build\bdist.win32\egg\MAVProxy\modules\lib\bugfix_mp.py", line 39, in get
AttributeError: 'mpQueue' object has no attribute 'qsizeCounter'

this is weird, as mpQueue clearly does have a qsizeCounter. I don't have time right now to poke into it, but will try to get back into it soon.

tcauchois:
The AttributeError makes no sense to me...  It's safe to only use mpQueue on OSX, though, since multiprocessing.Queue works fine elsewhere.  I was thinking of adding a function to create either mpQueue or multiprocessing.Queue, just to save that tiny bit of overhead on linux/windows.

tcr3dr:
I'll test and verify this branch works on my OS X box today.

@tridge I've come to believe OS X doesn't have GUI restrictions, just stricter memory access. It seems that wxPython \*module\* is not process-safe. By never calling `import wx` on the main process, `wx` is thus never "shared" by Python processes. Most code shuffling is splitting GUI and logic code that coexisted in the same class.

Additionally, this setup is at risk if a child process spawns its own process using wx, but that's avoidable.

tcauchois:
2a2e421 gets rid of my workaround queue on non-darwin platforms, and should fix the AttributeError.

I'm getting a segfault every time I try to load the map from the app now, though the map unit test works fine.  Everything non-map seems to work.  The segfault is a bit hard to debug as it occurs deep in pango init, and I'm having a bit of trouble getting gdb attached...

tridge:
with a few more minor fixes it is now working fine on windows, so I've pushed to master
many thanks to you and Tim for your work on this, much appreciated!

davidbuzz:
probably worth doing a release to pypi, for pip, as the version on there is from 2015-06-24 and doesn't work on OSX.

PR #166, opened 2015-07-16
map: map unloading fixes

peterbarker:
Closing the map window unloads the map module
Unloading the map module closes the map window


tridge:
applied, thanks!

PR #168, opened 2015-07-21
add cmd_mission_set_current

squilter:


tridge:
this seems to be the same as the existing "wp set" command

PR #171, opened 2015-07-28
Replaces "darkgreen" with "dark green" for WX

tcr3dr:
WX requires the string "dark green" (two words), or it outputs this error:

```
09:37:22 AM: Debug: wxColour::Set - couldn't set to colour string 'darkgreen'
```


tridge:
applied, thanks Tim!

PR #173, opened 2015-08-04
speech: Allow female speech synthesis. (espeak only)

mday299:
When switching back to the male voice at runtime, I couldn't figure out how to get exactly the same voice as the default voice that runs when set_voice() is never called.  I ended up choosing 'male3' which is slightly deeper and (I think) crisper and easier to understand than the default male -- of course that's only a personal opinion.

If I specify an empty string to the set_voice() method I can get the default voice back, but then it garbles words.


tridge:
I think it would be better to make the parameter be 'voice' and a string, then users can select any supported voice
John Cleese might be fun :-)

amilcarlucas:
ping @mday299

peterbarker:
We have a parameter for setting the voice now.

PR #175, opened 2015-08-10
Force python2

squilter:
`/usr/bin/python` is just an alias these days to either `python2` or `python3` depending on the OS.  Certain OS's (Arch especially) have it aliased to `python3`, and so programs starting with `#!/usr/bin/python` will run with the wrong version.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 '''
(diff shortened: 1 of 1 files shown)


tridge:
we better check if this works on MacOS and windows first
even better if we made mavproxy work with python3 .... I wonder how close the phoenix project is to working?

peterbarker:
We're getting Python3 support now, with tridge's latest PR.  That means it should work regardless of the shebang line.

PR #176, opened 2015-08-11
add live spectrogram module

jschall:


jschall:
figure_1

tridge:
I must be misunderstanding something about this. It looks like it is using the DEBUG msg, which sends one 4 byte float per msg. I can't understand how you can get any sort of reasonable spectrogram from that.
Can you explain what I've missed?

peterbarker:
@tridge This PR does allow you to source data from any message.  Could easily be an expression instead, I think.

@jschall I've had a play with this.  My understanding, however, is that you can't reasonably FFT unless your sample frequency is double the frequencies you are interested in.  Given for things like vibration we're interested in hundreds of Hz and that mavlink messages are maybe 10s-of-Hz, I can't see this being useful in that case.  Could you outline some use cases for this module, please?

jschall:
@peterbarker when I was using this I was sending the DEBUG message at 100hz. The vibration I was interested in at the time was down in the 10s of hz. It is likely that this will never be useful again and can probably be closed.

PR #177, opened 2015-08-11
modules: add position command

rmackay9:
This adds support to send positions using the SET_POSITION_TARGET_LOCAL_NED mavlink command.  Technically this is not a command-long but it's nearly identical to the velocity command which is in the mavproxy_cmdlong.py file.


tridge:
applied, thanks!

Issue #183, opened 2015-08-17
Add search support to the docs?


hamishwillee:
Hi @stephendade 

Might be worth moving MAVProxy to use an SSG that supports search and ideally sidebar navigation. The flat site structure is not particularly useful for navigation and it is much easier to work with markdown or reStructured text than html.

I recommend Sphinx or Jekyll. Happy to do the migration work if you want this.

Regards
H

stephendade:
Sounds like a good idea!
I'm not particularly well versed in setting up Sphinx or Jekyll environments, so I'll take you up on your offer of help.

hamishwillee:
Great. Do you have a strong preference for mardown or reStructured text for your inputs? If not, I recommend Sphinx, which uses restructured text, and probably the read the docs theme so it will looks something like (http://nailgun.readthedocs.org/en/latest/)

If you much prefer markdown then we'd probably look at Jekyll - maybe even this default theme http://jekyllrb.com/docs/home/

stephendade:
Sphinx looks good!

hamishwillee:
@stephendade I'm on it. Do you need the url filenames to be unbroken? I can keep them the same, but personally I think it would be nicer to add a directory structure, which will be definition mean that the urls change. Thoughts?

hamishwillee:
Hi @stephendade 
Checkout this first [build on my dropbox](https://dl.dropboxusercontent.com/u/3067678/share2/mavproxy/index.html). I split out the "Downloading" and installation from the front page into its own topic and put the license text in its own page linked from sidebar too.

I also put the text you used to have on every page in the front page "acknowledgements" section "MAVProxy is maintained by Andrew Tridgell and documentation is by Stephen Dade"

This is the "out of the box" read the docs theme. If you want some modifications to look and feel all that can be done "easily" is hiding that link to read the docs theme at the bottom and you can also make the sidebar "sticky" (though I don't plan to)/.If you want to modify this more significantly that is possible - but we'd need to modify the theme itself. Easy enough to do, but you'll need to provide direction.

Have a play, let me know what you think. I will then submit the sources for you to play with yourself and make live.

Cheerio
H

stephendade:
That looks really good @hamishwillee! A directory structure makes sense, as I expect the documentation to grow over time as more of MAVProxy is documented (and more modules created).

Also - I've fixed up the old documentation site at http://tridge.github.io/MAVProxy to redirect to the new site. So we only need to keep the documentation updated here.

[...]

PR #184, opened 2015-08-20
Sphinx docs

hamishwillee:
This is the docs and tools to build them - as per #183.


stephendade:
The old gh-pages branch and the master branch have been merged into here - they will need to be removed.
Compilation works nicely under Linux, but the files need to be output to the root folder so the main address (http://dronecode.github.io/MAVProxy/) still works.

hamishwillee:
@stephendade Yes, the old stuff will have to be removed. 

>Compilation works nicely under Linux, but the files need to be output to the root folder so the main address (http://dronecode.github.io/MAVProxy/) still works.

The docs are built to **\MAVProxy\docs\build\html\** but there is no dependency on this chain. You can just copy the content below that point into the root and everything should work. Does that make sense?

Do I need to do anything else?

stephendade:
@hamishwillee - it'll make things simpler from a git perspective if you cleaned up your branch before I merge it into the main tree

hamishwillee:
Sorry? How? The branch is just source - completely clean.

stephendade:
Use <code>git rm file</code> to delete files from your branch.
http://git-scm.com/docs/git-rm

All the previous html pages from the gh-pages branch are still in your branch

hamishwillee:
Hi @stephendade 

Done. Sorry about that.

Cheers
H

stephendade:
Looks good!   Thanks for doing this @hamishwillee

stephendade:
Ok, that didn't work. Github merged it into master rather than gh-pages.

stephendade:
Ahh, I think I see what the issue it:
@hamishwillee - Looking at the commit history on your sphinx_docs branch, you seem to have merged the master and your own changes together. Thus Github seems to think that this should be merged into master rather than gh-pages.

Could you create a new branch, based off dronecode/gh-pages and put your changes into there. Then it should cleanly merge into dronecode/gh-pages

stephendade:
Use <code>git checkout -b branch_name dronecode/gh-pages</code> to create a new branch based on the gh-pages branch in the dronecode repo

hamishwillee:
@stephendade My branch is supposed to merge with master - it adds the **source**. The plan would be to build the source and dump than on **gh-pages**. I thought you'd like to do that - but above it seems  you would like me to. That correct?

stephendade:
Ahh, turns out we were thinking of different things! I'll talk to @tridge and check he's OK with putting this into master.

hamishwillee:
Typically docs sources go in master and build go into gh-pages (in the other projects I've worked on). But yes - please check with tridge. 

I've created #186 which I think does the HTML modifications to gh-pages. This just deletes everything that was there already and then adds (in the root) the full output of my build. I hope that is correct/what is needed. If so, that should be able to go out.

PR #188, opened 2015-08-25
Add .nojekyll file to ensure relative paths don't get damaged

hamishwillee:
@stephendade The MAVProxy docs are broken because github runs some jekyll parser on the output. This patch fixes the problem (just adds a .nojekyll file)


stephendade:
Thanks!

Issue #191, opened 2015-09-02
Docs - add link to where to raise issues with documentation into theme


hamishwillee:
I can do this at some point. Adding as a placeholder.

PR #192, opened 2015-09-03
Adds "Page Bug" link to bottom right of footer

hamishwillee:
@stephendade This addresses #191. Basically adds a link which makes it easy to post bugs in github about a particular page (pre-seeds some information).


stephendade:
Thanks @hamishwillee !

PR #214, opened 2015-10-18
Logging: Added handling for daemon mode to log to /var/log. Also chec…

stephendade:
A couple of logfile fixes:
-Exception handling for initial log file open
-Check for enough (>200Mb) free disk space for logfile
-When in daemon mode, logs are written to /var/log


tridge:
applied, thanks!

PR #223, opened 2015-11-10
dataflash_logger: correct for new packet shapes

peterbarker:
dataflash_logger: drop packets not for me


tridge:
applied, thanks!

PR #225, opened 2015-11-14
cmdlong: add an actual "long" command to execute a MAVLink long command

peterbarker:
e.g.:
	long NAV_LAND
	long DO_PARACHUTE
	long 37


tridge:
applied, thanks!

PR #228, opened 2015-11-27
fix lat, lon of updated ADSB message for map module

SamuelDudley:
Display of simulated ADSB enabled aircraft was broken with mavlink/mavlink@7df9c75a19eab6292b580f3d26c07f6f8f92c196
This pull request fixes that issue.


tridge:
thanks, I'd actually fixed it myself, but forgot to push, pushed now
sorry!

PR #233, opened 2015-11-27
Console: display number of photos taken

AndKe:


tridge:
I think we should create that field in the console only when the first CAMERA_FEEDBACK comes in. If you are not doing a photo mission having "Photos --" there the whole flight doesn't really help

AndKe:
my methodology was to show features as soon as possible, reducing the chance that an problem there would be ignored because nothing is showing when no picture is taken.
But I do see some other features are like that, so I kind of agree with you too. :)

tridge:
I think this can be done same was as for XTrackError, with:
 console status add Photos CAMERA_FEEDBACK.msg_idx 4
(once we implement that!)

AndKe:
yes, another idea would be to have optional parameters after the "console status add" , options to that could hold threshold values for colouring values that needed attention.

PR #234, opened 2015-11-27
Console: display xtrack error

AndKe:


tridge:
I think it makes row 3 a bit wide by default. Maybe we should add another row for this and alterror and arspderror?

AndKe:
please see the attached screenshot from working console  - row 3 is still much shorter than the first row usually is.  - (please confirm that you think I should move them.)
screenshot from 2016-01-03 15-42-30

AndKe:
Collected AltError , ArspdError and XtrackError in row 4

magicrub:
Why do you want to show xtrack error? Are you having large errors that
you're wanting to monitor?

AndKe:
@magicrub nope - no unusual errors,  it's useful when flying bigger BLOS missions, I consider it a clear warning if there are too strong crosswinds in the mountains or if anything goes wrong. Telemetry is often very sparse on those missions (or with high latency like Iridium)

tridge:
I wonder if we should try to solve the general problem? What if we added a new command like this:
 console status add XTrackError NAV_CONTROLLER_OUTPUT.xtrack_error 4
we'd then implement the default console output by having a table of entries like that, and all of the hard coded cases would go away. The variable to display would be an arbitrary expression, much like you get in graph

AndKe:
sounds good to me :)

tridge:
maybe better to use a format string:
console status add "XTrackError: %.1f" NAV_CONTROLLER_OUTPUT.xtrack_error 4

tridge:
do you have any time to implement this? If not then I will do it, but not immediately

AndKe:
sorry, I have to use any programming time I have to make the Iridium_mavlink work

PR #241, opened 2015-12-25
param: add printing fields to help

justineaster:
This allows displaying info such as Range, Increment, and Units when "param help PARAM" is used.


tridge:
applied, thanks!
(and very useful too....)

PR #243, opened 2016-01-02
Added Geotag log to SmartCamera Module, Added support for Sony A7R and others

jmachuca77:
Added a log file to the smart camera module. Currently it is located in /sdcard/log, if the path does not exist the module currently fails to load. 
Added support for more sony cameras that failed to work before due to a missing command startRecMode. 
Added Initial setting of parameters to Shutter Priority Mode, ISO Auto, Shutter Speed 1/1600. These values can be modified by sending the appropriate commands after the module is loaded, or changing the default values in the python script.

PR #246, opened 2016-01-05
Added Geotag log to SmartCamera Module, Added support for Sony A7R and others

magicrub:
Re-submit of #243 for Jaime that is rebased and consolidated into a single commit:

--- original PR msg from Jaime ---------------
Added a log file to the smart camera module. Currently it is located in /sdcard/log, if the path does not exist the module currently fails to load. 
Added support for more sony cameras that failed to work before due to a missing command startRecMode. 
Added Initial setting of parameters to Shutter Priority Mode, ISO Auto, Shutter Speed 1/1600. These values can be modified by sending the appropriate commands after the module is loaded, or changing the default values in the python script.


magicrub:
@tridge please use this PR instead of https://github.com/Dronecode/MAVProxy/pull/243

tridge:
applied, thanks Tom!

PR #256, opened 2016-03-05
Smart camera module

jmachuca77:
Added an initial picture to be taken to let you know when the camera is ready, Fixed an issue with the GPS and Attitude Logging. Moved the wireless port to be used into the config file to allow for easy setting.


tridge:
applied, thanks!

PR #258, opened 2016-03-07
fix #257 - Get rid of cv2.cv in order to support OpenCV3

### LINKED ISSUE #257: Get rid of cv2.cv in order to support OpenCV3
I'm developing using OpenCV3 (master, compiled from sources). MavProxy's map does not like OpenCV3, mainly because of the outdated `import cv2.cv as cv` (and the respective legacy C-bindings of OpenCV), as OpenCV3 [got rid of cv2.cv](http://stackoverflow.com/a/26859781/2559632).

I don't think that the C-bindings dependency is still necessary (open for discussion) and am currently working (on an almost done) PR which allows the use OpenCV3 alongside OpenCV2 for the `map module`.

shadow1runner:
This fix exchanges (hopefully) all OpenCV1 image structures through numpy arrays, therefore all comparisons (equality etc.) had to be adapted;
Additionally, the signatures of many OpenCV functions have changed (and many copies were not necessary any more).

The following tests have been performed:
- change map provider
- zoom level
- brightness (note that increasing brightness might lead to overflow artefacts, was this always the case?)
- display/alter waypoints in a mission

Using `OpenCV 3`.

It would be great if somebody could look over these changes, as I'm not aware of the impact of this PR (e.g. maybe a module which makes use of map etc.) - thanks

Note that two of the changed files are part of `/modules/lib`, I've checked their usages and PyCharm couldn't find any other callers than map.
Furthermore, I made some indentation adaptations, as PyCharm (and Sublime) had problems with the mixed tabs/spaces.

### DIFF
--- a/MAVProxy/modules/mavproxy_map/mp_tile.py
+++ b/MAVProxy/modules/mavproxy_map/mp_tile.py
@@ -19,11 +19,8 @@ import os
 import sys
 import string
 import time
-
-try:
-	import cv2.cv as cv
-except ImportError:
-	import cv
+import cv2
+import numpy as np
 
 from MAVProxy.modules.lib import mp_util
 
@@ -44,8 +41,8 @@ TILE_SERVICES = {
 	"MicrosoftSat"   : "http://ecn.t${MS_DIGIT}.tiles.virtualearth.net/tiles/a${QUAD}.png?g=441&mkt=en-us&n=z",
 	"MicrosoftMap"   : "http://ecn.t${MS_DIGIT}.tiles.virtualearth.net/tiles/r${QUAD}.png?g=441&mkt=en-us&n=z",
 	"MicrosoftTer"   : "http://ecn.t${MS_DIGIT}.tiles.virtualearth.net/tiles/r${QUAD}.png?g=441&mkt=en-us&shading=hill&n=z",
-        "OviSat"         : "http://maptile.maps.svc.ovi.com/maptiler/v2/maptile/newest/satellite.day/${Z}/${X}/${Y}/256/png8",
-        "OviHybrid"      : "http://maptile.maps.svc.ovi.com/maptiler/v2/maptile/newest/hybrid.day/${Z}/${X}/${Y}/256/png8",
+	"OviSat"         : "http://maptile.maps.svc.ovi.com/maptiler/v2/maptile/newest/satellite.day/${Z}/${X}/${Y}/256/png8",
+	"OviHybrid"      : "http://maptile.maps.svc.ovi.com/maptiler/v2/maptile/newest/hybrid.day/${Z}/${X}/${Y}/256/png8",
 	"OpenStreetMap"  : "http://tile.openstreetmap.org/${ZOOM}/${X}/${Y}.png",
 	"OSMARender"     : "http://tah.openstreetmap.org/Tiles/tile/${ZOOM}/${X}/${Y}.png",
 	"OpenAerialMap"  : "http://tile.openaerialmap.org/tiles/?v=mgm&layer=openaerialmap-900913&x=${X}&y=${Y}&zoom=${OAM_ZOOM}",
@@ -56,8 +53,8 @@ TILE_SERVICES = {
 
 # these are the md5sums of "unavailable" tiles
 BLANK_TILES = set(["d16657bbee25d7f15c583f5c5bf23f50",
-                   "c0e76e6e90ff881da047c15dbea380c7",
-		   "d41d8cd98f00b204e9800998ecf8427e"])
+					"c0e76e6e90ff881da047c15dbea380c7",
+					"d41d8cd98f00b204e9800998ecf8427e"])
 
 # all tiles are 256x256
 TILES_WIDTH = 256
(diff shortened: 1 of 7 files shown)


tridge:
thanks for this!
main thing I need to test this against is the cuav extensions, which make extensive use of the map module

tridge:
Doing a bit of testing now. 
The first thing I've noticed is the vehicle icons don't appear on the map in SITL.

shadow1runner:
Ou you are right, sorry for that. Apparently I've spend too much time digging into details and got lost in them... 
I will look into it

shadow1runner:
Ok, I fixed it, was a rather stupid bug.
Now the icon (only tested with copter, but I guess that's the same) is displayed again:

image

tridge:
the icons are still not right. The old icons are solid, eg. plane icon can be seen here:
http://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html
new ones have weird artefacts, but at least they are visible :)

tridge:
it also needs a rebase on master

shadow1runner:
I thought that it looks rather weird, but I don't know what can be wrong with these lines of code:

CV:
```python
cv.SetImageROI(icon, (sx, sy, w, h))
cv.SetImageROI(img, (px, py, w, h))
cv.Add(icon, img, img)
cv.ResetImageROI(img)
cv.ResetImageROI(icon)
```

cv2:
```python
img[py:py+h, px:px+w] += icon[sy:sy+h, sx:sx+w]
```

Acc to the [docs](http://docs.opencv.org/2.4/modules/core/doc/operations_on_arrays.html#add) they should be equivalent

squilter:
@shadow1runner The docs say:
`dst += src1; // equivalent to add(dst, src1, dst);`

Which means that the original is adding icon to img, and yours is adding img to icon. Can you try it the other way around?

shadow1runner:
Correct me if I'm wrong but I'm adding icon to the img and the result is stored in img, and that's what we want, at least I don't see an obvious mistake here, unfortunately :S

shadow1runner:
@squilter mb you have looked at the snd last commit, there it was the case indeed, but I've fixed that by now. Still it's the blending that looks weird, mb it's a mismatch in image channels or sth like that, I need to investigate it further, but right now I'm busy with exams...

squilter:
Any updates here?

shadow1runner:
So, the icons for copter still look weird, but I guess that's the way they are, at least [here](http://ardupilot.org/dev/docs/sitl-native-on-windows.html) they look the same.

For comparison:
* Raw icon (overriding everything else)
mavproxyoverride

* Before fix
mavproxy_overlay

* Proposed
mavproxy cv2 add

@tridge, @squilter could you please test this? Unfortunately jsbsim doesn't build currently, so I'm not able to tests plane.

Also, I'm pretty sure, that there are more gotchas hidden somewhere, but Tridge is the expert on this :)

shadow1runner:
Sorry for the very late response, I'm currently busy with finishing my thesis. As soon as this one is done, hopefully in two week by now, I've got spare time to look into this issue, but I'm willing to fix it, cause it's an important for MavProxy. I've looked into it once for three hours and noticed another bug I fixed back then, but the artifacts still need more investigation.

On 2016-10-02 01:58, Sebastian Quilter wrote:

Any updates here?

squilter:
I tested with copter like above, and it worked with no problems. Plane also seemed to work fine. Here is a screenshot. 
screenshot from 2016-10-26 19-05-07

peterbarker:
Pretty font, but probably not appropriate :-)

fred

peterbarker:
Seems to work OK for me.

shadow1runner:
Thank you both for testing, I fixed the font problem, it should look now as it used to:
screenshot from 2016-10-27 09-00-30

MAVProxy/modules/mavproxy_map/mp_tile.py
```diff
@@ -382,35 +371,30 @@ def load_tile(self, tile):
...
+				except Exception:
+						self._download_pending[key] = tile
+				self.start_download_thread()
+					
 			# add it to the tile cache
 			self._tile_cache[key] = ret
 			while len(self._tile_cache) > self.cache_size:
 				self._tile_cache.popitem(0)
 			return ret
-		except IOError as e:
```
khancyr: Why remove this ?

shadow1runner: Because to me the only reason for the try/catch block was due the fact that `cv.LoadImage` throws an IOError. In case of cv2 a non-existent image always delivered `None`, therefore I thought it would be ok to remove it. Do you think that there are other reasons?

khancyr: It seems a valid reason ! I was asking because I doesn't see why it was removed ! But now I understand !

tridge:
I did some work on this yesterday. I managed to get cuav to work with some changes which I've put in this tree:
https://github.com/tridge/MAVProxy/tree/shadow1runner-master
I know the changes aren't pretty ....
Ideally I'd convert all of cuav to use cv2, but I haven't done that yet
Do you think the above is OK as an iterim?

shadow1runner:
Thank you for taking the time to test this.
I'm fine with these changes - they might not be ideal, but I think it might be helpful to many if this one goes into master

tridge:
ok, applied, thanks!

PR #265, opened 2016-04-05
Firmware manifests

peterbarker:
firmware manifest handling and firmware download


tridge:
I'll apply this, but I think we'd need a GUI to really make it useful

PR #267, opened 2016-04-25
Map: Added landing zone display for UAV Challenge

stephendade:
Popup menu on map now has option to show the 30-80m landing zone around the clicked position


tridge:
applied, thanks!

PR #269, opened 2016-04-27
link: change prompt immediately on mode change

peterbarker:
self.say(...) is still only called a maximum once/2 seconds

This should make the output from autotest clearer; currently several prompts appear with the old mode after a mode change


tridge:
applied, thanks!

PR #277, opened 2016-05-15
Maps: Fixed offline handling when there are no cached files

stephendade:
Fixed issues where there is no maps cache and the user is offline

### DIFF
--- a/MAVProxy/modules/mavproxy_map/srtm.py
+++ b/MAVProxy/modules/mavproxy_map/srtm.py
@@ -21,6 +21,7 @@ import tempfile
 
 childTileDownload = {}
 childFileListDownload = {}
+filelistDownloadActive = 0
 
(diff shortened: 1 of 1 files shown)


tridge:
Hi Stephen, it looks like this will put the map into offline mode if there is any error downloading any map tile. I don't see a way to get it out of offline mode again.
I think it would be better to handle the errors individually, otherwise if you have a slightly flakey internet link the map module could go offline until you restart (or reload the map module)

stephendade:
Tridge - I've fixed up the offline/online handling. Should work now.

tridge:
i'm seeing merge conflicts now, sorry for being slow

stephendade:
Rebased - should all be good now

amilcarlucas:
@tridge  conflicts are solved now

PR #291, opened 2016-06-16
Logging: OS check performed before free disk space check

stephendade:
Windows doesn't support the os.statvfs function so it is now skipped during the free disk space check, rather than just crashing.


magicrub:
@stephendade Does this mean you're doing windows builds? Would be nice to get an updated windows release build: https://github.com/Dronecode/MAVProxy/issues/289

stephendade:
@magicrub - Yep, that's exactly it. This bug was causing the Windows builds to fail.

PR #294, opened 2016-06-30
link: Stop erroneous 'Time has wrapped' messages.

mday299:
SET_ATTITUDE_TARGET and SET_POSITION_TARGET_GLOBAL_INT messages don't contain boot time and should not cause the 'Time has wrapped' warning.


mday299:
@tridge Any comments on this one?

mday299:
My mistake... these message do have boot time.

PR #295, opened 2016-07-05
Add workaround for Solo heartbeat issues.

jmachuca77:
This is a workaround for systems that do not interpret heartbeats correctly, by filtering them according to the system/component values. It sets heartbeat setting to 0 to stop sending heartbeats after 10 seconds. This is needed because the pixhawk needs to receive some heartbeats to learn that there is a component present. After that we can stop sending them and the Pixhawk will continue to send them out.

### DIFF
--- a/MAVProxy/modules/mavproxy_smartcamera/__init__.py
+++ b/MAVProxy/modules/mavproxy_smartcamera/__init__.py
@@ -16,7 +16,7 @@
 #
 # Responsible  : Jaime Machuca
 #
-# License      : CC BY-NC-SA
+# License      : GNU GPL version 3
 #
 # Editor Used  : Xcode 6.1.1 (6A2008a)
 #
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/mavproxy_smartcamera/__init__.py
```diff
@@ -111,13 +139,15 @@ def __vRegisterQXCamera(self,u8CamNumber):
...
                 print("Found QX Camera")
+                self.master.mav.statustext_send(6,"Camera found status text")
             else:
                 print("No Valid Camera Found, retry in 5 sec")
                 self.u8RetryTimeout = self.u8RetryTimeout + 1
                 self.CamRetryScheduler.enter(5, 1, self.__vRegisterQXCamera, [u8CamNumber])
                 self.CamRetryScheduler.run()
         else:
             print("Max retries reached, No QX Camera Found")
+            self.master.mav.statustext_send(6,"Camera not found status text")
```
magicrub: change "Camera not found status text" to "Camera QX not found"

MAVProxy/modules/mavproxy_smartcamera/__init__.py
```diff
@@ -111,13 +139,15 @@ def __vRegisterQXCamera(self,u8CamNumber):
             if new_camera.boValidCameraFound() is True:
                 self.camera_list = self.camera_list + [new_camera]
                 print("Found QX Camera")
+                self.master.mav.statustext_send(6,"Camera found status text")
```
magicrub: change "Camera found status text" to "Camera QX found"

jmachuca77:
reworking, I will reopen it when ready

PR #296, opened 2016-07-12
MAVCesium display as a submodule to MAVProxy

SamuelDudley:
Hi,
I have created a new map / HUD / synthetic FPV module for MAVProxy called [MAVCesium](https://github.com/SamuelDudley/MAVCesium). The module display runs in a webgl enabled browser and the module dependencies are pure python and _should_ be cross platform.

This pull request adds [MAVCesium](https://github.com/SamuelDudley/MAVCesium) as an optional submodule for MAVProxy.

I'm hoping to have [MAVCesium](https://github.com/SamuelDudley/MAVCesium) as part of the MAVProxy repo and this is the least intrusive way I could think about integrating it. I'm open to ideas and suggestions re other ideas.

Thanks,
Sam

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "MAVProxy/modules/mavproxy_cesium"]
+	path = MAVProxy/modules/mavproxy_cesium
+	url = https://github.com/SamuelDudley/MAVCesium.git
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-recursive-include MAVProxy/modules/mavproxy_cesium/app *
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include MAVProxy/modules/mavproxy_cesium/app *
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,25 @@
 from setuptools import setup
+import os
 
 version = "1.5.1"
 
+def package_files(directory):
+    paths = []
+    for (path, directories, filenames) in os.walk(directory):
+        for filename in filenames:
+            paths.append(os.path.join('..', path, filename))
+    return paths
+
+package_data = ['modules/mavproxy_map/data/*.jpg', 
+                'modules/mavproxy_map/data/*.png',
+                'tools/graphs/*.xml',
+]
+
+package_data.extend(package_files('MAVProxy/modules/mavproxy_cesium/app'))
+
 setup(name='MAVProxy',
       version=version,
-      zip_safe=False,
-      include_package_data=True,
+      zip_safe=True,
       description='MAVProxy MAVLink ground station',
       long_description='''A MAVLink protocol proxy and ground station. MAVProxy
 is oriented towards command line operation, and is suitable for embedding in
@@ -51,7 +65,5 @@ on how to use MAVProxy.''',
                'MAVProxy/modules/mavproxy_map/mp_slipmap.py',
                'MAVProxy/modules/mavproxy_map/mp_tile.py'],
       package_data={'MAVProxy':
-                    ['modules/mavproxy_map/data/*.jpg', 
-                     'modules/mavproxy_map/data/*.png',
-                     'tools/graphs/*.xml']}
+                    package_data}
     )
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ version = "1.5.1"
 
 setup(name='MAVProxy',
       version=version,
-      zip_safe=True,
+      zip_safe=False,
       description='MAVProxy MAVLink ground station',
       long_description='''A MAVLink protocol proxy and ground station. MAVProxy
 is oriented towards command line operation, and is suitable for embedding in
@@ -34,6 +34,7 @@ on how to use MAVProxy.''',
                 'MAVProxy.modules.mavproxy_map',
                 'MAVProxy.modules.mavproxy_misseditor',
                 'MAVProxy.modules.mavproxy_smartcamera',
+                'MAVProxy.modules.mavproxy_cesium',
                 'MAVProxy.modules.lib',
                 'MAVProxy.modules.lib.ANUGA',
                 'MAVProxy.modules.lib.optparse_gui'],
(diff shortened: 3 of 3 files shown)


rmackay9:
very pretty.  You don't happen to have any videos of it in action do you?  (This request is not a requirement to get it into master - that decision is up to Tridge).

peterbarker:
On Mon, 11 Jul 2016, Samuel Dudley wrote:

> I have created a new map / HUD / synthetic FPV module for MAVProxy called
> MAVCesium. The module display runs in a webgl enabled browser and the module
> dependencies are pure python and should be cross platform.
> 
> This pull request adds MAVCesium as an optional submodule for MAVProxy.
> 
> I'm hoping to have MAVCesium as part of the MAVProxy repo and this is the
> least intrusive way I could think about integrating it. I'm open to ideas
> and suggestions re other ideas.

My first impressions are "awesome"!

Could I suggest a patch to setup.py to install cesium when a "python 
setup.py build install" is used?

This is what I currently have:

diff --git a/setup.py b/setup.py
index a6c27c7..f4d8bde 100644
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ version = "1.5.1"

  setup(name='MAVProxy',
        version=version,
-      zip_safe=True,
+      zip_safe=False,
        description='MAVProxy MAVLink ground station',
        long_description='''A MAVLink protocol proxy and ground station. 
MAVProxy
  is oriented towards command line operation, and is suitable for embedding 
in
@@ -31,6 +31,8 @@ on how to use MAVProxy.''',
        license='GPLv3',
        packages=['MAVProxy',
                  'MAVProxy.modules',
+                'MAVProxy.modules.mavproxy_cesium',
+                'MAVProxy.modules.mavproxy_cesium.app',
                  'MAVProxy.modules.mavproxy_map',
                  'MAVProxy.modules.mavproxy_misseditor',
                  'MAVProxy.modules.mavproxy_smartcamera',
@@ -47,9 +49,13 @@ on how to use MAVProxy.''',
                 'MAVProxy/tools/mavflightview.py',
                 'MAVProxy/tools/MAVExplorer.py',
                 'MAVProxy/modules/mavproxy_map/mp_slipmap.py',
-               'MAVProxy/modules/mavproxy_map/mp_tile.py'],
+               'MAVProxy/modules/mavproxy_map/mp_tile.py',
+ 
'MAVProxy/modules/mavproxy_cesium/app/cesium_io_server.py',
+      ],
        package_data={'MAVProxy':
                      ['modules/mavproxy_map/data/*.jpg',
                       'modules/mavproxy_map/data/*.png',
-                     'tools/graphs/*.xml']}
+                     'tools/graphs/*.xml',
+                     'modules/mavproxy_cesium/app/api_keys.txt',
+                    ]}
      )


The zip_safe is a wart; it can't find the io server without that (or a 
more correct fix...).

Also, it seems that a bing API key is required; certainly I had to create 
an api_keys.txt or I got an IO error when it couldn't find it.  I now get 
a keyerror mentioning "bing":

   File 
"/home/pbarker/.local/lib/python2.7/site-packages/MAVProxy-1.5.1-py2.7.egg/MAVProxy/modules/mavproxy_cesium/app/cesium_io_server.py", 
line 55, in index
     return render_template('index.html', bing_api_key=api_keys['bing'])
KeyError: 'bing'
[2016-07-12 19:38:09,196] ERROR in app: Exception on / [GET]

I'd love to get this working, but that's a blocker for me :-)

Peter

peterbarker:
@rmackay9 Your browser isn't capable of seeing the live demo?  http://www.MAVCesium.io/

rmackay9:
@peterbarker, ah very nice.

SamuelDudley:
@peterbarker Thanks for giving it a try! I have updated the submodule with a fix for the missing [api_keys.txt](https://github.com/SamuelDudley/MAVCesium/blob/master/app/api_keys.txt). I had added it to .gitignore and it was not being pulled from github. As you have already made the file in the app directory you can just update the contents of [api_keys.txt](https://github.com/SamuelDudley/MAVCesium/blob/master/app/api_keys.txt) to match what is in the repo: `{"bing":""}`.

I have not used the "python setup.py build install" method before. Thanks for your efforts. I'll have a look at how I can make that process smoother. As an idea I might be able to add the io server to the path when the module loads.

@rmackay9 With the dynamic lighting option on (I've enabled it by default) the display will match the lighting conditions for your local time of day. So for me its currently night time and wont make for a very nice video :). I can add one tomorrow morning if you still would like.

Some other notes:
The 'clunky' update rate on the live demo is because of the default telemetry stream rate output by APM when connected to MAVProxy. If you run it locally try setting the stream rate higher with `set streamrate 10` and it becomes much smoother, almost like a flight sim!

The HUD was thrown together with some info that I personally find useful and can be toggled on / off by pressing the `h` key. You can also limit it to half width in the settings menu (currently a bit of a dumping ground and needs work / organisation). I case anyone is wondering the small circles on the roll gauge and pitch ladder display the nav roll and nav pitch target respectively. If there is anything you would like added to the HUD let me know and I'll put it on the list.

Cheers,
Sam

peterbarker:
On Tue, 12 Jul 2016, Samuel Dudley wrote:

> @peterbarker Thanks for giving it a try! I have updated the submodule with a
> fix for the missing api_keys.txt. I had added it to .gitignore and it was
> not being pulled from github. As you have already made the file in the app
> directory you can just update the contents of api_keys.txt to match what is
> in the repo: {"bing":""}.

That worked - I got it running with those changes, thanks!

> I have not used the "python setup.py build install" method before. Thanks
> for your efforts. I'll have a look at how I can make that process smoother.
> As an idea I might be able to add the io server to the path when the module
> loads.

Somewhere in here is probably the right way to do it:

http://peak.telecommunity.com/DevCenter/setuptools#accessing-data-files-at-runtime

This tree has a setup file which Works For Me - it probably isn't the best 
way to do things...

https://github.com/peterbarker/mavproxy/tree/MAVCesium

You'll have to wait for @tridge to comment on how this should all be done; 
obviously external modules aren't something that has been done, and tridge 
might prefer to pull the module content into MAVProxy's repository, or 
prefer people to reference the external module in some other way.

> Sam

Peter

SamuelDudley:
@peterbarker glad you got it working and thank you for your setup.py file :) . I was working on it at the same time and came to a different end point using a manifest.in file. I have updated this branch with the changes from my experimenting. It would be interesting to see if what I have works for you.

I'm pretty sure by adding the manifest.in I can side step some of the complexity of the setup.py script.
I don't know which approach is 'better', but as you said they are probably both not 'correct'...

SamuelDudley:
@peterbarker I have solved the zipped python egg issue and went with your code in setup.py in the end.
It ended up being much tidier than having a second file to do exactly what your code already did!

tridge:
looks very nice, thanks!
applied

PR #299, opened 2016-07-20
Logging: removed daemon hard-set directory of /var/log

stephendade:
Using /var/log as the daemon logfile directory required root. Removed this, so the user can decide where they want their logfiles stored.


tridge:
applied, thanks!

PR #300, opened 2016-07-26
mavproxy_tuneopt: update tune options

mirkix:
Update tune options to http://ardupilot.org/copter/docs/parameters.html#tune-channel-6-tuning


tridge:
applied, thanks!

PR #301, opened 2016-07-28
mavproxy_log: send mavlink log_request_end when download done

peterbarker:
Without this ArduPilot remains in the in_log_download=True state,
meaning logging will not restart after downloading a log


tridge:
applied, thanks!

PR #302, opened 2016-07-28
Pass mode through to set_mode

squilter:
This simply passes the desired mode (int or string), and lets pymavlink deal with the rest. Pymavlink already supports this, so this shouldn't break anything. I'll also submit a PR to pymavlink to add support for px4 modes, but this PR can safely be merged before that one.

### DIFF
--- a/MAVProxy/modules/mavproxy_mode.py
+++ b/MAVProxy/modules/mavproxy_mode.py
@@ -22,14 +22,8 @@ class ModeModule(mp_module.MPModule):
             print('Available modes: ', mode_mapping.keys())
             return
         if args[0].isdigit():
-            modenum = int(args[0])
-        else:
-            mode = args[0].upper()
-            if mode not in mode_mapping:
-                print('Unknown mode %s: ' % mode)
-                return
-            modenum = mode_mapping[mode]
-        self.master.set_mode(modenum)
+            self.master.set_mode(int(args[0]))
+        self.master.set_mode(args[0].upper())
 
     def available_modes(self):
         mode_mapping = self.master.mode_mapping()


tridge:
this breaks setting mode by mode number in mavproxy. I use that a lot when developing a new mode, before support has been added in pymavlink for the mode.

squilter:
Good point. Fixed.

weiminshen99:
Could you please explain how I can use this feature to support the OFFBOARD mode for PX4? Right now, when I switch to that mode, MAVProxy complains that the mode is UNKNOWN. Thanks!

peterbarker:
On Wed, 6 Dec 2017, weiminshen99 wrote:

> Could you please explain how I can use this feature to support the OFFBOARD
> mode for PX4? Right now, when I switch to that mode, MAVProxy complains
> that the mode is UNKNOWN. Thanks!

We notionally support decoding PX4 modes.

Looks like you've found a value which we don't decode.

https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L1562

That's the code which tries to work out what's going on.

Are you up to tracing that code through with the value you have?

weiminshen99:
Hi, Peter,

Thank you very much for the pointer into the code! 

I will trace it and do a fix/test, and I will let you know the results.

Best wishes,

Wingman

weiminshen99:
Hi, Peter,
After much search and testing, here is the conclusion I reached:
       pymavlink 2.0.6 will not work, and the problem is as I experienced and reported;
       pymavlink 2.2.4 will work with the OFFBOARD mode. However, this version still does not support some of the newer mode such as TAKEOFF, but that does not matter much for me.
In summary, MAVProxy with pymavlink2.2.4 will support offboard, and I tested on my UAV with real flights.
Finally, I would like to thank you all for the feedback and helps you gave me over the process. I hope this shared experience will be useful for people in the future. Thanks!

amilcarlucas:
Looks like the solution is to update pymavlink

PR #305, opened 2016-08-23
link: colour APM statustext messages based on severity

peterbarker:


tridge:
very nice, thanks!
It will be interesting to see how this does in bright sunlight on a laptop, we may need to adjust colours. I'll push it in now then we can adjust if need be.
applied

PR #306, opened 2016-08-24
Added kml reader module. Allows kml files to be rendered on map

stephendade:
Initial commit of module. Currently allows kml and kmz format files to be displayed on the map, cleared and the individual layers listed


tridge:
looks great, thanks! 
pushed

PR #308, opened 2016-09-02
modules: add a simple example module

peterbarker:
This module should be short, so people can copy it and trim very little.

It should also be long enough that you don't need to go hunting for
commonly-used things like mavlink message handling


peterbarker:
This was merged in.

PR #310, opened 2016-09-03
Readkml: Added right-click menu to map and ability to load to geofence

stephendade:
-Added to the right-click popup menu in the map
-Able to toggle display of individual layers
-Can send polygons to the geofence

### DIFF
--- a/MAVProxy/modules/mavproxy_kmlread.py
+++ b/MAVProxy/modules/mavproxy_kmlread.py
@@ -7,7 +7,7 @@ Released under the GNU GPL version 3 or later
 '''
 
 import time, math, random
-from pymavlink import mavutil
+from pymavlink import mavutil, mavwp
 from cStringIO import StringIO
 from xml.dom.minidom import parseString
 from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_kmlread.py
```diff
@@ -68,19 +188,49 @@ def loadkml(self, filename):
...
+                self.curtextlayers.append(point[1])                
+
+    def idle_task(self):
+        '''handle GUI elements'''
+        if self.module('map') is not None and not self.menu_added_map:
+            self.menu_added_map = True
+            self.module('map').add_menu(self.menu)
+        #(re)create the menu
+        if mp_util.has_wxpython and self.menu_added_map:
+            # we don't dynamically update these yet due to a wx bug
```
tridge: I don't think all this needs to be done on every call of idle_task(). That seems to just waste cpu

tridge:
looks very useful! just need to make the idle_task more efficient if we can

stephendade:
@tridge - I've fixed the menu refresh.

tridge:
applied, thanks!

PR #311, opened 2016-09-07
arm: warn if the user arms when any check is disabled

peterbarker:


MAVProxy/modules/mavproxy_arm.py
```diff
@@ -120,6 +120,25 @@ def cmd_disarm(self, args):
             0, # param6
             0) # param7
 
+    def all_checks_enabled(self):
+        ''' returns true if the UAV is skipping any arming checks'''
```
OXINARF: returns _false_

MAVProxy/modules/mavproxy_arm.py
```diff
@@ -120,6 +120,25 @@ def cmd_disarm(self, args):
...
+        if arming_mask == 1:
+            return True
+        for bit in arming_masks.values():
+            if not arming_mask & bit and bit != 1:
+                return False
+        return True
+
+    def mavlink_packet(self, m):
+        mtype = m.get_type()
+        if mtype == 'HEARTBEAT' and m.type != mavutil.mavlink.MAV_TYPE_GCS:
```
OXINARF: Is this correct? What if the heartbeat is an antenna tracker or gimbal type or etc.?

peterbarker: On Wed, 7 Sep 2016, Francisco Ferreira wrote:

> Is this correct? What if the heartbeat is an antenna tracker or gimbal type
> or etc.?

Wildly incorrect.  But I swiped it from mavproxy_link.py, so at least 
we're consistently incorrect.

I think the correct fix might be to check the hearbeat actually came from 
*our* mav.  I would dearly love to find lots of time to use the mav source 
system pymavlink code 
(https://github.com/ArduPilot/mavlink/compare/master...peterbarker:mavsystem-object?expand=1) 
to give MAVProxy proper multi-vehicle support.

We're relying on master.motors.armed() to be correct here, so I think the 
net result of that check being bodgy is that we might run that check more 
often than we need to.

MAVProxy/modules/mavproxy_arm.py
peterbarker: On Wed, 7 Sep 2016, Francisco Ferreira wrote:

> > +    def all_checks_enabled(self):
> > +        ''' returns true if the UAV is skipping any arming checks'''
> 
> returns false

Darn!  I inverted that function from "any_checks_disabled()"!

MAVProxy/modules/mavproxy_arm.py
OXINARF: Yeah, I guess there won't be a problem. It would be good to have this corrected though, I'll look how to make days have 48 hours instead of 24, I need it too :smile:

peterbarker:
This was merged.  Thanks @OXINARF.

PR #316, opened 2016-10-02
MAVExplorer: Graph and map processing moved to main process

stephendade:
For better Windows performance (and 1/10 of the memory usage), the graph and map processing has been moved to the MAVExplorer process.

### DIFF
--- a/MAVProxy/modules/lib/grapher.py
+++ b/MAVProxy/modules/lib/grapher.py
@@ -293,9 +293,15 @@ class MavGraph(object):
             self.x[i].append(xv)
 
 
-    def process_mav(self, mlog, timeshift):
+    def process_mav(self, mlog, timeshift, flightmode_selections, _flightmodes):
         '''process one file'''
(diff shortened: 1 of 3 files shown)


tridge:
This seems to break the flightmode and condition filtering.
For example, if you select one of the flight modes in the Flight Mode menu, and do a graph, and then select a different flight mode then you end up with nothing displayed in the new graph. The code is applying the filtering to the master copy of the data, so it ends up with less and less data each time you apply a condition or flight mode filter.

tridge:
a fix might be that instead of actually filtering the data we instead pass the condition and a list of (start,stop) timestamps to represent the applied condition and the flight mode ranges. We could teach mavmemlog about those, and just create those per-graph so we aren't copying data

stephendade:
I've fixed the filtering problem and squashed my commits into a single one. Should be good to use now!

tridge:
thanks, applied!
we should get a new windows release out soon

PR #321, opened 2016-10-07
Misseditor unload fixed

peterbarker:


MAVProxy/modules/mavproxy_misseditor/__init__.py
```diff
@@ -235,13 +248,35 @@ def child_task(self, q, l, gq, gl):
...
+        watcher_quitsem.acquire()
+        class CloseWindowSemaphoreWatcher(threading.Thread):
+            def __init__(self, task, sem, quitsem):
+                threading.Thread.__init__(self)
+                self.task = task
+                self.sem = sem
+                self.quitsem = quitsem
+            def run(self):
+                while not self.sem.acquire(False) and not self.quitsem.acquire(False):
+                    time.sleep(0.1)
```
tridge: do we really need a thread waking up 10 times a second just to monitor window close?

peterbarker:
On Thu, 6 Oct 2016, Andrew Tridgell wrote:

> +                while not self.sem.acquire(False) and not self.quitsem.acqu
> ire(False):
> +                    time.sleep(0.1)
> 
> do we really need a thread waking up 10 times a second just to monitor
> window close?

Nope.  New version pushed up.

I just eliminated one and slightly-dodgily released the close_window 
semaphore instead.

tridge:
applied, thanks!

PR #323, opened 2016-10-10
Clarify link and console signing information

peterbarker:


tridge:
applied, thanks!
(with small typo fix)

PR #324, opened 2016-10-12
mavflightview: add flight mode legend

peterbarker:


tridge:
just tried this on a DF log with mavexplorer, and it shows the legend menu item, but the legend doesn't show up on the map
also the legend menu item is showing in the map in normal live use of mavproxy, which doesn't make sense.
could we make the legend menu item only show if there is a legend to show?

peterbarker:
@tridge Oops!  I hadn't intended to do either the live view or MAVExplorer integration this round.  Note I changed the commit messages to "map".

My last patch to add the interface elements to toggle it on obviously (in hindsight) impact both.

I'll do the MAVExplorer integration and work out how to hide it in the live view.  Dynamically updating hte menu items based on whether someone's given you a legend doesn't sound like a lot of fun, mind you.

peterbarker:
This should all work now.

tridge:
the only thing I don't like about this is that the show_flightmode MAVExplorer setting doesn't work to disable it

tridge:
i've pushed this in anyway, but it would be nice to make it obey show_flightmode in future

peterbarker:
Huh.  Who knew that existed?

PR to honour that setting: https://github.com/ArduPilot/MAVProxy/pull/368

Issue #329, opened 2016-10-30
Quickstart - confusing statement in "networking" section


monkeypants:
URL: ../getting_started/quickstart
Title: Quickstart

The last sentence on the page says:

> The address to connect to must be your own IP address.

This implies you can't connect to another machines IP address, which isn't true. I don't understand the point of this statement.

monkeypants:
After reading through the pages in linear sequence, I am told twice about only connecting to local addresses(quickstart#networking then startup options#master) before finally being told about UDP  broadcast outputs (startup options#out).

monkeypants:
Fixed by #332

PR #330, opened 2016-10-31
clarify networking docs

monkeypants:
Per #329, reading the docs front-to-back after a long lay-off, I found it confusing that I couldn't connect to remote network addresses. This was clarified a few screens of reading later, but I would like to propose this admonition to prevent future readers suffering the same confusion.

Also note, adding in the sphinx.ext.autosectionlable raised a few new build warnings (same section headings used in multiple parts of the document). I think those are legitimate warnings, so I don't think the fact it's now raising them is an issue.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -35,6 +35,7 @@ extensions = [
     'sphinx.ext.pngmath',
     'sphinx.ext.ifconfig',
     'sphinx.ext.viewcode',
+    'sphinx.ext.autosectionlabel'
 ]
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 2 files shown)


monkeypants:
@stephendade did you mean to close this with Unmerged commits?

I didn't commit changes to html/, figured I should commit in "make clean" state but I'm not sure that's right, there seems to be stuff committed there (and doctrees, etc). What's the process for committing doc changes?

stephendade:
I'd assumed you were finished making changes :)

./docs/sphinxLinux.sh will build the documentation into the html/ dir

monkeypants:
I made two PRs with two collections of commits.

Yes, I ran sphinxLinux.sh to check the work locally, but didn't commit the generated code. Do you want a fresh PR with everything in it?

I got a bit confused by some commits that only changed html/ (not the rst it's presumably all supposed to be generated from), and didn't want to touch it.

stephendade:
Yeah, a fresh PR would be good.

PR #332, opened 2016-10-31
docs: fix overloaded section headings

monkeypants:
proposal to address overloaded section headings per: #330


stephendade:
Thanks Chris!

PR #345, opened 2016-11-14
teach mavproxy about new "joymap" module

larsks:
This is an alternative to the existing "joystick" module.  Unlike the joystick module, it loads joystick definitions dynamically from YAMLfiles, which makes it easier for someone to add support for a device without needing to patch mavproxy itself.  Joystick definitions can be loaded from a user configuration directory (in addition to from the source itself).

Additionally, the "joymap" module supports a wider variety of control types.  A "hat" can be used as a pair of toggle switches, and buttons can be used either as momentary switches or grouped together to act
like a multi-position switch.

I would appreciate feedback on this module; if it is of general interest I would be happy to do the additional work necessary to add joystick definitions for all the currently supported joysticks and write up some user-facing documentation to go with it.


MAVProxy/modules/mavproxy_joymap/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+            path = os.path.expanduser(path)
+            for dirpath, dirnames, filenames in os.walk(path):
+                for joyfile in filenames:
+                    root, ext = os.path.splitext(joyfile)
+                    if ext[1:] not in ['yml', 'yaml', 'json']:
+                        continue
+
+                    joypath = os.path.join(dirpath, joyfile)
+                    self.log('Loading definition from {}'.format(joypath), 2)
+                    with open(joypath, 'r') as fd:
```
peterbarker: We should probably close fd after using it.

larsks:
At @peterbarker's request I've modified this PR so that it replaces the existing `mavproxy_joystick.py` module.  I would like to emphasize that this has so far only been tested under SITL with a Logitech F310 gamepad, so if you happen to have another device handy and care to try things out (possibly writing a joystick definition) I would appreciate some feedback (and I would be happy to render assistance in setting things up).

JAR4x4:
@larsks flew my quad my quad with this module and a custom joystick interface today. Wrote a new .yml file and everything worked nice.

JAR4x4:
Is there any way to adjust the expo on each axis? I think the response around the middle is very fast..

larsks:
@JAR4x4, there is no facility to do that right now (largely because the original joystick module didn't offer that feature, either), but I'm happy to consider it.  How do you think it should look?

tridge:
thanks for this great work!
I've pushed it in with the addition of the taranis, a fix for the great planes, and a small editing fix in some of the yml files
I really like being able to add new joysticks in $HOME/.mavproxy/joysticks

PR #348, opened 2016-11-18
MAVProxy Horizon Indicator Pull Request

tbattz:
An artificial horizon display for MAVProxy aiming to improve the situational awareness of the aircraft.

Features
- Artificial horizon responds to the roll and pitch of the aircraft
- The distance between the pitch markers can be changed by using the up and down arrow keys when the module has focus
- Heading points for the current heading, north and the next target point (green pointer)
- Current mode and arm state (green/red)
- Battery information, voltage, current draw, percentage. NOTE: In SITL this works for ArduCopter but not ArduPlane, as it seems that ArduPlane doesn't simulate the battery. In this case the bar defaults to black.
- Current airspeed, altitude (above ground) and climb rates

image

### DIFF
--- a/MAVProxy/modules/lib/wxhorizon.py
+++ b/MAVProxy/modules/lib/wxhorizon.py
@@ -1,14 +1,14 @@
 #!/usr/bin/env python
 
 """
-  MAVProxy message console, implemented in a child process
+  MAVProxy horizon indicator.
 """
 import multiprocessing
 import time
 
 class HorizonIndicator():
     '''
-    a message console for MAVProxy
+    A horizon indicator for MAVProxy.
     '''
     def __init__(self,title='MAVProxy: Horizon Indicator'):
         self.title  = title
@@ -34,7 +34,7 @@ class HorizonIndicator():
         app.MainLoop()
 
     def close(self):
-        '''close the console'''
+        '''Close the window.'''
         self.close_event.set()
         if self.is_alive():
             self.child.join(2)
--- a/MAVProxy/modules/lib/wxhorizon.py
+++ b/MAVProxy/modules/lib/wxhorizon.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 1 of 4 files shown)


gmorph:
It worked great with Plane.  The battery indicator showed a green bar as a percentage in the top right.  However when I used it with Rover whilst it did work fine the battery indicator on the top right was black as it is in your picture above.  Why is that.

Other then that works great.  If you looking for enhancement suggestions I'd love to see an altimeter gauge centre left which just scrolls the numbers for the alt up and down.

Thanks, Grant.

tbattz:
The battery indicator is based off of the SYS_STATUS message, and gets the information in the same way that the battery module does. If the battery remain reaches zero, then the bar is blacked out, indicating either that the battery % is zero or that there has been no battery information (through SYS_STATUS) being passed through since the startup of MAVProxy. I would assume that in this case this is due to the latter, as testing with Rover (and for me Plane) in SITL, the Battery percentage and current draw do not appear in the console module, while they do when running ArduCopter in SITL, matching the cases where the information is and isn't present in the horizon module.

The altimeter gauge is a good idea, I'll give it a shot at adding it at a later date. Shouldn't be too hard, would be similar to the pitch markers without the rotation and more numbers, but I might have to rearrange the layout a little bit to fit it.

gmorph:
My bad.  I didn't have the battery monitoring enabled.  By setting BATT_MONITOR to 4 it all came good.
Love this module!
Thanks, Grant.

tbattz:
So I had a go at adding a scrolling attitude indicator, while it partially worked it didn't turn out well. In the end I've opted to add a scrolling altitude plot. Each group of widgets can now be toggled visible/hidden using the numbers 1-6 (not numpad numbers). The mode and arm state updating has been tweaked, along with some performance improvements.
Horizon Indicator

tridge:
really nice work, thanks!!
applied

PR #352, opened 2016-11-22
link: add force-connected option

peterbarker:
This will allow you to connect a named serial device after mavproxy has started.

It will also allow gdb_stopped in sim_vehicle.py again.


tridge:
would this have the same effect as --nowait does now?

tridge:
plus this looks like it needs a pymavlink change to add force_connected

khancyr:
@tridge this extend the no-wait and reconnect option by allowing to connect to a port that isn't already ready or reconnect when the port disconnect (like usb)
the pymavlink part is here https://github.com/ArduPilot/pymavlink/pull/39

Issue #357, opened 2016-12-19
Graph module's left side scale is odd if value is constant for too long


rmackay9:
![mavproxy_graph_leftsidescale](https://cloud.githubusercontent.com/assets/1498098/21301678/0d641f6e-c5f3-11e6-9926-98c495670522.png)

In the 2nd picture above you can see how the scale on the left always shows "0.10" if the value doesn't change for a while.

stephendade:
Yep, that's the exact behavior according to lines 108-110 of MAVProxy/modules/lib/live_graph_ui.py

If the graph line is a constant value, the y-axis limits go to +-0.1 around the constant value. Really need a method to have smarter y-axis limits

rmackay9:
Thanks.  It actually prints the wrong value though as well.  I could live with it squeezing down to just a small area around the value.. but look at the value shown!  It doesn't show 1850, instead it show 0.1... a completely unrelated value (as far as I can tell it's unrelated anyway).

stephendade:
It's the graph formatting - there's a exponent (partially) hiding behind the legend. Numbers are correct, just badly formatted. I've fixed it a #361, just have to get @tridge to merge it in.

stephendade:
Has been committed to master.

rmackay9:
Thanks very much!

PR #361, opened 2016-12-26
Live Graph UI: nicer scaling and formatting for y-axis

stephendade:
Fix for #357


rmackay9:
That issue (https://github.com/ArduPilot/MAVProxy/issues/357) is from me, so I'm very happy if this goes in.

PR #369, opened 2017-01-19
rccsetup: set parameters SERVOx_[TRIM|MIN|MAX]

mothacehe:
During a calibration it is now needed to set SERVOx_[TRIM|MIN|MAX] to
the same values as RCx_[TRIM|MIN|MAX].

Also add messages during trim calibration.

### DIFF
--- a/MAVProxy/modules/mavproxy_rcsetup.py
+++ b/MAVProxy/modules/mavproxy_rcsetup.py
@@ -32,6 +32,11 @@ class RCSetupModule(mp_module.MPModule):
             self.param_set('RC%u_MAX' % i, self.rc_cal[i][1], 5)
             self.console.writeln("Set: RC%u_MAX=%u" % (i, self.rc_cal[i][1]))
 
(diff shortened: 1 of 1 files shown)


tridge:
R/C calibration is about input calibration, not output calibration. We shouldn't assume that input ranges and trims match output ranges and trims.

mothacehe:
Ok, there must be something I don't understand then. To calibrate a plane with ardupilot master my procedure is:
- Start with default min/trim/max values for input and output.
- Set my RC trims correctly.
- Calibrate inputs using MAVProxy.

If I stop there, I end up in a situation where my input values are remapped on default output values.
I have to copy manually RCx_[TRIM|MIN|MAX] to SERVOx_[TRIM|MIN|MAX] parameters.

Is that the expected procedure ?

amilcarlucas:
I think the  latest code no longer needs this and this can be closed.

amilcarlucas:
@tridge or @OXINARF  can we close this one ?

PR #376, opened 2017-02-08
Windows installer: Fix quotes around paths with whitepaces in a simpl…

amilcarlucas:
…er way

Sorry for the mess I did with the previous PR, this is the correct location of the quotes


stephendade:
Looks good. Thanks!

PR #377, opened 2017-02-08
Windows installer: Rebuild and use pymavlink from sources if available

amilcarlucas:
If the user has a local copy of pymavlink chances are that he/she wants to use it :)

### DIFF
--- a/windows/MAVProxyWinBuild.bat
+++ b/windows/MAVProxyWinBuild.bat
@@ -22,12 +22,19 @@ rem -----build the changelog-----
 
 rem -----Upgrade pymavlink if needed-----
 if exist "..\..\pymavlink" (
- rem Rebuild and use pymavlink from sources if available
+ rem Rebuild and use pymavlink from pymavlink sources if available
  pushd ..\..\pymavlink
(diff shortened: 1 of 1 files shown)


windows/MAVProxyWinBuild.bat
```diff
@@ -21,7 +21,14 @@ rem -----build the changelog-----
 "%PYTHON_LOCATION%\python" createChangelog.py
 
 rem -----Upgrade pymavlink if needed-----
-"%PYTHON_LOCATION%\Scripts\pip" install pymavlink -U
+if exist "..\..\pymavlink" (
```
stephendade: Shouldn't this be "..\..\mavlink\pymavlink"? I suppose it depends on the user configuration - whether they're using the mavlink repository or pymavlink repository.

amilcarlucas: Yes, I will rebase this, with an "else if" for that case as well

amilcarlucas:
I think now it has the features you wanted

stephendade:
Thanks!

PR #381, opened 2017-02-20
Support dbl left clicking of urls in console window

SamuelDudley:
This PR adds support for click-able hyperlinks in the console window. When double left clicked (or wx event equivalent) a new tab is opened to the url address. This PR is to support the MAVCesium module [issue](https://github.com/SamuelDudley/MAVCesium/issues/4) but I feel it could also be a useful tool to direct users to support pages, documentation or FW downloads.

screenshot from 2017-02-20 21 53 26

### DIFF
--- a/MAVProxy/modules/lib/wxconsole_ui.py
+++ b/MAVProxy/modules/lib/wxconsole_ui.py
@@ -3,7 +3,7 @@ import os
 import mp_menu
 from wxconsole_util import Value, Text
 from wx_loader import wx
-import webbrowser # handle url's in console window
+import webbrowser # Support opening URLs from console window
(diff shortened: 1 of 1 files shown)


tridge:
hi Samuel, can we move the import of webbrowser into the function that uses it? I am concerned it may not be installed on all systems, and this would break all use of the console if it isn't

SamuelDudley:
Hi Tridge,
I agree with your comments and have moved the import to the event handler as requested. In the event handler I have allowed webbrowser import errors to pass by silently. 

Thanks!

tridge:
looks good, thanks! merging

PR #385, opened 2017-02-22
Condensed down mavlink version detection

stephendade:
Simplified mavlink version options into a single commandline option

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -36,6 +36,9 @@ except Exception:
 if __name__ == '__main__':
       freeze_support()
 
+#The MAVLink version being used (None, "1.0", "2.0")
+mavversion = None
+
 class MPStatus(object):
     '''hold status information about the mavproxy'''
     def __init__(self):
@@ -523,7 +526,7 @@ def process_master(m):
         sys.stdout.flush()
         return
 
-    if m.first_byte and (opts.mavversion == "" or opts.auto_protocol):
+    if m.first_byte and mavversion == False:
         m.auto_mavlink_version(s)
     msgs = m.mav.parse_buffer(s)
     if msgs:
@@ -548,7 +551,7 @@ def process_mavlink(slave):
     except socket.error:
         return
     try:
-        if slave.first_byte and (opts.mavversion == "" or opts.auto_protocol):
+        if slave.first_byte and mavversion == False:
             slave.auto_mavlink_version(buf)
         msgs = slave.mav.parse_buffer(buf)
     except mavutil.mavlink.MAVError as e:
@@ -866,7 +869,7 @@ def set_mav_version(mav10, mav20, autoProtocol, mavversion):
     '''Set the Mavlink version based on commandline options'''
     if(mav10 == True or mav20 == True or autoProtocol == True):
         print("Warning: Using depreciated --mav10, --mav20 or --auto-protocol options. Use --mavversion instead")
-    
+
     #sanity check the options
     if (mav10 == True or mav20 == True) and autoProtocol == True:
         print("Error: Can't have [--mav10, --mav20] and --auto-protocol both True")
@@ -874,21 +877,22 @@ def set_mav_version(mav10, mav20, autoProtocol, mavversion):
     if mav10 == True and mav20 == True:
         print("Error: Can't have --mav10 and --mav20 both True")
         sys.exit(1)
-    if mavversion == '1.0' and (mav20 == True or autoProtocol == True):
+    if mavversion is not None and (mav10 == True or mav20 == True or autoProtocol == True):
         print("Error: Can't use --mavversion with legacy (--mav10, --mav20 or --auto-protocol) options")
         sys.exit(1)
-    if mavversion == '2.0' and (mav10 == True or autoProtocol == True):
-        print("Error: Can't use --mavversion with legacy (--mav10, --mav20 or --auto-protocol) options") 
-        sys.exit(1)  
-    if mavversion == '' and (mav10 == True or mav20 == True):
-        print("Error: Can't use --mavversion with legacy (--mav10, --mav20 or --auto-protocol) options")
-        sys.exit(1)          
-    
-    #and set the specific mavlink version         
+
+    #and set the specific mavlink version (False = autodetect)       
     if mavversion == "1.0" or mav10 == True:
         os.environ['MAVLINK09'] = '1'
+        mavversion = "1"
+        print "Using 1.0"
     elif mavversion == "2.0" or mav20 == True:
         os.environ['MAVLINK20'] = '1'
+        mavversion = "2"
+        print "Using 2.0"
+    else:
+        mavversion = False
+        print "Using AUTO"
 
 if __name__ == '__main__':
     from optparse import OptionParser
@@ -960,7 +964,7 @@ if __name__ == '__main__':
 
     #set the Mavlink version, if required
     set_mav_version(opts.mav10, opts.mav20, opts.auto_protocol, opts.mavversion)
-    
+
     from pymavlink import mavutil, mavparm
     mavutil.set_dialect(opts.dialect)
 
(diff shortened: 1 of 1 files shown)


MAVProxy/mavproxy.py
```diff
@@ -908,10 +908,7 @@ def run_script(scriptfile):
         action='append',
         default=[],
         help='Load the specified module. Can be used multiple times, or with a comma separated list')
-    parser.add_option("--mav09", action='store_true', default=False, help="Use MAVLink protocol 0.9")
-    parser.add_option("--mav10", action='store_true', default=False, help="Use MAVLink protocol 1.0")
-    parser.add_option("--mav20", action='store_true', default=True, help="Use MAVLink protocol 2.0")
-    parser.add_option("--auto-protocol", action='store_true', default=False, help="Auto detect MAVLink protocol version")
+    parser.add_option("--mavversion", default="", help="Force MAVLink Version (1.0, 2.0). Otherwise autodetect version")
```
peterbarker: Should really use `1.0`` as the default here.

Should use choices([...]) to limit the options.

Also, it could force 0.9, so the comment's not quite right.

MAVProxy/mavproxy.py
```diff
@@ -930,9 +927,9 @@ def run_script(scriptfile):
     if os.path.exists("/usr/sbin/ModemManager"):
         print("WARNING: You should uninstall ModemManager as it conflicts with APM and Pixhawk")
 
-    if opts.mav09:
```
peterbarker: I think we should try to preserve the old arguments for a time.  There are going to be people out there with scripts using the old arguments.

I suggest a subroutine `set_mav_version` or similar, taking opts as an argument.

Passing mav10 and mavversion at the same time should be fatal.  Passing mav09/mav10 should issue a warning.

stephendade:
Fixed - I've added the old options back, with errors if they conflict with --mavversion. There's also a warning that the old options are depreciated if they are used.

MAVProxy/mavproxy.py
```diff
@@ -523,7 +523,7 @@ def process_master(m):
         sys.stdout.flush()
         return
 
-    if m.first_byte and opts.auto_protocol:
+    if m.first_byte and (opts.mavversion == "" or opts.auto_protocol):
```
peterbarker: Should use `is None`, not `== ""`; if no argument is given to MAVProxy that's the value it will have here.

MAVProxy/mavproxy.py
```diff
@@ -861,6 +861,34 @@ def run_script(scriptfile):
             mpstate.console.writeln("-> %s" % line)
         process_stdin(line)
     f.close()
+    
+def set_mav_version(mav10, mav20, autoProtocol, mavversion):
+    '''Set the Mavlink version based on commandline options'''
+    if(mav10 == True or mav20 == True or autoProtocol == True):
+        print("Warning: Using depreciated --mav10, --mav20 or --auto-protocol options. Use --mavversion instead")
+    
+    #sanity check the options
```
peterbarker: Unless you want different error messages here, this can be simplified.  Given only one of the options is set, you can increment a counter, and if that counter is >1 in the end you can throw the error.
e.g.
```
if maversion is not None:
  sources += 1
if mav10:
  sources += 1
.
.
```

MAVProxy/mavproxy.py
```diff
@@ -908,10 +936,10 @@ def run_script(scriptfile):
         action='append',
         default=[],
         help='Load the specified module. Can be used multiple times, or with a comma separated list')
-    parser.add_option("--mav09", action='store_true', default=False, help="Use MAVLink protocol 0.9")
     parser.add_option("--mav10", action='store_true', default=False, help="Use MAVLink protocol 1.0")
-    parser.add_option("--mav20", action='store_true', default=True, help="Use MAVLink protocol 2.0")
+    parser.add_option("--mav20", action='store_true', default=False, help="Use MAVLink protocol 2.0")
```
peterbarker: Sorry, I gave you a bum steer here.  Shouldn't *change* the default from `1.0` as part of this patch.  It should probably be changed, because it does cause issues defaulting to 2.0, but it shouldn't be part of this patch.

MAVProxy/mavproxy.py
```diff
@@ -930,10 +958,9 @@ def run_script(scriptfile):
     if os.path.exists("/usr/sbin/ModemManager"):
         print("WARNING: You should uninstall ModemManager as it conflicts with APM and Pixhawk")
 
-    if opts.mav09:
-        os.environ['MAVLINK09'] = '1'
-    if opts.mav20 and not opts.mav10:
-        os.environ['MAVLINK20'] = '1'
+    #set the Mavlink version, if required
+    set_mav_version(opts.mav10, opts.mav20, opts.auto_protocol, opts.mavversion)
```
peterbarker: Still think should just pass `opts` through here, but not that fussed.

MAVProxy/mavproxy.py
```diff
@@ -548,7 +548,7 @@ def process_mavlink(slave):
     except socket.error:
         return
     try:
-        if slave.first_byte and opts.auto_protocol:
+        if slave.first_byte and (opts.mavversion == "" or opts.auto_protocol):
```
peterbarker: I think by this point we should be able to check whether the version has been set in the environment.  `opts` is only in scope because Python sucks.

Yeah, you're only doing what the code currently does in terms of scope, but if the thing is going to be thoroughly tsted it should be Made Right, I think.

peterbarker:
Oh, please also look out for adding trailing whitespace; you might find an editor setting to highlight trailing whitespace.

MAVProxy/mavproxy.py
stephendade: Default is auto-detect via mavversion == None. Should (in theory) pick up the correct mavlink version without interaction from the user.

stephendade:
I've tidied up the error handling and whitespace. Should be good now

tridge:
hi @peterbarker, you ok with this now?

MAVProxy/mavproxy.py
```diff
@@ -523,7 +526,7 @@ def process_master(m):
         sys.stdout.flush()
         return
 
-    if m.first_byte and opts.auto_protocol:
+    if m.first_byte and mavversion == False:
```
peterbarker: Should be `None`

MAVProxy/mavproxy.py
```diff
@@ -548,7 +551,7 @@ def process_mavlink(slave):
     except socket.error:
         return
     try:
-        if slave.first_byte and opts.auto_protocol:
+        if slave.first_byte and mavversion == False:
```
peterbarker: Should be `None`

MAVProxy/mavproxy.py
```diff
@@ -861,6 +864,32 @@ def run_script(scriptfile):
...
+
+    #and set the specific mavlink version (False = autodetect)       
+    if mavversion == "1.0" or mav10 == True:
+        os.environ['MAVLINK09'] = '1'
+        mavversion = "1"
+    elif mavversion == "2.0" or mav20 == True:
+        os.environ['MAVLINK20'] = '1'
+        mavversion = "2"
+    else:
+        mavversion = False
```
peterbarker: Should be `None`

MAVProxy/mavproxy.py
```diff
@@ -861,6 +864,32 @@ def run_script(scriptfile):
             mpstate.console.writeln("-> %s" % line)
         process_stdin(line)
     f.close()
+    
+def set_mav_version(mav10, mav20, autoProtocol, mavversion):
+    '''Set the Mavlink version based on commandline options'''
+    if(mav10 == True or mav20 == True or autoProtocol == True):
+        print("Warning: Using depreciated --mav10, --mav20 or --auto-protocol options. Use --mavversion instead")
```
peterbarker: `s/depreciated/deprecated/`

stephendade:
Made changes as per comments by @peterbarker 
Also added global statements that I'd missed.

peterbarker:
Pushed in.

Commented out the deprecation warning for the time being; tridge likes his `--mav10` :-)

PR #391, opened 2017-03-18
add setting for graph title

kd0aij:
This adds a tab to the settings menu for displaying a graph title.
It would be nice to default the title to something like the log name, too.


peterbarker:
I've rebased this and pushed it up here: https://github.com/peterbarker/MAVProxy/tree/add_title

It works as advertised.

However, isn't the title really a per-graph thing?

I'm inclined to put this in anyway; even if we did find a way to allow a user to specify a title on the graph line it could override the `set` anyway.

Should I put this in?

kd0aij:
stale

PR #392, opened 2017-03-22
Map: add option to show mission direction arrows on map

SamuelDudley:
This PR adds optional arrows to the mission and rally paths to show the intended vehicle movement:

screenshot from 2017-03-27 17 12 45

- The arrows can be turned on by the following command: `map set showdirection true`
- The arrows on the loiter and rally circles automatically change direction when switching from CW to CCW via a negative loiter radius.

### DIFF
--- a/MAVProxy/modules/mavproxy_map/__init__.py
+++ b/MAVProxy/modules/mavproxy_map/__init__.py
@@ -49,7 +49,9 @@ class MapModule(mp_module.MPModule):
               ('showahrs3pos', int, 0),
               ('brightness', float, 1),
               ('rallycircle', bool, False),
-              ('loitercircle',bool, False)])
+              ('loitercircle',bool, False),
+              ('showdirection', bool, False)])
+        
         service='OviHybrid'
         if 'MAP_SERVICE' in os.environ:
             service = os.environ['MAP_SERVICE']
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/mavproxy_map/__init__.py
```diff
@@ -187,7 +188,19 @@ def display_waypoints(self):
 
                     if (self.map_settings.loitercircle and
                         self.module('wp').wploader.wp_is_loiter(next_list[j])):
-                        self.mpstate.map.add_object(mp_slipmap.SlipCircle('Loiter Circle %u' % (next_list[j] + 1), 'LoiterCircles', polygons[i][j], abs(loiter_rad), (255, 255, 255), 2))
+                        wp = self.module('wp').wploader.wp(next_list[j])                    
+                        if wp.command != mavutil.mavlink.MAV_CMD_NAV_LOITER_TO_ALT and wp.param3 != 0:
```
SamuelDudley: Annoyingly MAV_CMD_NAV_LOITER_TO_ALT has its radius parameter in param2 while the other loiter commands use param3...

tridge:
very nice! merged
any reason not to enable by default?

SamuelDudley:
Nope, no reason. I personally always enable loiter circles and the direction arrows as it helps me quickly visualise if everything is OK.

PR #394, opened 2017-03-23
Link: add optional human recognizable link names

SamuelDudley:
#### EDITED 07/12/2017
When using multiple master links it can become confusing (at least for me!) as to what hardware (RFD, 3G,  Microhard, VPN, Iridium, WIFI, etc..) is attached to each link number (Link 1, 2, 3, etc...). This PR adds an optional label to a link which replaces the link number in the console.

Example of adding a master link called 'RFD900x': `--master=/dev/ttyUSB0:'{"label":"RFD900x"}'`
Example of removing a link called 'RFD900x': `link remove RFD900x`
Example of adding a link called 'FTDI': `link add /dev/ttyUSB0:{"label":"FTDI"}`

_Notes_:

- You can still use the link number to remove links if you don't want to type the label.
- Typing `link list` will show the link number and associated label (if any)

console

If no json dict with a label key is passed when creating the link it will use the existing link numbering system, making the PR backwards compatible. I'm open to suggestions to improve the implementation and syntax if required.
Cheers,
Sam

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -110,7 +110,7 @@ class LinkModule(mp_module.MPModule):
             except AttributeError as e:
                 # some mav objects may not have a "signing" attribute
                 pass
-            print("link %s %s (%u packets, %.2fs delay, %u lost, %.1f%% loss%s)" % (self.link_string(master),
+            print("link %s %s (%u packets, %.2fs delay, %u lost, %.1f%% loss%s)" % (self.link_label(master),
                                                                                     status,
                                                                                     self.status.counters['MasterIn'][master.linknum],
                                                                                     linkdelay,
@@ -123,8 +123,8 @@ class LinkModule(mp_module.MPModule):
         print("%u links" % len(self.mpstate.mav_master))
         for i in range(len(self.mpstate.mav_master)):
             conn = self.mpstate.mav_master[i]
-            if conn.handle is not None:
-                print("%u (%s): %s" % (i, conn.handle, conn.address))
+            if conn.label is not None:
+                print("%u (%s): %s" % (i, conn.label, conn.address))
             else:
                 print("%u: %s" % (i, conn.address))
 
@@ -148,11 +148,11 @@ class LinkModule(mp_module.MPModule):
             conn.mav.set_send_callback(self.master_send_callback, conn)
         conn.linknum = len(self.mpstate.mav_master)
         try:
-            handle = link_components[1]
-            handle = handle.rstrip(')')
-            conn.handle = handle
+            label = link_components[1]
+            label = label.rstrip(')')
+            conn.label = label
         except:
-            conn.handle = None
+            conn.label = None
         conn.linkerror = False
         conn.link_delayed = False
         conn.last_heartbeat = 0
@@ -186,7 +186,7 @@ class LinkModule(mp_module.MPModule):
             return
         for i in range(len(self.mpstate.mav_master)):
             conn = self.mpstate.mav_master[i]
-            if str(i) == device or conn.address == device or conn.handle == device:
+            if str(i) == device or conn.address == device or conn.label == device:
                 print("Removing link %s" % conn.address)
                 try:
                     try:
@@ -327,7 +327,7 @@ class LinkModule(mp_module.MPModule):
         if mtype in activityPackets:
             if master.linkerror:
                 master.linkerror = False
-                self.say("link %s OK" % (self.link_string(master)))
+                self.say("link %s OK" % (self.link_label(master)))
             self.status.last_message = time.time()
             master.last_message = self.status.last_message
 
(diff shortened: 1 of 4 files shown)


tridge:
nice idea!
Could you put the repetitive code (the if statements) in a common place, perhaps in mp_module.py?

SamuelDudley:
@tridge No problem! I have removed the duplicated code and implemented a static method in mp_module.py. As the static method functionality is only used to access the function from mavproxy.py, perhaps the if statement should remain here? or maybe there is a cleaner way to obtain the same functionality without requiring a static method?

Let me know if your thoughts and if you would like any more changes.
Cheers,
Sam

MAVProxy/modules/mavproxy_link.py
```diff
@@ -110,23 +110,29 @@ def show_link(self):
...
                                                                                     master.packet_loss(),
                                                                                     sign_string))
+
     def cmd_link_list(self):
         '''list links'''
         print("%u links" % len(self.mpstate.mav_master))
         for i in range(len(self.mpstate.mav_master)):
             conn = self.mpstate.mav_master[i]
-            print("%u: %s" % (i, conn.address))
+            if conn.handle is not None:
```
SamuelDudley: I have left this as an if statement because I want to be explicit about what data is being (or not being) displayed

SamuelDudley:
I feel that this is good to merge. Please let me know if you would like any other changes.
Cheers!

MAVProxy/modules/mavproxy_link.py
```diff
@@ -110,23 +110,29 @@ def show_link(self):
...
+                print("%u (%s): %s" % (i, conn.handle, conn.address))
+            else:
+                print("%u: %s" % (i, conn.address))
 
-    def link_add(self, device):
+    def link_add(self, descriptor):
         '''add new link'''
         try:
+            link_components = descriptor.split('(')
```
peterbarker: I wonder if we should do something more comprehensive in terms of adding attributes to links here?  Perhaps name-value pairs?

At the moment you're putting stuff in prens which is being dealt with by MAVProxy.  Another option would be to extend the link strings which pymavlink understands and have it remember the link names.  So your link might be:

`link add /dev/ttyUSB0:label=rfd900z`

MAVProxy/modules/mavproxy_link.py
```diff
@@ -137,11 +143,16 @@ def link_add(self, device):
...
             conn.set_rtscts(True)
-        conn.linknum = len(self.mpstate.mav_master)
         conn.mav.set_callback(self.master_callback, conn)
         if hasattr(conn.mav, 'set_send_callback'):
             conn.mav.set_send_callback(self.master_send_callback, conn)
         conn.linknum = len(self.mpstate.mav_master)
+        try:
+            handle = link_components[1]
+            handle = handle.rstrip(')')
+            conn.handle = handle
```
peterbarker: Perhaps "label" would be better than "handle"; less confusing with a "file handle" that way...

SamuelDudley:
Hi,
Thanks for the feedback :+1: I agree with the re-naming from "handle" to "label".
I'm on the fence about where the string parsing should be located but I'll move to the name value pair arrangement. If we like it as an idea we can pull the parsing from here and insert it into mavutil.py

SamuelDudley:
@peterbarker Please have a look at the changes when you have some time and let me know your thoughts. Cheers.

MAVProxy/modules/mavproxy_link.py
```diff
@@ -147,6 +161,11 @@ def link_add(self, device):
         conn.last_heartbeat = 0
         conn.last_message = 0
         conn.highest_msec = 0
+        for attr in optional_attributes:
+            if not hasattr(conn, attr):
```
SamuelDudley: There might be a reason you would want to do this... possibly to do something funky like set target_system here on a per link basis. However to do this we would need to convert the 'value' string into the correct type (int in the case of target_system). This could be done by passing in a single dict of key, value pairs and parsing the dict... /dev/ttyUSB0:{"target_system":123,"label":"link_for_system_123"}

peterbarker:
On Sun, 9 Jul 2017, Samuel Dudley wrote:

> Thanks for the feedback 👍 I agree with the re-naming from "handle" to
> "label".
> I'm on the fence about where the string parsing should be located but I'll
> move to the name value pair arrangement. If we like it as an idea we can
> pull the parsing from here and insert it into mavutil.py

I like it!

Perhaps @stephendade could comment?

stephendade:
Works nicely on Windows! I vote for merging.

SamuelDudley:
Hi, just bumping this PR. I would still very much like to see this merged in the future (when time permits). I am still willing to make any further changes as requested in order to have it merged.
Thanks again,
Sam

SamuelDudley:
modified to support json format and rebased off master

SamuelDudley:
merged with #480 
Thanks to everyone who helped get this in!

PR #395, opened 2017-03-25
GPS_INPUT module

jaxxzer:
This module allows an external application that is determining the position of the vehicle to connect to mavproxy via udp, and forward the position information to the autopilot via GPS_INPUT messages. The information is transported between the external application and MAVProxy via JSON encoded python dictionaries.


tridge:
looks good, thanks!

PR #404, opened 2017-04-11
Param: Support non-upper case param names when using param help

SamuelDudley:
Currently "param help gps_type" will error. This PR fixes that issue
by converting the param name to upper case prior to searching the help
tree.


tridge:
thanks Samuel!

PR #405, opened 2017-04-14
DGPS Module: Switching DGPS injection to use GPS_RTCM_DATA message.

njoubert:
Using GPS_RTCM_DATA messages support fragmenting and reassembling up to 720 bytes of GPS-related data that will be injected into the onboard GPSes. Can be used for RTCM, SBP, UBX, or what-have-you.

### DIFF
--- a/MAVProxy/modules/mavproxy_DGPS.py
+++ b/MAVProxy/modules/mavproxy_DGPS.py
@@ -7,39 +7,86 @@ import socket, errno
 from pymavlink import mavutil
 from MAVProxy.modules.lib import mp_module
 
+import socket, errno
+from pymavlink import mavutil
+from MAVProxy.modules.lib import mp_module
+
 class DGPSModule(mp_module.MPModule):
     def __init__(self, mpstate):
-        super(DGPSModule, self).__init__(mpstate, "DGPS", "DGPS injection support")
+        super(DGPSModule, self).__init__(mpstate, "DGPS", "DGPS injection support for SBP/RTCP/UBC")
         self.portnum = 13320
         self.port = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
         self.port.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         self.port.bind(("127.0.0.1", self.portnum))
         mavutil.set_close_on_exec(self.port.fileno())
         self.port.setblocking(0)
-        print "Listening for DGPS packets on UDP://%s:%s" % ("127.0.0.1", self.portnum)
+        self.inject_seq_nr = 0
+        print "DGPS: Listening for RTCM packets on UDP://%s:%s" % ("127.0.0.1", self.portnum)
+    
+    def send_rtcm_msg(self, data):
+        msglen = 180;
+        
+        if (len(data) > msglen * 4):
+            print "DGPS: Message too large", len(data)
+            return
+        
+        # How many messages will we send?
+        msgs = 0
+        if (len(data) % msglen == 0):
+            msgs = len(data) / msglen
+        else:
+            msgs = (len(data) / msglen) + 1
+        
+        for a in range(0, msgs):
+            
+            flags = 0
+            
+            # Set the fragment flag if we're sending more than 1 packet.
+            if (msgs) > 1:
+                flags = 1
+            
+            # Set the ID of this fragment
+            flags |= (a & 0x3) << 1
+            
+            # Set an overall sequence number
+            flags |= (self.inject_seq_nr & 0x1f) << 3
+            
+            
+            amount = min(len(data) - a * msglen, msglen)
+            datachunk = data[a*msglen : a*msglen + amount]
+            
+            self.master.mav.gps_rtcm_data_send(
+                flags,
+                len(datachunk),
+                bytearray(datachunk.ljust(180, '\0')))
+        
+        # Send a terminal 0-length message if we sent 2 or 3 exactly-full messages.     
+        if (msgs < 4) and (len(data) % msglen == 0) and (len(data) > msglen):
+            flags = 1 | (msgs & 0x3)  << 1 | (self.inject_seq_nr & 0x1f) << 3
+            self.master.mav.gps_rtcm_data_send(
+                flags,
+                0,
+                bytearray("".ljust(180, '\0')))
+            
+        self.inject_seq_nr += 1
+
+
 
     def idle_task(self):
         '''called in idle time'''
         try:
-            data = self.port.recv(200)
+            data = self.port.recv(1024) # Attempt to read up to 1024 bytes.
         except socket.error as e:
             if e.errno in [ errno.EAGAIN, errno.EWOULDBLOCK ]:
                 return
             raise
-        if len(data) > 110:
-            print("DGPS data too large: %u bytes" % len(data))
-            return
         try:
-
-            self.master.mav.gps_inject_data_send(
-                self.target_system,
-                self.target_component,
-                len(data),
-                bytearray(data.ljust(110, '\0')))
+            self.send_rtcm_msg(data)
 
         except Exception,e:
-            print "DGPS Failed:", e
+            print "DGPS: GPS Inject Failed:", e
 
 def init(mpstate):
     '''initialise module'''
     return DGPSModule(mpstate)
+


MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -7,39 +7,75 @@
...
+    
+    def send_rtcm_msg(self, data):
+        msglen = 180;
+
+        if (len(data) > msglen * 4):
+            print "DGPS: Message too large", len(data)
+            return
+
+        msgs = (len(data) / msglen) + 1
+        if (len(data) % msglen == 0):
```
WickedShell: In order to do it this way (which is what the MP implentation works), you need to send an extra packet if the message didn't consist of 4 fragments. See: https://github.com/ArduPilot/MissionPlanner/blob/master/Mavlink/MAVLinkInterface.cs#L2586

MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -7,39 +7,75 @@
...
+            msgs = len(data) / msglen
+
+        for a in range(0, msgs):
+
+            flags = 0
+
+            if (msgs) > 0:
+                flags = 1
+
+            flags |= (a & 0x3) << 1
```
WickedShell: Probably worth adding a comment that says this is setting the fragment ID, the one below is sequence and the one above is the is fragmented flag.

If the packet isn't fragmented you can opt to not set any of the bits in the packet, which might be worthwhile, I'm not sure if PX4 looks at the fragment bit or not (and if they look at the bit vs looking at the whole flags field)

MAVProxy/modules/mavproxy_DGPS.py
njoubert: Ah ok. 
It looks to me like MP has a bug: it does not increment the sequence ID of the terminal packet. Am I misreading that?

WickedShell: Sequence ID shouldn't change on the terminating one, but I think your correct that the fragment ID isn't updated...

MAVProxy/modules/mavproxy_DGPS.py
njoubert: Cool, I'll opt to do that.

MAVProxy/modules/mavproxy_DGPS.py
njoubert: Ah. Looking at https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_GPS/AP_GPS.cpp#L1020
suggests to me that, if the len is zero, it doesn't matter what the sequence ID is. Is that correct?

njoubert:
@WickedShell I think I got it now. The fragment ID of the terminal packet doesn't matter, so I'm reusing flags from before. This code is currently sending SBP up to my Piksi-on-Iris :)

WickedShell:
@njoubert why are you saying the final frag ID doesn't matter? It definitely does, and if it didn't then its all broken and not doing what its supposed to...

WickedShell:
We only inject if https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_GPS/AP_GPS.cpp#L1037 is true, which is only true once we got the final frag id... (or if we have all 4 fragments)

WickedShell:
If you reuse the last ID then it's possible to accept the packet with only the last full payload spot, which is incorrect, you need to set that frag id.

njoubert:
I'm specifically talking about the terminal packet: Notice that MP reuses the fragment id for the terminal packet! I claim it's okay to have the same fragment Id for the *terminal* packet as the packet right before it, because this line won't copy anything since len is zero: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_GPS/AP_GPS.cpp#L1020

njoubert:
MP, here, reuses the same fragment Id as the previous packet: https://github.com/ArduPilot/MissionPlanner/blob/master/Mavlink/MAVLinkInterface.cs#L2587

WickedShell:
That's insufficent. Lets say you have 2 full length fragments, fragment 1
arrives first, then fragment 3 (the 0 length one)  because 3 (the 0 length
packet) reused the ID from 2 we will flush the buffer at that point, and
that will be without having fragment 2. The MP side is an implementation
bug.

njoubert:
Ah cool, so we found a bug in MP :) I was worried about that. Shall I file a bug report, or do you want to?

Check my logic: if we send 4 fragments, then we don't need to send a terminal packet, since that last check will pass.

WickedShell:
I sent @meee1 a message directly, but it's probably worth opening an issue.

Correct on a 4 packet message there is no terminating packet needed (if you look that is special cased here: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_GPS/AP_GPS.cpp#L1029 )

meee1:
fixed in MP :)

njoubert:
@WickedShell ping!

MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -7,39 +7,85 @@
...
         try:
-            data = self.port.recv(200)
+            data = self.port.recv(1024) # Attempt to read up to 1024 bytes.
         except socket.error as e:
             if e.errno in [ errno.EAGAIN, errno.EWOULDBLOCK ]:
                 return
             raise
-        if len(data) > 110:
-            print("DGPS data too large: %u bytes" % len(data))
+        if len(data) > 720:
```
WickedShell: I'd remove this check here, send_rtcm_msg is actually doing size checking for you already (line 29), and it will always know best what size can be sent out :)

MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -7,39 +7,85 @@
...
+        print "DGPS: Listening for RTCM packets on UDP://%s:%s" % ("127.0.0.1", self.portnum)
+    
+    def send_rtcm_msg(self, data):
+        msglen = 180;
+        
+        if (len(data) > msglen * 4):
+            print "DGPS: Message too large", len(data)
+            return
+        
+        msgs = (len(data) / msglen) + 1
```
WickedShell: Can't this hit 5? (720 / 180) + 1 == 4 + 1. This would be a problem as you don't want to actually shift out a 5th element, as it could step on the first one if there was a sequence problem. With that being said I think if you just take the minimum of this equation or 4 you would be fine.

njoubert: The next 2 lines should take care of that, since (720 % 180 == 0) then msgs = (720 / 180 ) = 4

MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -7,39 +7,85 @@
...
+            
+            amount = min(len(data) - a * msglen, msglen)
+            datachunk = data[a*msglen : a*msglen + amount]
+            
+            self.master.mav.gps_rtcm_data_send(
+                flags,
+                len(datachunk),
+                bytearray(datachunk.ljust(180, '\0')))
+            
+        if (msgs < 4) and (len(data) % msglen == 0) and (len(data) > msglen):
```
WickedShell: Unless I'm missing something haven't you already done this by this point? (By adding 1 to the number of messages earlier) I think this whole if is unneeded. But I'm still waking up and python's not my thing :)

MAVProxy/modules/mavproxy_DGPS.py
WickedShell: I definitely need more coffee....  That's totally fine :)

MAVProxy/modules/mavproxy_DGPS.py
WickedShell: Given the line I missed above this would be fine :P (If you did my min approach and dropped your  % == 0 line you could drop this to, but I don't think thats a big deal, at that point it's all style.

njoubert: Ah, no, because the above line doesn't send a terminal null-length message. I have to handle this case separately, since it only ocmes up if we sent 2 or 3 messages, each with exactly 180 bytes of data. 

The above line you mentioned doesn't add a message if we filled up packets exactly.

njoubert:
I've been flying this for the last 2 weeks and it's working great, can we merge it?

tridge:
Thanks @njoubert, and thanks also to @WickedShell for the review!

PR #408, opened 2017-04-19
RTK GPS data injection implementation using TCP client and…

amilcarlucas:
… GPS_RTCM_DATA MAVLink messages


WickedShell:
Is there a reason this doesn't instead leverage #408 These look to be doing the same thing? (and duplicating a lot of logic?)

tridge:
@amilcarlucas can we combine it with the existing functionality?

amilcarlucas:
@tridge yes, please do

amilcarlucas:
This is not as easy as I originally thought.
The DGPS module is a UDP server and this one is a TCP client .... both protocol and s´master/ server are different. I have no idea how to provide configuration parameters into a module so that all combinations would be possible.
Furthermore on a perfect world the client would get configurable address and port information.

For now the RTCM module is the only working solution for Emlid Reach RTK modules. And the DGPS for Drotec and Swiftnav modules.

amilcarlucas:
We started to work on a more flexible solution to this. We will add some new commits to this soon.

geniusxie1:
@amilcarlucas Thanks very much for creating this module and I have to say it is working very well with Reach RTK and you are right, it is the only module that support Reach RTK. I am just wondering if it is easy enough to have a TCP Server version of this module or a quick tweak on the existing code which can support TCP Server listens on a port and make Reach RTK as the RTCM client to feed correction to this module. Thanks very much and looking forward to your next updates as well! Much appreciated!

amilcarlucas:
Updated it. To use it as UDP client:
`module load RTCM
RTCM add udp:127.0.0.1:6000`
For TCP/IP client:
`module load RTCM
RTCM add tcp:127.0.0.1:6000`

You should of course replace 127.0.0.1 with your desired IP address and 6000 with your desired port.

geniusxie1:
@amilcarlucas Wow, this is a quick update and it works. However, I found out that Reach RTK is configured to send RTCM correction messages as a client through only TCP protocal. I just tested the updated module and seems the UDP client is not receiving the correction. Is it possible to add another option to add a TCP server that always listens in the background? Much appreciated!

stw2nf:
do you have a build that I could test/run? Really interested in this feature. we have a Emlid Reach M2 we would like to automate RTK corrections with via Rpi, but having troubles rebasing this branch on master and actually testing functionality

PR #411, opened 2017-05-04
Map: Correct and clean rover icons

khancyr:
The size was incorrect leading to weird display. I also repaint the frame to have more uniform color display


gmorph:
Tested and working.  Please merge in.

PR #412, opened 2017-05-07
MAVExplorer - spaces now allowed in log file name and path

stephendade:


MAVProxy/tools/MAVExplorer.py
```diff
@@ -428,9 +429,13 @@ def cmd_param(args):
 def cmd_loadfile(args):
     '''callback from menu to load a log file'''
     if len(args) != 1:
-        print "Error loading file"
+        fileargs = " ".join(args)
```
peterbarker: Should we allow quoting of the filename here instead?  Use shlex to break the line up for us?  Taking logically separate arguments and squishing them together is a little awkward.

MAVProxy/tools/MAVExplorer.py
```diff
@@ -446,6 +451,14 @@ def loadfile(args):
     load_graphs()
     setup_menus()
 
+def shlex_quotes(value):
```
peterbarker: Perhaps `shlex_split` instead?

stephendade: Yep, will fix that

MAVProxy/tools/MAVExplorer.py
stephendade: cmd_loadfile should only expect 1 argument anyway - the name of the logfile to be loaded. Not sure why multiple arguments should be expected here.

tridge:
applied with rebase fixup, thanks!

PR #414, opened 2017-05-17
Allow decimal altitude to the guided command

h4writer:
As far as I can see there is no reason why it doesn't accept decimal numbers.
I tested it locally with the simulator and it works.


amilcarlucas:
ping :)

h4writer:
Not sure if I need to do something extra here? This patch is ready to get merged.

peterbarker:
Merged, thanks!

Apologies for the slow response.

PR #421, opened 2017-06-13
Map - Fix error on brightness change

SamuelDudley:
When changing the brightness of the map (multiplying self.map_img and state.brightness), the resulting image array was being cast to dtype float64. This change ensures that the map_img remains a uint8 dtype, fixing the CV2 error.

### DIFF
--- a/MAVProxy/modules/mavproxy_map/mp_slipmap.py
+++ b/MAVProxy/modules/mavproxy_map/mp_slipmap.py
@@ -34,7 +34,7 @@ class MPSlipMap():
                  service="MicrosoftSat",
                  max_zoom=19,
                  debug=False,
-                 brightness=1.0,
+                 brightness=0,
                  elevation=False,
                  download=True,
                  show_flightmode_legend=True):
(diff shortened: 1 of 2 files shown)


SamuelDudley:
_Note:_ this error exists when using opencv 3.2 (python bindings built from source) and numpy == 1.13.0 (current pip version), there does not seem to be an issue with older versions of opencv.

peterbarker:
Here's what I get on brightness-up:

feeling-blue

It does fix the exception, 'though!

SamuelDudley:
Good catch, sorry! Yes any brightness multiplier greater than 1 leads to an increasingly psychedelic looking map... So brightness >1 is currently causing this (you can decrease and then increase an it will be fine).

SamuelDudley:
So it looks like the issue is that the map is a uint8 dtype and the multiplication with a scalar causes roll over in some of the pixel [r,g,b] values depending on their initial value.

SamuelDudley:
The following code works if brightness is first mapped to a 0 - 255 range
``
self.map_img = np.where((255 - self.map_img) < brightness,255,self.map_img+brightness)
``

SamuelDudley:
Okay, in my testing this new set of commits fixes the brightness error for both increasing and decreasing brightness. I have changed the way the brightness value is interpreted with 0 = no change, -255 is a completely black image and 255 is completely white. So the possible range for the brightness value is now [-255, 255]. Feedback welcome!

peterbarker:
On Fri, 20 Oct 2017, Samuel Dudley wrote:

> Good catch, sorry! Yes any brightness multiplier greater than 1 leads to an
> increasingly psychedelic looking map... So brightness >1 is currently

Let's declare it a Feature! :-)

amilcarlucas:
This one seams to fix some issues with #450

tridge:
applied, thanks!

PR #425, opened 2017-07-15
General speech capabilities for MAVLink messages

njoubert:
This PR adds general speech capabilties to MAVProxy's CLI.
Any MAVLink message entry can be announced in multiple ways.

Three types of announcements are supported:
- Periodic (the default)
- Conditional. Announces when value leaves expected range.
- On Change. Announces when a value changes.

Individual announces as well as the entire speech engine can be enabled or disables.
Individual announces can be added on-the-fly or removed.

Configuration happens through the speech CLI command.
`usage: speech <say|list|enable|disable|every|add|remove>`

**Suggested usage:** Add initialization commands to .mavinit.scr

**Example commands:**
    `module load mavproxy_speech`
    `speech add ALTITUDE altitude_relative -r 30 -n "altitude"`
    `speech add VFR_HUD airspeed -r 10 -f %0.0f -s 1.0 -n "airspeed"`

**Known limitations:**
- The current backend is blocking. 
- There are no unit conversions built-it. 
- There's no sense of priority between messages.

**Attribution:**
This work is done under the gracious support of FlightWave, where I'm using these tools during our flight testing.

### DIFF
--- a/MAVProxy/modules/mavproxy_speech.py
+++ b/MAVProxy/modules/mavproxy_speech.py
@@ -16,20 +16,27 @@ class SpeechAnnounce(object):
         self.lastTime = 0
 
     def match(self, msg):
-        return self.enabled and (msg.get_type() == self.msg_type)
+        '''Returns true if this announce is ready to handle the given message'''
+        return (msg.get_type() == self.msg_type)
+
+    def dispatch(self, msg, backend):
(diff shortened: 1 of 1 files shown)


njoubert:
I found a few bugs and improvements that i'll address first before sending this again.

PR #427, opened 2017-07-18
Fix Live Graph for macOSX

bugobliterator:


njoubert:
I tested this PR, and it works great! Thanks @bugobliterator !

peterbarker:
I've tested this on Linux - no worries.

I'm a little concerned about that `set_facecolor` thing; why is that present here?

@stephendade Does this work OK on Windows with the `set_facecolor` change?

stephendade:
@peterbarker - all good on Windows

bugobliterator:
@peterbarker the `set_facecolor` change is done because MAVProxy was throwing warning of deprecation with matplotlib 2.0.0.

peterbarker:
@bugobliterator OK, without the set_facecolor, this patch is a shoe-in.

With it, I'm leery of breaking people's graphing if they are using older matplotlib.

Will things work if we use the deprecated function?

njoubert:
@bugobliterator it'd be great if you can drop in your MAVExplorer patches into the same PR. As a primarily-mac user, it's **really powerful** to have all the utilities available on Mac!

bugobliterator:
@peterbarker resurrecting this PR, I have made the change you requested.

peterbarker:
Merged, thanks!

PR #428, opened 2017-07-18
Support General Speech Announcements

njoubert:
This PR adds general speech capabilties to MAVProxy's CLI.
Any MAVLink message entry can be announced in multiple ways.

Three types of announcements are supported:
- Periodic (the default)
- Conditional. Announces when value leaves expected range.
- On Change. Announces when a value changes.

Individual announces as well as the entire speech engine can be enabled or disables.
Individual announces can be added on-the-fly or removed.

Configuration happens through the speech CLI command.
`usage: speech <say|list|enable|disable|every|add|remove>`

**Suggested usage:** Add initialization commands to .mavinit.scr

**Example commands:**
    `module load mavproxy_speech`
    `speech add ALTITUDE.altitude_relative -r 30 -n "altitude"`
    `speech add VFR_HUD.airspeed -r 10 -f %0.0f -s 1.0 -n "airspeed"`

**Known limitations:**
- The current backend is blocking. 
- There are no unit conversions built-it. 
- There's no sense of priority between messages.

**Attribution:**
This work is done under the gracious support of FlightWave, where I'm using these tools during our flight testing.


njoubert:
@peterbarker @bugobliterator This is a completion of the previous speech PR I submitted. 

We now have a generic speech system, in a similar spirit to module_graph - We can announce arbitrary mavlink data periodically, on change, or if outside some range.

peterbarker:
On Tue, 18 Jul 2017, Niels Joubert wrote:

> This PR adds general speech capabilties to MAVProxy's CLI.
> Any MAVLink message entry can be announced in multiple ways.

Yowsers!  This is cool - tridge and I discussed this a *long* time ago :-)

> module load mavproxy_speech
> speech add ALTITUDE.altitude_relative -r 30 -n "altitude"
> speech add VFR_HUD.airspeed -r 10 -f %0.0f -s 1.0 -n "airspeed"

syntax-wise I think it would be more in-line with MAVProxy-ish things 
to do:
speech add ALTITUDE.altitude_relative repeat 30 name "altitude"

We may even be better off going name-value pairs:
speech add ALTITUDE.altitude_relative repeat=30 name="altitude"

I went with name-value pairs for the firmware module and I think it's a 
reasonable approach....


yours,
-- 
Peter Barker                          |   Programmer,Sysadmin,Geek.
pbarker@barker.dropbear.id.au	      |   You need a bigger hammer.
:: It's a hack! Expect underscores! - Nigel Williams

njoubert:
Ah, thanks Peter!

When we do flight testing, we have a powered PA speaker (QSC K10) announcing flight values, super useful, but I needed more flexibility.

Good feedback. I like the name-value pairs a lot. That should also give me a better way to pass named parameters to functions. Adding now.

njoubert:
Updated to support name-value pairs. This nicely cleaned up the command parsing implementation as well.

SamuelDudley:
This will be __very__ useful! :+1:

SamuelDudley:
Just trying it now with the espeak backend now... One thing that I have noted is that some of the commands are currently case sensitive e.g.: `speech enable all` fails due to case mismatch (`"all" != "ALL"`) . If case does not truly matter you could add a `foo.lower()` to the user input to remove the case issues

njoubert:
@SamuelDudley that's a good point, there's no reason why case should matter. This is a bit of a corner case, but it's user-friendy to ignore case. I'll update.

How are you finding it otherwise?

stephendade:
Very nice! Unfortunately, there's no python bindings for espeak on Windows. So no text-to-speech for Windows users.

peterbarker:
On Wed, 19 Jul 2017, Stephen wrote:

> Very nice! Unfortunately, there's no python bindings for espeak on Windows.
> So no text-to-speech for Windows users.

https://pypi.python.org/pypi/speech/0.5.1

We already do multiple backends :-)

njoubert:
Indeed, all the previous backends are still supported!

stephendade:
Ahh, that works nicely. Works nicely on my Windows box now

peterbarker:
On Wed, 19 Jul 2017, Stephen wrote:

> Ahh, that works nicely. Works nicely on my Windows box now

Wow!

So... going to one-up the rest of "alternative OS" users and do speech 
recognition now?  :-)

SamuelDudley:
@njoubert I really like it! We often have a person in the GCS calling airspeed and altitude over a radio to the flight line, but sometimes lack the extra body to do so. I can see myself scripting this to announce periodically during takeoff, climb and landing phases and setting announcement thresholds for key values for the entire flight. It will take a bit of tweaking with the timings so that its not too 'chatty' but overall its an awesome feature.

njoubert:
@peterbarker anything I can do to move this along and get it merged?

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -1,26 +1,209 @@
...
+'''
 
-import time, os
+Speech Engine for audio announcements of MAVLink messages during flight.
+
+Suggested usage: Add initialization commands to .mavinit.scr
+
+Example commands:
+    module load speech
+    speech add VFR_HUD.airspeed onchange every=3.0 as=airspeed format=%0.0f threshold=2.0
```
peterbarker: Why not dispense with `as` and require the user to put that in the format?

You could default format to "airspeed %0.0f" (replacing airspeed with whatever is after the "." in the nominated field...

Allowing `VFR_HUD.airspeed` to be a Python expression would be nice; we already have code floating around which lets you pull out field-y looking things from expressions.

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -1,26 +1,209 @@
...
+        if (now - self.lastTime > self.sayEveryS):
+            self.lastTime = now
+            value = self.msg_entry_to_text(msg)
+            backend(self.to_speech_text(value))
+
+    def to_speech_text(self, value):
+        '''Returns the text announced before the value'''
+        return "%s %s" % (self.get_speech_name(), self.convert(value))
+
+    def get_speech_name(self):
```
peterbarker: If "as" was removed you'd want a "name" field, I guess.

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -1,26 +1,209 @@
...
+            self.entry_name, 
+            self.get_speech_name(), 
+            self.sayEveryS,
+            self.scale,
+            self.formatstr)
+
+class ConditionalSpeechAnnounce(PeriodicSpeechAnnounce):
+    '''
+    A SpeechAnnounce subclass that provides rate-limited announcements
+    only when a mavlink value is outside an expected range.
```
peterbarker: This is a subset of possible limitations you may want on announcements.

Why not allow a python expression to be supplied to determine if an announcement is warranted?

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -1,26 +1,209 @@
...
+
+    def __str__(self):
+        s = super(OnChangeSpeechAnnounce, self).__str__()
+        return "%s when value changes by more than %f" % (s, self.threshold)  
+
+import threading
+import Queue
+import subprocess
+
+class SaySpeechBackend(threading.Thread):
```
peterbarker: This is a separate change to the "conditional announcement" stuff etc, right?

Could we get this in a separate patch so it might get merged earlier?

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -1,26 +1,209 @@
...
+        self.msgs_to_announce = []
+
         self.old_mpstate_say_function = self.mpstate.functions.say
-        self.mpstate.functions.say = self.say
+        self.mpstate.functions.say = self.print_and_say
+        self.enabled_for = "ALL"
+        self.say_speech_backend = None
+
         try:
-            self.settings.set('speech', 1)
```
peterbarker: This looks like a bugfix which should at least be in a seperate patch?

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -1,26 +1,209 @@
...
+            self.settings.get('speech')
         except AttributeError:
             self.settings.append(('speech', int, 1))
         try:
             self.settings.set('speech_voice', '')
         except AttributeError:
             self.settings.append(('speech_voice', str, ''))
         self.kill_speech_dispatcher()
-        for (backend_name,backend) in [("speechd",self.say_speechd), ("espeak",self.say_espeak), ("speech", self.say_speech), ("say", self.say_say)]:
+        for (backend_name,backend) in [("speechd",self.say_using_speechd), ("espeak",self.say_using_espeak), ("speech", self.say_using_speech), ("say", self.say_using_say)]:
```
peterbarker: These renames are good, but should be in a separate commit.

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -68,50 +255,156 @@ def say_speechd(self, text, priority='important'):
...
+            cmd_usage += "  Say arbitrary text. Useful for testing"
             if len(args) < 2:
-                print("usage: speech say <text to say>")
+                print(cmd_usage)
                 return
             self.say(" ".join(args[1::]))
+        elif args[0] == "enable":
+            cmd_usage  = "usage: speech enable (ALL|MODE|MAV|NONE)\n"
+            cmd_usage += "  ALL  - reports all console and mavlink messages\n"
+            cmd_usage += "  MODE - reports only mode changes and mavlink messages\n"
```
peterbarker: Probably want to make this a bitmask-equivalent rather than modal.

MAVProxy/modules/mavproxy_speech.py
```diff
@@ -68,50 +255,156 @@ def say_speechd(self, text, priority='important'):
...
+                    print cmd_usage
+        elif args[0] == "list":
+            print "Speech module will announce:", self.enabled_for
+            print "Speech module knows how to report: "
+            for idx, announcement in enumerate(self.msgs_to_announce):
+                print "#%d: %s" % (idx, announcement)
+        elif args[0] == "add":
+            usage =  "usage: speech\n"
+            usage += "add <msg_type>.<msg_entry> periodic    (every=<seconds>) (as=<say_as>) (scale=<scale>) (format=\"<formatstr>\")\n"
+            usage += "add <msg_type>.<msg_entry> conditional (every=<seconds>) (as=<say_as>) (scale=<scale>) (format=\"<formatstr>\") (max=<val>) (min=<val>)\n"
```
peterbarker: If my suggestion for allowing announcements of expressions (based on whether an expressoin returns true....) is accepted, then conditional and onchange can probably be merged into one.  Old value and new value would need to be supplied to the expression, and some method for state storage (we already do something like this using `m[]`.

Doing this would allow reading out deltas rather than absolute values, for example.

peterbarker:
On Thu, 20 Jul 2017, Niels Joubert wrote:

> @peterbarker anything I can do to move this along and get it merged?

You could nudge me, and get a review.

So that worked.

Most important thing is to get the syntax right.  Don't want to annoy 
tridge by breaking his .mavinit.scr when it is decided the command syntax 
needs fixing ;-)

peterbarker:
@njoubert https://github.com/ArduPilot/MAVProxy/pull/428 - I think the desired functionality in that proposed module could be accomplished with the changes I mooted here.

auturgy:
Is this pr still live?

amilcarlucas:
This needs a rebase

PR #430, opened 2017-07-21
Fixing a small omission in console: Add battery voltage and current display.

njoubert:
I've been using the gui console extensively during flight testing.

The only missing value is current battery information!

This four-line PR adds current and voltage display to the gui console.

### DIFF
--- a/MAVProxy/modules/mavproxy_console.py
+++ b/MAVProxy/modules/mavproxy_console.py
@@ -55,6 +55,8 @@ class ConsoleModule(mp_module.MPModule):
         mpstate.console.set_status('AspdError', 'AspdError --', row=3)
         mpstate.console.set_status('FlightTime', 'FlightTime --', row=3)
         mpstate.console.set_status('ETR', 'ETR --', row=3)
+        mpstate.console.set_status('Vbat', 'Vbat: --', row=3)
+        mpstate.console.set_status('Abat', 'Abat: --', row=3)
 
         mpstate.console.ElevationMap = mp_elevation.ElevationModel()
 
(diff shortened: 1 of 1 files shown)


njoubert:
@peterbarker this is a pretty small and innocuous change but i'd like to get it in before our next flyday! Mind doing a quick merge?

peterbarker:
@njoubert Sorry for not providing quick feedback on this.  There are two reasons I haven't merged this.

The first is that it's a UI change which will change everybody's console.  You've inserted the items in the middle of the lists, meaning that everything else gets pushed to the right.  Already some of the entries are scrolled off to the right in the default console screen, so you'll be making more disappear :-)

The second is that a lot of people don't ever calibrate these sensors - or the vehicle doesn't even have them.  So we'd be adding interface clutter for no good reason for them.  This bugs me - but it wouldn't stop me merging it :-)

What tridge and I have previously talked about is being able to configure these using your `.mavinit.scr`, so you can remove irrelevant entries (e.g. ARSP for your multicopters).  That's a pretty ambitious project, of course :-)

I think I'd be happy to merge these so long as the entries were tacked onto the end of the line, so people would have to be quite observant to notice the change :-)

Ah.  One last thing.  The idiom for monitoring battery in MAVProxy appears to really be `bat`; you give MAVProxy the cellcount (e.g. `set numcells 4`) and then `bat` will give you charge remaining in terms of 1S e.g. 3.92V.  Perhaps that could also be added?  This saves doing calculations when you're working with multiple different vehicle types with different cell counts.

njoubert:
OK, awesome, I can definitely make those changes. I unfortunately don't have the bandwidth to take on integrating mavinit.scr capabilities into this right now, but i can move 'em to the end of the line.

I'm surprised that anyone is flying without a battery meter, that seems like madness to me!

njoubert:
OK! Added 'em to the end of the last row. I'd prefer them here: still visible in the default size of the console but doesn't push anything else around.

Also added a catch for the unsigned int battery voltage: display 0 in that case.

peterbarker:
Sorry, now I'm confused.  Just testing the patch and kind of noticed (and remembered...) that we do already have this information:

bat

Top-left-hand corner

njoubert:
Ah. I did some digging. You know what - I believe you have the "battery" module loaded. This isn't loaded by default on my side. cool. in that case, let's get rid of this PR :)

njoubert:
More digging: In actual fact, the battery module is loaded by default, but isn't correct for PX4-based autopilots. I'm flying PX4, not Ardupilot, hence the discrepancy. I will send another PR to fix up the battery module.

PR #432, opened 2017-07-26
BATTERY: Ensure PX4 compatibility

njoubert:
PX4 autopilots does not have the BATT_MONITOR parameter. These autopilots still report useful battery information! This PR ensures that the console will correctly report battery information when the parameter is not present but battery information is available.

### DIFF
--- a/MAVProxy/modules/mavproxy_battery.py
+++ b/MAVProxy/modules/mavproxy_battery.py
@@ -44,7 +44,7 @@ class BatteryModule(mp_module.MPModule):
                                                               self.vcell_to_battery_percent(self.per_cell)))
 
     def battery_report(self):
-        batt_mon = int(self.get_mav_param('BATT_MONITOR',0))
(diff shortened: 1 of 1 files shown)


njoubert:
@peterbarker This is a continuation of the previous battery PR!

peterbarker:
I don't think we can't change existing behaviour, which is that if you don't have battery monitoring configured you don't get those messages etc.

Is there something else in the mavlink stream we can look at to determine if "PX4 Pro" is sending through sensible battery monitoring?

njoubert:
This doesn't change the behavior on an autopilot that sends the BATT_MONITOR parameter, so if you're flying ardupilot without a battery monitor, you're fine. It's only if the parameter doesn't exist.

peterbarker:
On Thu, 27 Jul 2017, Niels Joubert wrote:

> 
> This doesn't change the behavior on an autopilot that sends the BATT_MONITOR
> parameter, so if you're flying ardupilot without a battery monitor, you're
> fine. It's only if the parameter doesn't exist.

Not true; we may not have seen all parameters yet :-)

PR #433, opened 2017-08-01
Update help of "setspeed" command

h4writer:
the command is actually called 'setspeed' instead of 'speed'. The help information is showing 'speed' which might be confusion. I assume this was called 'speed' before and refactored, but the help message wasn't updated?


tridge:
thanks!

PR #435, opened 2017-08-03
Add sync function to send timesync message

h4writer:
This is something I need for my own tools. It will send a "timesync" message and print the response to commandline. My tools use this to correlate the drone time with the local time. I have a fork with this change, but proposing to add it if others can use it.

### DIFF
--- a/MAVProxy/modules/mavproxy_sync.py
+++ b/MAVProxy/modules/mavproxy_sync.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+
+import time, os
+from pymavlink import mavutil
+
+from MAVProxy.modules.lib import mp_module
+
+class SyncModule(mp_module.MPModule):
+    def __init__(self, mpstate):
+        super(SyncModule, self).__init__(mpstate, "sync")
+        self.add_command('sync', self.cmd_sync, "sync")
+
+    def cmd_sync(self, args):
+        import time
+        if ( len(args) != 1):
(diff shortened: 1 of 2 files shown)


peterbarker:
This looks good.  But could we rename this to `timesync` rather than just `sync`, please?

As a future enhancement, it would be nice to be able to set the system time based on a received timesync message.  New PR if you ever feel like doing that, 'though ;-)

h4writer:
Updated.

Wrt setting the system time. What do you mean?
- Updating the drones time? Not sure that is possible...
- Updating the GCS time? Not sure that is wanted...

Currently what I do is run a sync every second and keep the following:
time_send
time_receive

As a result the delta between GCS time and drone time is:
dtime = (time_send + time_receive) / 2 - drone_time;

And when doing that a lot the dtime will average to the correct delta. As a result it is now possible to know exactly when a "GLOBAL_POSITION_INT" was created in GCS time.

I can add a function that would report back GCS time, if you give it a drone time ... (in a new PR)

peterbarker:
Sorry, I misunderstood the purpose of the module.

I've merged this in with the following changes:
 - added your description of the module's use at the top
 - moved all imports to top and split them to be one-per-line
 - remove the time temporary variable (shadowing a package name wasn't nice)
 - imported print_function from future
  - fixed non-function use of print
 - removed unused imports
 - whitespace fixes
 - made the commit message little more formal

Thanks!

PR #437, opened 2017-08-10
Error-exit if >2 serial devices detected

peterbarker:
Old behaviour is that we would listen on port 14550 as if no
serial devices were detected


khancyr:
LGTM

SamuelDudley:
LGTM

peterbarker:
Thanks @khancyr @amilcarlucas @SamuelDudley

PR #440, opened 2017-08-16
mavproxy_param.py: announce values for correct fetch_one'd parameters

peterbarker:
Without this any PARAM_VALUE that comes in is considered a fetch-one
parameter


amilcarlucas:
This one has already been merged. @peterbarker please close this

Issue #443, opened 2017-08-27
mavproxy_link : Does not filter by the target component in settings


crashmatt:
mavproxy link uses and forwards messages to all modules regardless of source.

[master_callback processes messages here](https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_link.py#L302)
Result is continuous conflicting voice announcements as messages from different systems are processed

Some modules require all messages and some modules require only target messages. Either modules takes care of filtering on a by use basis or there is a module based switch for messages being pre-filtered.

Issue #442, opened 2017-08-26
MAVProxy : Does not set target system/component in mavutil.mavfile


crashmatt:
MAVProxy does not set target system in mavutil so mavutil can't filter on the correct system.

[Relates to this issue in pymavlink](https://github.com/ArduPilot/pymavlink/issues/106)

PR #444, opened 2017-08-27
WIP : filter_target setting option to support links with multiple heartbeats

crashmatt:
**WIP : Do not pull**

Setting option to enforce filtering of messages by target system id
Enables configurations that have multiple system heartbeats on one link. Use as follows:
1. Configure main mavproxy without filtering to forward messages to outputs.  Run this with only the outputs module.  Can be headless.
2. Run an instance of mavproxy per target system which connects to the main instance.  Set --TARGET-SYSTEM and --filtertarget for each instance.

When each instance requests params for its target then those params only show in that instance.  They do not show in the other instances.

[Relates to this mavlink pr](https://github.com/ArduPilot/pymavlink/pull/111)
[This mavproxy_link issue](https://github.com/ArduPilot/MAVProxy/issues/443)
[This MAVProxy issue](https://github.com/ArduPilot/MAVProxy/issues/442)

The screenshot shows px4 posix_sitl_default running with modified script to communicate with a hardware radio instead of udp.
The terminals at right show the mavproxy windows. In order from top to bottom:
1. MAVProxy connected to the gcs end of the radio and outputting to the instances below.  It is showing random mode messages since it is confused about having multiple heartbeats.
2. MAVProxy instance connected to PX4.  Shows 670 parameters.
3. MAVProxy instance connected to the local end of the radio. Shows 8 parameters.
4. MAVProxy instance connected to the remote end of the radio. Shows 8 parameters

image

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -157,6 +157,7 @@ class MPState(object):
               MPSetting('source_component', int, 0, 'MAVLink Source component', range=(0,255), increment=1),
               MPSetting('target_system', int, 0, 'MAVLink target system', range=(0,255), increment=1),
               MPSetting('target_component', int, 0, 'MAVLink target component', range=(0,255), increment=1),
+              MPSetting('filter_target', bool, False, 'MAVLink target component'),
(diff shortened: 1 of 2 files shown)


amilcarlucas:
Can you rebase this?

PR #445, opened 2017-08-29
Module Map: fix possible encoding error

khancyr:
prevent encoding path in unicode when using mission load from map GUI
to reproduce the problem : have a computer in french with Ubuntu 16.04.3 LTS
 , launch SITL, use the map GUI to load a mission file from Téléchargements (Downloads in english). It will crash


peterbarker:
Merged, thanks!

PR #446, opened 2017-08-30
create mavproxy_surfer.py

auturgy:
This does not work, but it doesn't break anything either.  Assistance in completing it is welcomed.
This module is just a modification of mavproxy_speech.py.  It has been started in response to an ArduPilot user who is using APMRover to provide navigation guidance to a blind paddleboarder.  Currently haptic feedback is provided to the paddleboarder via pwm activated vibrating anklets, however voice directions are the next step. I think this is the easiest way to do that.  Intent is SERVO_OUTPUT_RAW is read for direction announcements every 5 seconds (0.2Hz) and distance to next waypoint from NAV_CONTROLLER_OUTPUT is announced every 10 seconds (0.1Hz).  If there is a better way to do it: please, go for it!
Thanks guys - your help is appreciated.
r/ James


auturgy:
Cancelled.  Peter pointed out https://github.com/ArduPilot/MAVProxy/issues/428 which will do this, but is far more versatile.

PR #449, opened 2017-09-20
mavproxy_cmdlong: add support for sending command_int

rmackay9:
This adds support for sending command_int messages.

It borrows heavily from the "long" command which sends command_long messages although this new one is slightly less flexible in that the user must provide all arguments instead of defaulting later arguments to zero.


SamuelDudley:
looks good

peterbarker:
After a little conversation with Randy I've pushed up this code with a few small modifications.

Thanks @rmackay9!

PR #452, opened 2017-09-30
Support loading and reloading modules with optional keyword arguments

SamuelDudley:
#### EDITED 07/12/2017

This PR adds the ability to supply optional keyword arguments to a module as they are loaded. If supported by the module it allows users to configure the module at the ``__init__`` stage.
The MAVCesium module really needs this ability in order to take a path to a user settings file. See https://github.com/SamuelDudley/MAVCesium/issues/21

This PR does not affect existing module functionality and handles errors gracefully when users attempt to supply keyword arguments to a module that does not support them.

**Example of usage:** ``module load cesium:{"config":"~/MAVProxy/mavcesium.ini","foo":10}``

If entered from the MAVProxy console the above string will load the cesium module and pass `{config:~/MAVProxy/mavcesium.ini, foo:10}` into the `__init__` method. 

The PR also supports passing keyword arguments to a module as you launch mavproxy. One can now pass multiple optional arguments to modules at launch.

**Example of usage:** `--load-module cesium:'{"config":"~/MAVProxy/mavcesium.ini","foo":10}'` 

The above launch argument would load the `cesium` module and pass `{config:~/MAVProxy/mavcesium.ini, foo:10}` into the `__init__` method.

**Supporting optional kwargs**
 **Note:** Existing modules not wanting to make use of this feature don't need to be modified

For the module to support the keyword arguments the following is needed (note the addition of **`**kwargs`**):

```
class ExampleKeywordModule(mp_module.MPModule):

    def __init__(self, mpstate, **kwargs):
        super(ExampleModule, self).__init__(mpstate, "example", "Example keyword module")
       
       # do something with the kwargs here
       for key, value in kwargs.iteritems():
            print "%s == %s" %(key,value)
        ....
        ....

def init(mpstate, **kwargs):
    '''initialise module'''
    return ExampleKeywordModule(mpstate, **kwargs)
````

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -14,6 +14,7 @@ import traceback
 import select
 import shlex
 import platform
+import json
 
 from MAVProxy.modules.lib import textconsole
 from MAVProxy.modules.lib import rline
@@ -266,7 +267,20 @@ def cmd_watch(args):
     mpstate.status.watch = args[0]
     print("Watching %s" % mpstate.status.watch)
 
-def load_module(modname, quiet=False):
+def generate_kwargs(args):
+    kwargs = {}
+    module_components = args.split(":{", 1)
+    module_name = module_components[0]
+    if (len(module_components) == 2 and module_components[1].endswith("}")):
+        # assume json
+        try:
+            module_args = "{"+module_components[1]
+            kwargs = json.loads(module_args)
+        except ValueError:
+            print('Invalid JSON argument: {0}'.format(module_args))
+    return (module_name, kwargs)
+
+def load_module(modname, quiet=False, **kwargs):
     '''load a module'''
     modpaths = ['MAVProxy.modules.mavproxy_%s' % modname, modname]
     for (m,pm) in mpstate.modules:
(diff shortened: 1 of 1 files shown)


fnoop:
+1 This would be a great addition to modules in mavproxy.  Currently you can't run more than one instance of mavproxy if modules consume any resources like a port, as they will clash.  I'm sure there are many other uses that this would be useful as well :)

amilcarlucas:
Yeap, the RTCM module would benefit from this.

SamuelDudley:
Quick update: In the gitter channel there were issues pointed out with using `:` as the split point in the strings. For example `config=C:/some_windows_path/conf.ini` would break at the `:`. As python does a pretty good job of parsing json strings I'll modify the PR and have another attempt at the syntax for comment.

SamuelDudley:
modified to support json format and rebased off master

peterbarker:
Isn't it the case that JSON lacks an escape character?

peterbarker:
OK, sorted things out.

I've incorporated this PR within https://github.com/ArduPilot/MAVProxy/pull/473 (it was the perfect thing, thanks!)

SamuelDudley:
@peterbarker The lack of escape char is a limitation, but it just means the JSON has to be last. Not so much an issue for this PR but for #394 order will matter. e.g. `link add tcp:127.0.0.1:14550:{"label":"foo"}` will work but `link add tcp:127.0.0.1:{"label":"foo"}:14550` will not.

 __TL:DR__ everything after the first `:{` is assumed to be JSON _if_ the final char is `}`. 

Glad your finding it useful! :+1:

SamuelDudley:
Merged with https://github.com/ArduPilot/MAVProxy/commit/b816e9fda96457aa7136b7a5e7e6ab2a1d35daab

peterbarker:
On Tue, 2 Jan 2018, Samuel Dudley wrote:

> Merged with b816e9f

Thanks Samuel!

PR #458, opened 2017-10-22
MAVCesium update v0.1

SamuelDudley:
*This PR updates the MAVCesium module to **v0.1***

This is the first release of MAVCesium bringing with it quite a few changes from the prior codebase:
* Sockets and static files are now served by a single Tornado server
* The module can now be reloaded in MAVProxy
* Cesium version update
* Application prefixes are now supported with`/mavcesium` being the default
* Experimental support for displaying sensor footprints on terrain
* UI update to bootstrap 4
* Added the ability to run MAVCesium stand alone from MAVProxy


peterbarker:
Merged, thanks!

PR #466, opened 2017-11-27
wxconsole: fix freeze_support

justineaster:
Fix bug that was introduced when multiprocessing was switched to billiards


peterbarker:
Merged, thanks!

PR #473, opened 2017-12-16
Support specification of XML filepath for parameters

peterbarker:
So instead of downloading from the interwebs based on the sort of vehicle we're connecting to, you can directly specify a file to use.

This will be used by an option to `sim_vehicle.py` so your parameters are always fresh.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1103,6 +1103,11 @@ if __name__ == '__main__':
     # call this early so that logdir is setup based on --aircraft
     (mpstate.status.logdir, logpath_telem, logpath_telem_raw) = log_paths()
 
+    for module in opts.load_module:
+        modlist = module.split(',')
+        for mod in modlist:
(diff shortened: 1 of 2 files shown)


MAVProxy/mavproxy.py
```diff
@@ -1103,6 +1103,11 @@ def quit_handler(signum = None, frame = None):
     # call this early so that logdir is setup based on --aircraft
     (mpstate.status.logdir, logpath_telem, logpath_telem_raw) = log_paths()
 
+    for module in opts.load_module:
+        modlist = module.split(',')
+        for mod in modlist:
+            process_stdin('module load %s' % (mod))
```
SamuelDudley: The left over `()` enclosing `mod` was an accidental left over from my PR dev... I can remove it there or since you are moving the logic it could be done here (or optionally not at all!)

SamuelDudley:
Apart from my one comment above (which is trivial) LGTM

peterbarker:
I've pushed this in.

PR #479, opened 2017-12-22
Add non interactive flag

estechnical:
I have added the possibility to run more easily in docker container. The addition of the `--non-interactive` flag will prevent the cli from running, much like the `--daemon` mode, without detaching.

Simple set of changes, seem to work fine for my really minimal use case - your thoughts appreciated here - I'm not experienced with python.

Tested with:
```
mavproxy.py --master=/dev/ttyACM0 --out=udpin:0.0.0.0:14550 --non-interactive --load-module=log,output
```

Thanks :)

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1140,7 +1140,7 @@ if __name__ == '__main__':
     # up on exit
     while (mpstate.status.exit != True):
         try:
-            if (opts.daemon == True or opts.non_interactive == True):
+            if opts.daemon or opts.non_interactive:
                 time.sleep(0.1)
             else:
                 input_loop()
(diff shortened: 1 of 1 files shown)


MAVProxy/mavproxy.py
```diff
@@ -1139,7 +1140,7 @@ def quit_handler(signum = None, frame = None):
     # up on exit
     while (mpstate.status.exit != True):
         try:
-            if opts.daemon:
+            if (opts.daemon == True or opts.non_interactive == True):
```
peterbarker: Just asserting rather than checking for equivalence is sufficient here.  Also, Python usually excludes the prens unless required.  So
```if opts.daemon or opts.non_interactive:```

estechnical:
Ok, I cleaned that up.

estechnical:
Regarding the use case of this flag, it's just to make it easier to run in docker. Without it when the container is run daemonised, the process would quit and docker would clean up. The point of docker is to make micro-services, not building something like a virtual machine. MAVproxy, for me, is literally just a proxy for mavlink data in at least one setup. I can easily demonstrate the difference with a few docker commands if you have docker available?

peterbarker:
@estechnical OK, I understand.  You're aware that there are a couple of other proxy solutions out there?  `cmavnode` and `mavlink-router`.  They're both c-based so are somewhat less resource-intensive :-)

That being said, that doesn't stop this going in.

peterbarker:
I've squash the commits together, fixed the commit message (which I should have caught earlier!) and pushed it in.

Thanks!

PR #480, opened 2018-01-03
Named datalinks

peterbarker:
This PR is based on Samuel's https://github.com/ArduPilot/MAVProxy/pull/394/commits

It:
 - adds a command to change the link attributes
 - corrects an issue in rline
 - adds the labels to tab completion

@SamuelDudley I have a concern about the way we're storing these link attributes.  Because we're throwing them directly into the connection object itself, there's the risk of blatting an existing attribute (e.g. `conn.mav`.  Thoughts?

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -723,7 +723,7 @@ def check_link_status():
         mpstate.status.heartbeat_error = True
     for master in mpstate.mav_master:
         if not master.linkerror and (tnow > master.last_message + 5 or master.portdead):
-            say("link %u down" % (master.linknum+1))
+            say("link %s down" % (mp_module.MPModule.link_label(master)))
             master.linkerror = True
 
 def send_heartbeat(master):
--- a/MAVProxy/modules/lib/mp_module.py
+++ b/MAVProxy/modules/lib/mp_module.py
@@ -150,3 +150,11 @@ class MPModule(object):
             prompt = self.settings.vehicle_name + ':' + prompt
         self.mpstate.rl.set_prompt(prompt)
             
+    @staticmethod
+    def link_label(link):
+        '''return a link label as a string'''
+        if hasattr(link, 'label'):
+            label = link.label
+        else:
+            label = str(link.linknum+1)
+        return label
--- a/MAVProxy/modules/lib/rline.py
+++ b/MAVProxy/modules/lib/rline.py
@@ -3,6 +3,7 @@ readline handling for mavproxy
 '''
 
 import sys, glob, os, platform
+import re
 
 rline_mpstate = None
 redisplay = None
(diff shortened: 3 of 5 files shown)


SamuelDudley:
@peterbarker originally I thought about avoiding attribute blatting, but as it stands it is possible to set other link attributes such as `target_system`. I will agree that it's not a very safe way to modify the link object so perhaps we could add a whitelist of attributes that are safe? For now that could be just `[label]`, or we can avoid setting existing attributes all together. 
Currently I don't have a use for setting other attributes (other than the label) so either option would be fine with me. What are your thoughts?

peterbarker:
On Wed, 3 Jan 2018, Samuel Dudley wrote:

> @peterbarker originally I thought about avoiding attribute blatting, but as
> it stands it is possible to set other link attributes such as
> target_system. I will agree that it's not a very safe way to modify the
> link object so perhaps we could add a whitelist of attributes that are
> safe? For now that could be just [label], or we can avoid setting existing
> attributes all together.
> Currently I don't have a use for setting other attributes (other than the
> label) so either option would be fine with me. What are your thoughts?

Maybe leave it as-is - it's not external to MAVProxy in any way so we can 
fix it later if we need to.  The only thing that could reasonably go wrong 
is pymavlink starts to use "label" for something - and we can watch for 
that :-)

Are you OK with the changes I've made?

SamuelDudley:
Yep! LGTM
All-round the changes are much clearer (read better). I'm still learning but I greatly appreciate your time and effort getting this to a pushable state! Thanks again :)

peterbarker:
@stephendade I've squished all of this stuff together ready to go into the master branch.  Since you've fought with links more than most, could check that this will help with CUAV's link soup we have for the OBC, please?

SamuelDudley:
Bump :) still would love to see this merged in if possible

peterbarker:
@stephendade  - bump?

stephendade:
This is indeed quite useful! One issue I've found is when launching mavproxy:
```
mavproxy.py --master=udpout:127.0.0.1:14550:{"label":"local"}
Failed to connect to udpout:127.0.0.1:14550:{label:local} : global name 'link_args' is not defined
```

peterbarker:
On Tue, 20 Feb 2018, Stephen wrote:

> This is indeed quite useful! One issue I've found is when launching mavproxy:
> 
> mavproxy.py --master=udpout:127.0.0.1:14550:{"label":"local"}
> Failed to connect to udpout:127.0.0.1:14550:{label:local} : global name 'link_args' is not defined

I've pushed up a fix for this.

The failure was actually in an error path, because the commandline is 
actually invalid under Linux!

The `"` in your commandline are being slurped by the shell, so you end up 
passing invalid JSON into mavproxy and therefor to the JSON parser.

A correct line might be:

mavproxy.py '--master=udpout:127.0.0.1:14550:{"label":"local"}'

Please check it works for you now :-)

SamuelDudley:
I have found the following works for me as a launch argument (note the single quotes around the curly braces):

``` bash
mavproxy.py --master=/dev/ttyUSB0:'{"label":"RFD900x"}'
```
when mavproxy is running you should omit them:

``` bash
link add /dev/ttyUSB0:{"label":"FTDI"}
```

stephendade:
Yep, that all looks good!

Issue #483, opened 2018-01-11
set moddebug 4 not working


magicrub:
set moddebug 1 or 2 or 3 work but 4+ give an error

MAVProxy 1.6.2
Windows 7

stephendade:
Fixed! Just need to merge it in

magicrub:
Thank you!

PR #484, opened 2018-01-16
Added better option parsing for --moddebug

stephendade:
Fix for #483. Ensure that we have a valid --moddebug value before continuing

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -989,6 +989,10 @@ if __name__ == '__main__':
           print("ERROR: mavproxy takes no position arguments; got (%s)" % str(args))
           sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


MAVProxy/mavproxy.py
```diff
@@ -989,6 +989,10 @@ def set_mav_version(mav10, mav20, autoProtocol, mavversionArg):
           print("ERROR: mavproxy takes no position arguments; got (%s)" % str(args))
           sys.exit(1)
 
+    if opts.moddebug not in [0, 1, 2, 3]:
```
peterbarker: This won't catch someone doing a `set moddebug 5`

stephendade: The `set moddebug X` is already covered by existing code. The issue was the exisitng code did not check the commandline parameters.

Issue #486, opened 2018-01-18
Crash while connecting to a simulator or APM via radio


whxru:
Firstly I built MAVProxy according to [Windows Development Environment](http://ardupilot.github.io/MAVProxy/html/development/mavdevenvwindows.html) and started a simulator:
```shell
dronekit-sitl copter-3.3 --home=31.8872318,118.8193952,5,353
```
Then I try to connect to it via MAVProxy:
```shell
py -2 mavproxy.py --master=tcp:127.0.0.1:5760 --moddebug=3
```
An exception was threw:
```
Connect tcp:127.0.0.1:5760 source_system=255
Log Directory:
Telemetry log: mav.tlog
MAV> fence breach
GPS lock at 0 meters
online system 1
STABILIZE> Mode STABILIZE
Exception in thread main_loop:
Traceback (most recent call last):
  File "C:\Python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "C:\Python27\lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "mavproxy.py", line 783, in main_loop
    master.wait_heartbeat()
  File "C:\Python27\lib\site-packages\pymavlink\mavutil.py", line 383, in wait_heartbeat
    return self.recv_match(type='HEARTBEAT', blocking=blocking)
  File "C:\Python27\lib\site-packages\pymavlink\mavutil.py", line 344, in recv_match
    m = self.recv_msg()
  File "C:\Python27\lib\site-packages\pymavlink\mavutil.py", line 318, in recv_msg
    msg = self.mav.parse_char(s)
  File "C:\Python27\lib\site-packages\pymavlink\dialects\v10\ardupilotmega.py", line 9257, in parse_char
    self.__callbacks(m)
  File "C:\Python27\lib\site-packages\pymavlink\dialects\v10\ardupilotmega.py", line 9234, in __callbacks
    self.callback(msg, *self.callback_args, **self.callback_kwargs)
  File "build\bdist.win32\egg\MAVProxy\modules\mavproxy_link.py", line 375, in master_callback
    mavutil.mavlink.MAV_TYPE_DODECAROTOR]:
AttributeError: 'module' object has no attribute 'MAV_TYPE_DODECAROTOR'
```

Even though I tried to connect to a real copter with an APM board on it via 3DR radio, it crashed as well.

---

* OS: Windows 10
* pip list:
    * backports.functools-lru-cache (1.4)
[...]

PR #487, opened 2018-01-18
Upversioned required pymavlink to 2.2.8

stephendade:
Based on comments in #486 - MAVProxy will crash if it uses any pymavlink older than 2.2.8, so makes sense (and makes the install much smoother) to make this the minimum required pymavlink.


SamuelDudley:
LGTM

PR #490, opened 2018-02-14
Enable mavproxy to watch multiple messages

patrickelectric:
Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -264,7 +264,7 @@ def cmd_watch(args):
     if len(args) == 0:
         mpstate.status.watch = None
         return
-    mpstate.status.watch = args[0]
+    mpstate.status.watch = args
     print("Watching %s" % mpstate.status.watch)
 
 def generate_kwargs(args):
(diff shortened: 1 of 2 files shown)


peterbarker:
Sorry, I haven't tried to run this.  But where are the changes which makes this a dict?

How do you "unwatch" something?  That's currently done with "watch" bare.

patrickelectric:
Is not a dict, is a list with the messages names.
To 'unwatch', just uses watch as a normal case.

OXINARF:
We should probably get out of the loop once the message is printed right? We don't need to print it multiple times.

patrickelectric:
> We should probably get out of the loop once the message is printed right? We don't need to print it multiple times.

that's what `status` command do.

OXINARF:
@patrickelectric What I meant is that it's looping through the entire watch list, even after it has been printed. You can have two watch types that will match the same message but you don't want to print it twice, right?

patrickelectric:
@OXINARF check now, now it'll break if the message is find.

patrickelectric:
Any update ?

patrickelectric:
72 days and counting..

MAVProxy/mavproxy.py
```diff
@@ -561,11 +561,12 @@ def process_mavlink(slave):
         return
     if msgs is None:
         return
-    if mpstate.settings.mavfwd and not mpstate.status.setup_mode:
-        for m in msgs:
-            if mpstate.status.watch is not None:
-                if fnmatch.fnmatch(m.get_type().upper(), mpstate.status.watch.upper()):
+    if mpstate.settings.mavfwd and not mpstate.status.setup_mode and mpstate.status.watch:
+        for msg_type in mpstate.status.watch:
+            for m in msgs:
```
OXINARF: I think this still isn't OK. The for loops should be switched. What we want is for all the messages we have received (`msgs`), check if it matches at most one watch regex that has been set - what it does now is for every regex that has been set check if there is at most one received message that matches.

patrickelectric: This way will avoid unnecessary loop interactions.
I did a python script if you want to try :)

patrickelectric:
@OXINARF any update ?

PR #495, opened 2018-03-08
map: add 'Set Origin' options

peterbarker:


rmackay9:
This change is important because setting the EKF's origin is required when using ardupilot without GPS.  So for example, wheel encoders or visual odometry is used, the EKF needs to have it's origin set, this message does this!

peterbarker:
@rmackay9 said this worked for him :-)

PR #500, opened 2018-04-12
map: new, better statkart map.

AndKe:


stephendade:
Looks good! Merging this in.

PR #502, opened 2018-04-26
Fix loading of modules with multiple arguments.

amilcarlucas:
This fixes b816e9fda96457aa7136b7a5e7e6ab2a1d35daab
the proper JSON format separator is ':' and not '=' therefore:

e.g.: module load cesium:{"arg_name":"arg_value","foo":10}

also support loading modules with keyword args at launch
e.g.: --load-module cesium:'{"arg_name":"arg_value","foo":10}'

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -269,15 +269,18 @@ def cmd_watch(args):
 
 def generate_kwargs(args):
     kwargs = {}
-    module_components = args.split(":{", 1)
+    module_components = args.split(":", 1)
     module_name = module_components[0]
-    if (len(module_components) == 2 and module_components[1].endswith("}")):
-        # assume json
-        try:
-            module_args = "{"+module_components[1]
-            kwargs = json.loads(module_args)
-        except ValueError as e:
-            print('Invalid JSON argument: {0} ({1})'.format(module_args,
+    if len(module_components) == 2:
+        if module_components[1].startswith("'") and module_components[1].endswith("'"):
+            module_components[1] = module_components[1][1:-1] # strip the ' quotes
+        if module_components[1].startswith("{") and module_components[1].endswith("}"):
+            # assume json
+            try:
+                module_args = module_components[1]
+                kwargs = json.loads(module_args)
+            except ValueError as e:
+                print('Invalid JSON argument: {0} ({1})'.format(module_args,
                                                            repr(e)))
     return (module_name, kwargs)
 
@@ -1112,7 +1115,7 @@ if __name__ == '__main__':
     (mpstate.status.logdir, logpath_telem, logpath_telem_raw) = log_paths()
 
     for module in opts.load_module:
-        modlist = module.split(',')
+        modlist = shlex.split(module, ',')
         for mod in modlist:
             process_stdin('module load %s' % (mod))
 


SamuelDudley:
I have not tested your changes but I can say that the commit message that you are referring to is not a correct reflection of how both the labelled links and keyword argument code was implemented. As you can see from the code it is looking for a JSON sub string, so yes a `:` is needed not `=`. The commit message is a throwback to an earlier version of the PR which was reviewed and dropped in favour for the JSON formatting.

SamuelDudley:
The `:` inside your strings should not be an issue with the existing code as everything inside the first `:{` to the last `}` will be treated as JSON. So perhaps there is some other formatting error that is occurring?

Reading your changes it seems the only functional difference is this additional if statement to check for leading and trailing `'`. 

https://github.com/ArduPilot/MAVProxy/pull/502/files#diff-2da3cf671c984b488ffffdd674c8e0cfR275

Using the existing code and entering `module load RTCM:{"addr":"tcp:192.168.3.1:5000","inst":"2"}` what happens? Is there an error? If so can you please post it here?

MAVProxy/mavproxy.py
```diff
@@ -1112,7 +1115,7 @@ def quit_handler(signum = None, frame = None):
     (mpstate.status.logdir, logpath_telem, logpath_telem_raw) = log_paths()
 
     for module in opts.load_module:
-        modlist = module.split(',')
+        modlist = shlex.split(module, ',')
```
amilcarlucas: The module name spiting logic separates the JSON into two modules one called `RTCM:{"addr":"bla"` and one called `"inst:"2"}` , hence the `shlex` call

MAVProxy/mavproxy.py
```diff
@@ -269,15 +269,18 @@ def cmd_watch(args):
 
 def generate_kwargs(args):
     kwargs = {}
-    module_components = args.split(":{", 1)
+    module_components = args.split(":", 1)
```
amilcarlucas: When there is a `:'{` the old code breaks

amilcarlucas:
@SamuelDudley can you please review this ?

amilcarlucas:
@SamuelDudley friendly ping

amilcarlucas:
```
%PYTHON2%\python mavproxy.py --mav10 --streamrate -1 --out 127.0.0.1:14550 --out 127.0.0.1:14551 --out 127.0.0.1:14552 --master COM8 --load-module RTCM:'{\"addr\":\"tcp:192.168.3.1:5000\",\"inst\":\"2\"}'
Connect COM8 source_system=255
  File "mavproxy.py", line 311, in load_module
    m = import_package(modpath)
  File "mavproxy.py", line 455, in import_package
    mod = __import__(name)
ImportError: No module named mavproxy_RTCM:'{"addr":"tcp:192.168.3.1:5000"

  File "mavproxy.py", line 311, in load_module
    m = import_package(modpath)
  File "mavproxy.py", line 455, in import_package
    mod = __import__(name)
ImportError: No module named RTCM:'{"addr":"tcp:192.168.3.1:5000"

Failed to load module: No module named RTCM:'{"addr":"tcp:192.168.3.1:5000".
  File "mavproxy.py", line 311, in load_module
    m = import_package(modpath)
  File "mavproxy.py", line 455, in import_package
    mod = __import__(name)
ImportError: No module named mavproxy_"inst"

  File "mavproxy.py", line 311, in load_module
    m = import_package(modpath)
  File "mavproxy.py", line 455, in import_package
    mod = __import__(name)
ImportError: No module named "inst"
```

There are the errors that I do not get when using this patch

MAVProxy/mavproxy.py
SamuelDudley: IMO rather than use `shlex.split()` here I think we should do a json argument detection. That would save the json formatting and keep the existing downstream code as it it

MAVProxy/mavproxy.py
SamuelDudley: I have not found this to be the case. I believe the problem is with `modlist = module.split(',')` cutting the json string into what it thinks are module names

SamuelDudley:
I can replicate the error when loading modules with multiple arguments at launch. Loading them at run time is not an issue. @amilcarlucas I have added my thoughts as comments to your PR. The other reason I would like to keep the downstream code the same is that the named datalinks use a duplicated bit of code which I would like to refactor into a single function.

MAVProxy/mavproxy.py
```diff
@@ -269,15 +269,18 @@ def cmd_watch(args):
...
     module_name = module_components[0]
-    if (len(module_components) == 2 and module_components[1].endswith("}")):
```
SamuelDudley: Ideally I would like to see this section of code remain as is and handle the comma separated module list prior to `modlist = module.split(',')`

MAVProxy/mavproxy.py
amilcarlucas: Yes, `module.split(',')` cutting the json string into what it thinks are module names is the problem.

MAVProxy/mavproxy.py
amilcarlucas: What exactly do you mean with " json argument detection"?

SamuelDudley: Replace module.split(',') with a function that is looking for json strings within the module string and avoids splitting them. A regular expression could be used to identify json substrings within the module string. Possibly something like `([:{].*?[}])` (note: untested)
Then split the string into a list knowing to avoid any split locations that are in a json substring. That's one way of doing it anyway...

SamuelDudley: A quick play gives `,+(?=[^{}]*(?:\{|$))` as a working solution

so
```
regex = r",+(?=[^{}]*(?:\{|$))"
modlist = re.split(regex , module)
```
should work

see: https://regex101.com/r/uN8oK4/1

SamuelDudley:
@amilcarlucas To summarize my notes above:

- As you pointed out, the issue comes from `modlist = module.split(',')` splitting json arguments when trying to parse multiple comma separated module names supplied as a launch argument.
- Solving the issue at the `module.split(',')` function leaves the remaining code as is and makes it suitable for refactoring along with the named datalink json code.
- Using a regular expression to split the `module` string looks like a feasible option that will solve the issue.

peterbarker:
On Wed, 3 Oct 2018, Samuel Dudley wrote:

>  *  Using a regular expression to split the module string looks like a
>     feasible option that will solve the issue.

Generally regular expressions can't be used to parse arbitrary expressions 
- usually you'll need an extention to do that.

Parsing it out as JSON seems best - 'though I haven't followed the 
arguments against that, sorry.

SamuelDudley:
@peterbarker I have an example of a regular expression ignoring the json and only splitting at the correct locations here: https://regex101.com/r/uN8oK4/5

peterbarker:
On Thu, 4 Oct 2018, Samuel Dudley wrote:

> @peterbarker I have an example of a regular expression ignoring the json
> and only splitting at the correct locations here:
> https://regex101.com/r/uN8oK4/5

RTCM:{"addr":"tcp:192.168.3.1:5000","inst":"2"},adsb,example:{"foo":10},console,bar:{"x,{":1},

SamuelDudley:
Fair enough. Passing a dict is a pretty valid use case.

amilcarlucas:
How to proceed with this? I think it is pretty clear that the PR fixes a bug

arshPratap:
@amilcarlucas  Sorry for the delay sir. I think the solution as stated above (the one that required splitting the module variable through a regex pattern) although seemed to work at first, was probably not complete as shown by Mr.Peter Barker.The problem was that the regex pattern was splitting the string at some particular commas but seems to malfunction in cases like this : 

> RTCM:{"addr":"tcp:192.168.3.1:5000","inst":"2"},adsb,example:{"foo":10},console,bar:{"x,{":1},

One way to solve the issue is to modify the regex pattern so that it takes into account the "," and "{" in the quotation marks and prevents the splitting to take place in that case
A second way is to define a splitter function as follows:-

- Definition of function 
```
def_ splitter(s):
    l=re.findall(r'"(.*?)"',txt)
    l=list(filter(re.compile(".*,").match,l))
    unwanted_commas=[s.find(i)+i.find(",") for i in l]
    total_commas=[m.start() for m in re.finditer(r",+(?=[^{}]*(?:\{|$))",s)]
    unwanted_commas=set(unwanted_commas)
    total_commas=set(total_commas)
    if len(total_commas.intersection(unwanted_commas))>0:
        total_commas=total_commas-unwanted_commas
    wanted_commas=list(total_commas)
    wanted_commas.append(-1)
    wanted_commas.sort()
    l.clear()
    l=[s[i+1:j] for i,j in zip(wanted_commas,wanted_commas[1:]+[None])]
    return l
```
- Usage of the function
`modlist=splitter(module)`

Please do take a look! Thanks

PR #509, opened 2018-05-14
system_time: add module for sending the system_time message

peterbarker:


MAVProxy/modules/mavproxy_system_time.py
```diff
@@ -0,0 +1,93 @@
...
+            print(self.usage())
+
+    def status(self):
+        '''returns information about module'''
+        return "Perfectly happy.  All is good with the world"
+
+    def uptime(self):
+        ret = 0
+        try:
+            with open("/proc/uptime") as f:
```
OXINARF: This won't work on Windows, right? Isn't there a proper Python API?

stephendade: Correct - won't work in Windows. There should be something in the pywin32 package to help with this.

peterbarker:
On Tue, 22 May 2018, Francisco Ferreira wrote:

> +            with open("/proc/uptime") as f:
> 
> This won't work on Windows, right? Isn't there a proper Python API?

Maybe?  I couldn't find one in the 1 minute of searching I put into it.

OXINARF:
There is https://pythonhosted.org/uptime/ (which isn't updated for a long time, but maybe it just works).
psutil provides boot time (http://psutil.readthedocs.io/en/latest/#psutil.boot_time). Both put a new dependency into MAVProxy.

tridge:
the prints every few seconds in this module drive me nuts.
Also, I think "time since boot" could be time since mavproxy started

PR #516, opened 2018-06-07
Mavflightview: add treat POS message as GPS in case of noGPS mode

khancyr:


peterbarker:
Won't this plot both `GPS` *and* `POS` in the case they're both present?

Perhaps we need an `options.pos` instead?

In fact, I think we really should replace `GPS` with `POS` in the default case, paired against `GLOBAL_POSITION_INT` - they're more analogous.  We'd then need an `options.gps`...

tridge:
it could check if we have GPS msgs by doing:

 if not 'GPS' in mav.messages:
    types.append('POS')

PR #518, opened 2018-06-14
live_graph: cope with removed method in matplotlib

peterbarker:


peterbarker:
I've pushed something like this in after testing it doesn't break my system :-)

PR #519, opened 2018-06-15
Added mavreplay.py

Perseus101:
The mavreplay script replays telemetry from a tlog file and outputs it to a mavlink socket that a GCS can connect to.


peterbarker:
On Fri, 15 Jun 2018, Colin Moore wrote:

> The mavreplay script replays telemetry from a tlog file and outputs it to a mavlink socket that a GCS can connect to.

Kind of like mavplayback.py, then?

https://github.com/ardupilot/pymavlink/tree/master/tools

PR #521, opened 2018-07-03
Use sub icons and ArduSub with submarines

patrickelectric:
Add sub icons and update submarine type to use sub and ardusub


peterbarker:
Telling front from back is tough with these icons.  Still, the guts work, and they can be improved later.

PR #524, opened 2018-07-30
"arm check foo" changes

peterbarker:
These changes disambiguate some "arm check" uses.

Now:
`arm check all` - sets ARMING_CHECK to 1 (rather than bits individually)
`arm uncheck all` - sets ARMING_CHECK to 0
`arm uncheck LOGGING` - IF ARMING_CHECK was "1", sets it to zero then sets all bits we know about (https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_arm.py#L9), *then* sets LOGGING bit to zero.

In the third case, I'm wondering if we should set the first 28 bits or so to 1 instead (rather than just the bits we know about), so newer arming checks aren't disabled by `arm uncheck FOO`.

### DIFF
--- a/MAVProxy/modules/mavproxy_arm.py
+++ b/MAVProxy/modules/mavproxy_arm.py
@@ -6,21 +6,41 @@ import time, os
 from MAVProxy.modules.lib import mp_module
 from pymavlink import mavutil
 
+# note that the number of bits here is contrained by the float
+# transport mechanism.  25 bits is the limit.  Given the transport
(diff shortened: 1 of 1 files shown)


OXINARF:
> In the third case, I'm wondering if we should set the first 28 bits or so to 1 instead (rather than just the bits we know about), so newer arming checks aren't disabled by arm uncheck FOO.

I think that makes more sense. As far as I know, it's not an issue on ArduPilot side.

amilcarlucas:
@peterbarker ping

PR #527, opened 2018-08-09
makes mavproxy more "friendly" for python3

patrickelectric:


peterbarker:
On Thu, 9 Aug 2018, Patrick José Pereira wrote:

.... but somewhat less friendly for Python2:

ImportError: cannot import name reload

patrickelectric:
Isn't moving from Python 2 to 3 so much fun ?
@peterbarker It should work now, ty.

MAVProxy/mavproxy.py
```diff
@@ -9,13 +9,31 @@
...
+import serial, select
 import traceback
 import select
 import shlex
 import platform
 import json
 
+try:
+    # Python 2.7
+    reload
```
peterbarker: python-future.org seems to indicate `imp` under Python 2 contains reload... this would simplify this code a chunk by not doing the NameError check.  I'm uncomfortable with executing reload to test if it's a valid symbol :-)

http://python-future.org/compatible_idioms.html

patrickelectric:
@peterbarker ty, as you can see, I don't have python super powers.
I can add future as a dependency to create a better commit for Queue import, but mavproxy needs pymavlink that has future as a dependency, I don't know if this is a pythonic thing to do. Or we can just add future as a dependency also for mavproxy.

peterbarker:
@patrickelectric tested on both Python2 and Python3?

patrickelectric:
In a native python3 machine running Arch and in python2 machine running Mac OS.

peterbarker:
OK, so we appear to get further with starting MAVProxy under Python3 - now I'm getting something about "init" function not found on terrain....

peterbarker:
Thanks!

PR #529, opened 2018-08-12
mavproxy: don't print "no link" continuously

OXINARF:
Not sure what you think of this, but I find it rather annoying to continuously have "no link" printed in MAVProxy console.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -718,7 +718,7 @@ def set_stream_rates():
 def check_link_status():
     '''check status of master links'''
     tnow = time.time()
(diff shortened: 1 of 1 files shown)


peterbarker:
discussed atthe devcall; using text-to-speech out flying this is actually a useful thing.

amilcarlucas:
Can this one be closed?

OXINARF:
@amilcarlucas Why would it be closed? It's going to be converted to be a setting, as discussed on the dev call.

amilcarlucas:
sorry, I was not at that dev-call, just saw the comment from Peter

amilcarlucas:
ping

PR #530, opened 2018-08-14
mavproxy.py: catch shlex exception

peterbarker:
Without this patch, typing:

foo = "

at the mavproxy command prompt is enough to kill the main stdin loop
dead.


peterbarker:
We should decide whether we should catch exceptions further up here somehow - maybe move `process_stdin` to `process_stdin_wrapped` and make `process_stdin` wrap it in a try-except block.

PR #536, opened 2018-08-24
Param: auto fetch param list when param count doens't match param filled

khancyr:
as parameter aren't stored with their index, we cannot just fetch the ones missing. 
So I propose a simple fetch all to retreive missing parameter.


peterbarker:
Won't this only trigger if the count is different at the end of a fetch-all?

This won't catch the case where someone else sets an enable parameter, right?

Pretty sure it's the case that we could have GCSs on the same link walking all over each other if we go down this sort of path - any time a GCS does a fetch-all we effectively abort any streaming we're doing for a different GCS.  With timeouts/retries in GCS we might end up starving everyone indefinitely.

khancyr:
Hum you are right ... I will rework that to use and ordered param list ... it will be worth it

khancyr:
close for now

Issue #535, opened 2018-08-23
TLOG file corrupted by line ending translations on WIndows


newts:
@peterbarker  this is the issue we gittered about last week

I traced this back to file opening not in binary mode in mavproxy.py.  There could be other binary files with similar problem?

I have tested the fix and will file the pull request shortly.

amilcarlucas:
Looks like this one can be closed.

amilcarlucas:
Works in the latest release. Closing it now. Please reopen if the problem persists in the latest release.

PR #537, opened 2018-08-24
fix corrupted tlog files from line ending mangling on Windows

newts:
@peterbarker 

This fixes the insanity with the tlog files for windows.  Linux obviously didn't exhibit the problem.

Dunno if there are other cases.  The log module looks like it does it right (I was initially confused with the lods module vs tlog writing in the mavproxy.py).

#535


peterbarker:
Merged, thanks!

This is a ridiculously bad one to live so long.

The only other place that sprung to mind was the dataflash-logger module - but it seems to get it right.

PR #541, opened 2018-08-28
Added "python-opengl" pre-requisite for magical module.

AndKe:


peterbarker:
Thanks for this, but we've killed off the gh-pages (thankfully!)

AndKe:
yes. that was >2 years ago :)

Issue #543, opened 2018-09-09
Unable to wait for the incomming connections on --master due to unexpected keyword argument 'timeout'


brodnev:
Both --master=tcpin and --master=udpin returns "timeout" after 2 sec
pi64 ~ # mavproxy.py --master=tcpin:0.0.0.0:14550 --baudrate 57600 --out=udp:192.168.10.59:14555
Connect tcpin:0.0.0.0:14550 source_system=255
Log Directory:
Telemetry log: mav.tlog
Waiting for heartbeat from tcpin:0.0.0.0:14550
 MAV> Exception in thread main_loop:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/lib/python-exec/python2.7/mavproxy.py", line 798, in main_loop
    master.wait_heartbeat(timeout=0.1)
TypeError: wait_heartbeat() got an unexpected keyword argument 'timeout'

pi64 ~ # mavproxy.py --master=udpin:0.0.0.0:14550 --baudrate 57600 --out=udp:192.168.10.59:14555
Connect udpin:0.0.0.0:14550 source_system=255
Log Directory:
Telemetry log: mav.tlog
Waiting for heartbeat from 0.0.0.0:14550
 MAV> Exception in thread main_loop:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/lib/python-exec/python2.7/mavproxy.py", line 798, in main_loop
    master.wait_heartbeat(timeout=0.1)
TypeError: wait_heartbeat() got an unexpected keyword argument 'timeout'

brodnev:
P.S. sources has been downloaded today, 09.09.2018

karthikdesai:
https://github.com/ArduPilot/MAVProxy/pull/544 try it with this PR @brodnev

umer936:
Had this issue as well since I needed to be back a pymavlink version. Did the PR manually and worked fine. 

More details on my situation (ie why I needed pymavlink==2.2.7) here: https://github.com/ArduPilot/pymavlink/issues/222

[...]

PR #544, opened 2018-09-09
MAVProxy: add exception to catch using older pymavlink

peterbarker:


karthikdesai:
This should fix #543 ?

khancyr:
````
MAV> Exception in thread main_loop:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/bin/mavproxy.py", line 805, in main_loop
    master.wait_heartbeat(timeout=0.1)
TypeError: wait_heartbeat() got an unexpected keyword argument 'timeout'
````
The patch is good but we should reduce the exception scope to  TypeError

peterbarker:
I'm going to kill this one.  tridge has now done a release so master now specifies more recent versions of pymavlink.

PR #547, opened 2018-09-13
removed APM reference in favour of ArduPilot

auturgy:


rmackay9:
Seems like a good safe change to me.

PR #561, opened 2018-10-16
live_graph_ui: fix drawing

khancyr:
Fix graphing in Ubuntu 18.04 both with python2 and python3, from matplotlib 2 to matplotlib 3

### DIFF
--- a/MAVProxy/modules/lib/live_graph_ui.py
+++ b/MAVProxy/modules/lib/live_graph_ui.py
@@ -137,6 +137,7 @@ class GraphFrame(wx.Frame):
             self.plot_data[i].set_ydata(ydata)
 
         self.canvas.draw()
+        self.canvas.Refresh()
 
     def on_pause_button(self, event):
         self.paused = not self.paused


tridge:
what is not working that this fixes? I've tested py2 and py3

khancyr:
@tridge here is a video with pip release showing the issue, and the result with the patch : https://www.youtube.com/watch?v=NjawwXP-bRY&feature=youtu.be

I just remember that I am using wayland and not X11, it could also be the cause of the issue.

tridge:
that video is completely black for me

> I just remember that I am using wayland and not X11, it could also be the cause of the issue.

ahh, could be. The reason I don't just apply this is I'd want to analyse CPU cost. The live graph code is very sensitive to small changes making a huge difference in CPU usage

khancyr:
Working video ! ( f***************** wayland...) https://www.youtube.com/watch?v=DsDhM52XJQM&feature=youtu.be

tridge:
ok. I looked at CPU usage and was disappointed to find high cpu usage with or without your patch in master at the moment. Darn!
Looks like there is some work to do there ...

khancyr:
Close as merged in https://github.com/ArduPilot/MAVProxy/commit/f7f37efab6d097ea1f4b8d08223edbba907dbfb5

PR #575, opened 2018-12-04
mavproxy_messagerate: simply display an average message rate

peterbarker:


OXINARF:
I wonder if this shouldn't be part of the _status_ command instead.

peterbarker:
@OXINARF One doesn't play with basics like that without permission :-)  But yeah, that would kind of make sense.

Also, can't merge this one right now - the 80/20 filter thing isn't appropriate here.  I'll move it to a record-history system today sometime.

tridge:
I'd prefer separate status command for msg rates

PR #584, opened 2019-01-22
live_graph: fix live graph on macos

bugobliterator:
Tested working on macos Mojave


MAVProxy/modules/lib/live_graph_ui.py
```diff
@@ -66,7 +66,7 @@ def init_plot(self):
             self.axes.set_facecolor('white')
         except AttributeError as e:
             # this was removed in matplotlib 2.2.0:
-            self.axes.set_axis_bgcolor('white')
```
tridge: are you sure this is right? We only get to this line if calling set_facecolor() caused an AttributeError exception

Issue #600, opened 2019-03-03
MAVProxy crashing after connecting Mission Planner


jo-m:
I also saw this very similar issue: https://github.com/ArduPilot/MAVProxy/issues/599, but the error seems to occur in a different place. So if those two issues point to the same underlying problem in MAVProxy, this one can be closed.

Also, not sure if this bug belongs here or in pymavlink.

# Poblem/How to reproduce
1. Run SITL
2. Run MAVProxy `./venv/bin/mavproxy.py --master tcp:localhost:5760 --out=127.0.0.1:14551`
3. Connect Mission Planner on UDP port 14551

Will result in
```
Exception in thread main_loop:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "./venv/bin/mavproxy.py", line 883, in main_loop
    process_mavlink(m)
  File "./venv/bin/mavproxy.py", line 633, in process_mavlink
    mpstate.master().write(m.get_msgbuf())
  File "xxx/venv/lib/python3.6/site-packages/pymavlink/dialects/v20/ardupilotmega.py", line 100, in get_msgbuf
    return bytearray(self._msgbuf)
TypeError: string argument without an encoding
```

I added the following debug lines in `ardupilotmega.py` on line 98:
```
import sys
sys.stderr.write(">>>>>>> %s %s\n" % (type(self._msgbuf), repr(self._msgbuf)))
```

which results in the following output right before the crash above:
```
>>>>>>> <class 'array.array'> array('B', [254, 22, 135, 1, 1, 193, 150, 188, 101, 60, 144, 154, 237, 59, 217, 0, 141, 59, 165, 164, 0, 59, 248, 30, 227, 59, 127, 3, 183, 249])
>>>>>>> <class 'array.array'> array('B', [254, 32, 136, 1, 1, 241, 13, 80, 29, 99, 0, 0, 0, 0, 164, 43, 188, 59, 163, 111, 193, 59, 139, 244, 29, 61, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 225, 114])
[...]

PR #602, opened 2019-03-04
don't try to forward BAD_DATA messages

jo-m:
Created because of this issue: https://github.com/ArduPilot/MAVProxy/issues/600

But forwarding BAD_DATA is nothing useful to do regardless.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -630,6 +630,8 @@ def process_mavlink(slave):
         return
     if mpstate.settings.mavfwd and not mpstate.status.setup_mode:
         for m in msgs:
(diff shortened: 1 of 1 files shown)


peterbarker:
A mavproxy connected to an output of this mavproxy will show this bad data in its console.

Sometimes it is just line-noise, sometimes it is actually useful data which just isn't in mavlink packets.

Fixing the crash is the correct fix, sorry.

jo-m:
Thanks. With the fix in pymavlink [merged now](https://github.com/ArduPilot/pymavlink/pull/285), the pymavlink version required my MAVProxy will have to be bumped after the next release.

PR #603, opened 2019-03-09
Python 3 Fixes

stephendade:
A bunch of Python 3 fixes. Also changing the Windows build over to Python 3.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -538,7 +538,7 @@ def process_stdin(line):
 
     if cmd == 'help':
         k = command_map.keys()
-        k.sort()
+        k = sorted(k)
         for cmd in k:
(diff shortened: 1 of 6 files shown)


MAVProxy/mavproxy.py
```diff
@@ -538,7 +538,10 @@ def process_stdin(line):
 
     if cmd == 'help':
         k = command_map.keys()
-        k.sort()
+        if sys.version_info.major < 3:
+            k.sort()
+        else:
+            sorted(k)
```
tridge: I thought sorted() returns a new list? Can be use:
 k = sorted(k)
for both versions?

stephendade: Fixed - using ``k = sorted(k)``

PR #611, opened 2019-04-20
mavproxy_console: add periodic warning of autopilot software failure

peterbarker:


MAVProxy/modules/mavproxy_console.py
```diff
@@ -349,6 +350,21 @@ def mavlink_packet(self, msg):
...
+                for field_num in range(1, 5):
+                    field = "errors_count%u" % field_num
+                    x = getattr(msg, field, None)
+                    if x is None:
+                        self.say("Failed to get field %s" % field)
+                        self.last_sys_status_errors_announce = now
+                        break
+                    if x != 0:
+                        self.last_sys_status_errors_announce = now
+                        self.say("System failure detected.  Land now.")
```
tridge: need to make the msg shorter. It takes a long time to announce these things, and meanwhile the pilot doesn't have airspeed etc announcements

MAVProxy/modules/mavproxy_console.py
```diff
@@ -349,6 +350,21 @@ def mavlink_packet(self, msg):
...
             self.last_sys_status_health = msg.onboard_control_sensors_health
 
+            # check for any error bits being set:
+            now = time.time()
+            if now - self.last_sys_status_errors_announce > self.mpstate.settings.sys_status_error_warn_interval:
+                for field_num in range(1, 5):
+                    field = "errors_count%u" % field_num
+                    x = getattr(msg, field, None)
+                    if x is None:
+                        self.say("Failed to get field %s" % field)
```
tridge: i don't think we should do a say() here

MAVProxy/modules/mavproxy_console.py
peterbarker: Shortened to "Critical failure"

MAVProxy/modules/mavproxy_console.py
peterbarker: Changed to be a writeln instead.

PR #612, opened 2019-04-20
Misseditor Distance and Gradient

Akshath-Singhal:
Added Distance and Gradient fields for better mission planning using MAVProxy GCS. Distance is displayed in meters and Gradient in degrees.
misseditor

Future Prospects:
1. Handling altitude based on reference frame (Relative or Absolute)
2. Better handling of different waypoints instead of checking LAT=0 & LON = 0

I have tried to include most of the changes suggested by @tridge.


Akshath-Singhal:
@tridge  Changing LAT, LON, and ALT based on user input changes in Dist and Grad would indeed be an interesting feature, however, I am skeptical due to the increased probability of crashes due to changing these values carelessly.

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
```diff
@@ -673,6 +689,40 @@ def fix_jumps(self, row_selected, delta):
...
+	    command_prev = self.grid_mission.GetCellValue(row - 1, ME_COMMAND_COL)
+            lat = float(self.grid_mission.GetCellValue(row, ME_LAT_COL))
+            lon = float(self.grid_mission.GetCellValue(row, ME_LON_COL))
+            if (lat == 0) and (lon == 0):
+                dist = 0.0
+                grad = 0.0
+	    else :
+              if "NAV" in command:
+                row_prev = row - 1
+		while "NAV" not in command_prev:
```
tridge: also check for row_prev going to zero

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
```diff
@@ -673,6 +689,40 @@ def fix_jumps(self, row_selected, delta):
...
+                    command_prev = self.grid_mission.GetCellValue(row_prev - 1, ME_COMMAND_COL)
+                    row_prev = row_prev - 1
+		prev_lat = float(self.grid_mission.GetCellValue(row_prev, ME_LAT_COL))
+		prev_lon = float(self.grid_mission.GetCellValue(row_prev, ME_LON_COL))
+		prev_alt = float(self.grid_mission.GetCellValue(row_prev, ME_ALT_COL))
+                while (prev_lat == 0) and (prev_lon == 0):
```
tridge: check for row_prev zero here too (don't want to end up negative)

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
```diff
@@ -673,6 +689,40 @@ def fix_jumps(self, row_selected, delta):
...
+                    prev_lat = float(self.grid_mission.GetCellValue(row_prev - 1, ME_LAT_COL))
+                    prev_lon = float(self.grid_mission.GetCellValue(row_prev - 1, ME_LON_COL))
+                    row_prev = row_prev - 1
+		dist = mp_util.gps_distance(lat, lon, prev_lat, prev_lon)
+		grad = math.atan2(float(self.grid_mission.GetCellValue(row, ME_ALT_COL)) - prev_alt, dist) *180 / 3.14159
```
tridge: use math.pi not 3.14159

tridge:
one more thing, I think we should have the function to get the gradient and distance be part of the mavproxy_wp module. I think it will be used by other pieces of code in the future.
Modules can call functions in other modules as long as they are public (wp is public)

tridge:
it would also be good to implement the handling of frames. That is a bit complex, but quite possible and would be very useful
I'd suggest having a function to a wp alt to absolute first, then do everything in absolute. For home height use the height of home location in terrain database (the mission editor already knows home location and displays it). For AGL wps use terrain database.

Akshath-Singhal:
Thanks. The idea of adding distance and gradient functions to wp module seems like a nice idea.
For reference frame issues, I was planning to save the current home altitude as variable and use it on the go instead of shifting all altitudes to absolute frame.

The changes might take some time due to my university examinations. Sorry for the delay being caused.

tridge:
no worries at all, thanks for your nice work on this!

stephendade:
I had a quick look at this - you're using tabs instead of spaces for the indents. You'll need to change them to spaces for consistency.

Akshath-Singhal:
> I had a quick look at this - you're using tabs instead of spaces for the indents. You'll need to change them to spaces for consistency.
@stephendade 
Corrected that. Thanks for informing.

tridge:
@Akshath-Singhal any update on this?

Akshath-Singhal:
@tridge I have added the frame handling part for gradient calculation. I will shift it to wp module soon.

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
Akshath-Singhal: done

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
Akshath-Singhal: done

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
Akshath-Singhal: done

tridge:
hmm, minor thing, col width too small by default for me:
image

Akshath-Singhal:
@tridge Squashed to 1 commit. Increased the default size for columns.

peterbarker:
Small reword on the commit message would be good.  We don't fuss *too* much in mavproxy, but something like "misseditor: Add Dist and Grad columns" would make it a bit clearer what's being changed in the PR.

MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
```diff
@@ -585,11 +601,11 @@ def on_mission_grid_cell_left_click(self, event):  # wxGlade: MissionEditorFrame
 
     def on_mission_grid_cell_changed(self, event):  # wxGlade: MissionEditorFrame.<event_handler>
         self.set_modified_state(True)
+        #self.set_grad_dist()
```
peterbarker: Stray line?

Akshath-Singhal: Corrected

tridge:
looks good, thanks!

PR #613, opened 2019-04-21
Joystick: support Logicool Gamepad F310

tajisoft:
It won't detect my gamepad(Logicool Gamepad F310) bought in Japan. This PR is may useful for Japanese users.


peterbarker:
Merged, thanks!

PR #614, opened 2019-04-23
setup: Add extras_require configuration

patrickelectric:
extras_require allow users to install mavproxy with some extra functionalities:
Like: `pip install mavproxy[server] --user` to install flask and everything necessary for server functionality 

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,10 @@ on how to use MAVProxy.''',
                 'MAVProxy.modules.lib.MacOS',
                 'MAVProxy.modules.lib.optparse_gui'],
       install_requires=requirements,
+      extras_require={
+        # restserver module
+        'server': ['flask'],
+      },
       scripts=['MAVProxy/mavproxy.py',
                'MAVProxy/tools/mavflightview.py',
                'MAVProxy/tools/MAVExplorer.py',


patrickelectric:
I'm aiming to add this in the documentation also to update it with the new restserver.

patrickelectric:
@peterbarker ^

setup.py
```diff
@@ -80,6 +80,10 @@ def package_files(directory):
                 'MAVProxy.modules.lib.MacOS',
                 'MAVProxy.modules.lib.optparse_gui'],
       install_requires=requirements,
+      extras_require={
+        # restserver module
+        'server': ['flask'],
```
peterbarker: Gah.  This was supposed to be `restserver` from a technical perspective...

patrickelectric: I think that extras_require should be not module specific but in a more generic context, in this case, with web and server features, that's why I'm using server as key here.

patrickelectric: @peterbarker

patrickelectric: Ping @peterbarker

patrickelectric: Ping @peterbarker

peterbarker:
Merged, thanks!

PR #616, opened 2019-04-23
Util: Try downloading the files through proxy before reporting an error

amilcarlucas:


MAVProxy/modules/lib/mp_util.py
```diff
@@ -250,12 +250,21 @@ def download_url(url):
...
         from urllib.request import urlopen as url_open
         from urllib.error import URLError as url_error
+        from urllib.proxy_handler import ProxyHandler as proxy_handler
     try:
         resp = url_open(url)
         headers = resp.info()
+    except url_error as e:
+        pass
+    try:
+        print('Trying to download with Proxy')
```
peterbarker: Too noisy :-)

amilcarlucas: done

MAVProxy/modules/lib/mp_util.py
```diff
@@ -250,12 +250,20 @@ def download_url(url):
     if sys.version_info.major < 3:
         from urllib2 import urlopen as url_open
         from urllib2 import URLError as url_error
+        from urllib2 import ProxyHandler as proxy_handler
     else:
         from urllib.request import urlopen as url_open
         from urllib.error import URLError as url_error
+        from urllib.proxy_handler import ProxyHandler as proxy_handler
     try:
```
peterbarker: So... in the case that we succeed we drop through and try fetching from the proxy as well.  I'd call that sub-optimal.

peterbarker:
@amilcarlucas did you want to pursue this?

amilcarlucas:
I completly forgot about this one.... sorry. It's easy to fix I'll give it a try.

PR #617, opened 2019-04-23
Do not tell to "set moddebug 3" when it already has been done

amilcarlucas:


MAVProxy/mavproxy.py
```diff
@@ -348,7 +348,10 @@ def load_module(modname, quiet=False, **kwargs):
             if mpstate.settings.moddebug > 1:
                 import traceback
                 print(traceback.format_exc())
-    print("Failed to load module: %s. Use 'set moddebug 3' in the MAVProxy console to enable traceback" % ex)
+    help_traceback = ""
+    if mpstate.settings.moddebug != 3:
```
peterbarker: Maybe `< 3`?

wx4cb: why not:

if mpstate.settings.moddebug > 1:
   .....
else:
 ....
end

sorry, my python is rusty after a few beers :D

amilcarlucas: done

PR #621, opened 2019-05-02
MAVExplorer.py: use 'sorted()' instead of '.sort()' for Python3 compatibility

Williangalvani:
Without this, I get an error when I type "help" in the console.


peterbarker:
Merged, thanks!

PR #633, opened 2019-06-04
Created module movinghome

AndKe:
This module will update the home position of a vehicle with the position of the GCS.
A NMEA device, connected to the MAVProxy, will provide the position of the GCS, this module will then update the vehicle home position every x seconds if the position is changed by >x meters.

Very useful for operating from a ship.  - in case of connection loss, home is the latest known position of the ship.

### DIFF
--- a/MAVProxy/modules/mavproxy_movinghome.py
+++ b/MAVProxy/modules/mavproxy_movinghome.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python
+'''
+movinghome module
+André Kjellstrup, Norce
+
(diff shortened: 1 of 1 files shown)


AndKe:
@tridge - please pull?

tridge:
nice idea!
it would be nice to use the mp_util existing functions for GPS distance, but I merged anyway as this is obviously tested

peterbarker:
@AndKe A few tweaks here and we can send `GLOBAL_POSITION_INT` from the GCS.  This will let us use Copter's `FOLLOW` mode to follow the GCS...

PR #636, opened 2019-06-08
Console dynamic field display

Akshath-Singhal:
This feature allows user to add fields for displaying data to the console dynamically using the following command:

console_add UNIQUE_IDENTIFIER DISPLAYNAME%variabletype MESSAGE_TYPE EXPRESSION ROW

eg. console_add WPBEAR TARBEAR%u NAV_CONTROLLER_OUTPUT (msg.target_bearing*10) 4

The above command will display the following

console

### DIFF
--- a/MAVProxy/modules/mavproxy_console.py
+++ b/MAVProxy/modules/mavproxy_console.py
@@ -25,6 +25,8 @@ class ConsoleModule(mp_module.MPModule):
         self.max_link_num = 0
         self.last_sys_status_health = 0
         self.last_sys_status_errors_announce = 0
+        self.user_added = []
+        self.add_command('console_add', self.cmd_add, "console module", ['add'])
         mpstate.console = wxconsole.MessageConsole(title='Console')
 
         # setup some default status information
@@ -76,6 +78,11 @@ class ConsoleModule(mp_module.MPModule):
                                                MPMenuItem('Map', 'Load Map', '# module load map')]))
             self.vehicle_menu = MPMenuSubMenu('Vehicle', items=[])
             self.add_menu(self.vehicle_menu)
+
+    def cmd_add(self, args):
+        new_field = args[:]
+        self.user_added.append(new_field)
+        print self.user_added
 
     def add_menu(self, menu):
         '''add a new menu'''
@@ -517,6 +524,11 @@ class ConsoleModule(mp_module.MPModule):
         elif type == 'PARAM_VALUE':
             rec, tot = self.module('param').param_status()
             self.console.set_status('Params', 'Param %u/%u' % (rec,tot))
+
+        for i in range(0,len(self.user_added)):
+            if type == self.user_added[i][2]:
+                code = "self.console.set_status('"+self.user_added[i][0]+"','" +self.user_added[i][1]+"' %"+self.user_added[i][3]+", row = "+self.user_added[i][4]+")"
+                exec(code)
 
 def init(mpstate):
     '''initialise module'''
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_console.py
```diff
@@ -75,6 +77,16 @@ def __init__(self, mpstate):
             self.vehicle_menu = MPMenuSubMenu('Vehicle', items=[])
             self.add_menu(self.vehicle_menu)
 
+    def cmd_add(self, args):
+        new_field = []
+        new_field.append(args[0])
```
tridge: isn't this the same as
  new_field = args[:]
?

MAVProxy/modules/mavproxy_console.py
```diff
@@ -514,6 +526,11 @@ def mavlink_packet(self, msg):
                 alt_error = "%d%s" % (msg.alt_error, alt_error_sign)
             self.console.set_status('AltError', 'AltError %s' % alt_error)
             self.console.set_status('AspdError', 'AspdError %.1f%s' % (msg.aspd_error*0.01, aspd_error_sign))
+        for i in range(0,len(self.user_added)):
```
tridge: we only want to trigger on msgs that we know are in expressions the user has added, otherwise we waste a lot of cpu

MAVProxy/modules/mavproxy_console.py
```diff
@@ -514,6 +526,11 @@ def mavlink_packet(self, msg):
...
+            if type == self.user_added[i][2]:
+                code = "self.console.set_status('"+self.user_added[i][0]+"','" +self.user_added[i][1]+"' %"+self.user_added[i][3]+", row = "+self.user_added[i][4]+")"
+                exec(code)
```
tridge: why exec? we can just call set_status() directly after evaluating the expression
also needs to catch exceptions for bad expressions

MAVProxy/modules/mavproxy_console.py
Akshath-Singhal: Yes

MAVProxy/modules/mavproxy_console.py
```diff
@@ -75,6 +77,11 @@ def __init__(self, mpstate):
             self.vehicle_menu = MPMenuSubMenu('Vehicle', items=[])
             self.add_menu(self.vehicle_menu)
 
+    def cmd_add(self, args):
+        new_field = args[:]
+        self.user_added.append(new_field)
+        print self.user_added
```
tridge: need to use print() for py3

MAVProxy/modules/mavproxy_console.py
```diff
@@ -75,6 +77,11 @@ def __init__(self, mpstate):
             self.vehicle_menu = MPMenuSubMenu('Vehicle', items=[])
             self.add_menu(self.vehicle_menu)
 
+    def cmd_add(self, args):
+        new_field = args[:]
+        self.user_added.append(new_field)
```
tridge: better to create a new small class that has the broken out fields with names

MAVProxy/modules/mavproxy_console.py
```diff
@@ -515,6 +522,13 @@ def mavlink_packet(self, msg):
             self.console.set_status('AltError', 'AltError %s' % alt_error)
             self.console.set_status('AspdError', 'AspdError %.1f%s' % (msg.aspd_error*0.01, aspd_error_sign))
 
+        if type in [row[2] for row in self.user_added]:   
+            for i in range(0,len(self.user_added)):
+                if type == self.user_added[i][2]:
+                    val = eval(self.user_added[i][3])
```
tridge: need to catch exceptions for eval()

tridge:
I've fixed it up to be more efficient

PR #652, opened 2019-07-31
mavproxy_map: correct use of uninitialised symbol

peterbarker:


peterbarker:
This fixes things for MdB.

PR #654, opened 2019-07-31
mavproxy.py: limit the time we will wait for module unload method to …

peterbarker:
…complete

In the case that a module's unload method does not return we completely
lose MAVProxy's text input, which is not great.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -391,7 +391,13 @@ def unload_module(modname):
     for (m,pm) in mpstate.modules:
         if m.name == modname:
             if hasattr(m, 'unload'):
-                m.unload()
+                t = threading.Thread(target=lambda : m.unload(), name="unload %s" % modname)
+                t.start()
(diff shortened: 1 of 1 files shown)


MAVProxy/mavproxy.py
```diff
@@ -391,7 +391,11 @@ def unload_module(modname):
     for (m,pm) in mpstate.modules:
         if m.name == modname:
             if hasattr(m, 'unload'):
-                m.unload()
+                t = threading.Thread(target=lambda : m.unload(), name="unload %s" % modname)
+                t.start()
+                t.join(timeout=1)
+                if t.isAlive():
+                    print("unload on module %s did not complete" % m.name)
```
OXINARF: Shouldn't we kill the thread and return False here?

peterbarker: That's unclear to me.  I guess we can always add that later.

But I have increased the timeout to 5 seconds and returned false here after discussions with tridge.

PR #662, opened 2019-08-13
FLIGHT mode

Akshath-Singhal:
Highlight current flight mode. Resolved bugs on setting parameters and redraw based on search key.

### DIFF
--- a/MAVProxy/modules/mavproxy_paramedit/param_editor.py
+++ b/MAVProxy/modules/mavproxy_paramedit/param_editor.py
@@ -85,6 +85,7 @@ class ParamEditorMain(object):
     def __init__(self, mpstate):
         self.param_received = {}
         self.paramchanged = {}
+        self.fltmode_rc = None
         self.event_queue = multiproc.Queue()
         self.event_queue_lock = multiproc.Lock()
         self.gui_event_queue = multiproc.Queue()
@@ -144,7 +145,7 @@ class ParamEditorMain(object):
                 self.needs_unloading = True
 
     def mavlink_packet(self, m):
-        if m.get_type() == 'PARAM_VALUE':
+        if m.get_type() in ['PARAM_VALUE', 'RC_CHANNELS', 'RC_CHANNELS_RAW']:
             self.mavlink_message_queue.put(m)
 
     def process_mavlink_packet(self, m):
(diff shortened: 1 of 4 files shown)


Akshath-Singhal:
@ tridge I am working on adding categorical search of parameters. Using 'FLTMODE' in search key provides the desired behaviour for setting flight modes as we discussed earlier. Highlighting the parameter name makes it intuitive to use and does not confuse with parameter change which highlights the value column. Please have a look.
Screenshot from 2019-08-13 22-30-56

MAVProxy/modules/mavproxy_paramedit/param_editor.py
```diff
@@ -155,6 +156,12 @@ def process_mavlink_packet(self, m):
                 del self.paramchanged[m.param_id.upper()]
             self.gui_event_queue.put(ParamEditorEvent(
                 ph_event.PEGE_WRITE_SUCC, paramid=m.param_id.upper(), paramvalue=m.param_value))
+        if mtype in ['RC_CHANNELS', 'RC_CHANNELS_RAW']:
+            rc_received = float(getattr(m, 'chan%u_raw' % self.mpstate.module('param').mav_param['FLTMODE_CH']))
```
tridge: we need to range check FLTMODE_CH. The user can set it to zero (or any invalid channel) to disable changing modes using RC. Also, the name of the parameter varies with firmware, on rover it is MODE_CH and the modes are MODE1 to MODE6, so we'll need an abstraction for that

MAVProxy/modules/mavproxy_paramedit/param_editor.py
```diff
@@ -85,6 +85,7 @@ class ParamEditorMain(object):
     def __init__(self, mpstate):
         self.param_received = {}
         self.paramchanged = {}
+        self.fltmode_rc = 1100
```
tridge: I think it should default to None, and check for None in the code that uses it

MAVProxy/modules/mavproxy_paramedit/param_editor_frame.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, *args, **kwds):
         self.requires_redraw = False
         self.last_grid_update = time.time()
         self.htree = {}
+        self.selected_fltmode = 1
```
tridge: better to use None by default I think

MAVProxy/modules/mavproxy_paramedit/param_editor.py
Akshath-Singhal: Corrected

MAVProxy/modules/mavproxy_paramedit/param_editor_frame.py
Akshath-Singhal: Done

MAVProxy/modules/mavproxy_paramedit/param_editor.py
Akshath-Singhal: Done

MAVProxy/modules/mavproxy_paramedit/param_editor.py
```diff
@@ -155,6 +156,28 @@ def process_mavlink_packet(self, m):
                 del self.paramchanged[m.param_id.upper()]
             self.gui_event_queue.put(ParamEditorEvent(
                 ph_event.PEGE_WRITE_SUCC, paramid=m.param_id.upper(), paramvalue=m.param_value))
+        if mtype == 'RC_CHANNELS_RAW':
```
tridge: why not use:
  if mtype in ['RC_CHANNELS_RAW', 'RC_CHANNELS']
as its just copy&paste

Akshath-Singhal: It's because of channel numbers. Both contain different number of channels. I will change it to remove code duplication to as much extent as possible.

Akshath-Singhal: Done.

tridge:
I have merged, but I would like to see a drop-down list of parameter categories that limits the parameters displayed (shortcut to search but on param names only)
So user can select "FlightModes" in the drop down
Then also have "Compass", "INS", "Battery" and other key categories

PR #670, opened 2019-08-22
mavproxy_misseditor: ignore non-mission-related mavlink packets

peterbarker:
Stops the waypoint editor getting confused as rally points are transferred with `MISSION_ITEM` messages now.

### DIFF
--- a/MAVProxy/modules/mavproxy_misseditor/mission_editor.py
+++ b/MAVProxy/modules/mavproxy_misseditor/mission_editor.py
@@ -220,6 +220,9 @@ class MissionEditorMain(object):
 
 
     def mavlink_packet(self, m):
+        if (getattr(m, 'mission_type', None) is not None and
+            m.mission_type != mavutil.mavlink.MAV_MISSION_TYPE_MISSION):
+            return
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_misseditor/mission_editor.py
```diff
@@ -229,6 +232,9 @@ def process_mavlink_packet(self, m):
 
         # if you add processing for an mtype here, remember to add it
         # to mavlink_packet, above
+        if (getattr(m, 'mission_type', None) is not None and
+            m.mission_type != mavutil.mavlink.MAV_MISSION_TYPE_MISSION):
+            return
```
OXINARF: Is this one really needed? I mean, if it isn't put in the queue (given the new code in `mavlink_packet`), then it doesn't get here, right?

peterbarker:
On Thu, 22 Aug 2019, Francisco Ferreira wrote:

> +        if (getattr(m, 'mission_type', None) is not None and
> +            m.mission_type != mavutil.mavlink.MAV_MISSION_TYPE_MISSION):
> +            return
> 
> Is this one really needed? I mean, if it isn't put in the queue (given the new code in mavlink_packet), then it doesn't get here, right?

Being defensive.

PR #671, opened 2019-08-23
Fix typo, use rem instead of rm

ppicazo:


ppicazo:
Result of build:
```
Uploading artifacts...
[1/2] WindowsCompiledFiles...Zipping to WindowsCompiledFiles.zipMaximum allowed artifact storage size of 50000 Mb will be exceeded
```

peterbarker:
Merged, thanks.

Someone is actively working on the CI problem, thanks for the heads up.

PR #672, opened 2019-08-23
CI: Only keep artifacts for master branch builds

stephendade:
Attempt to fix the failing builds - The build files will only be saved for master branch builds.

It may still fail, as there is a 50000MB file limit and Appveyor keeps the last 6 months of builds. We may have to wait until Appveyor cleans up some of these old builds.

[Doesn't appear to be a method to manually remove old builds though :disappointed: ]

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -78,13 +78,16 @@ after_test:
   - "ISCC.exe /dMyAppVersion=%VERSION% mavproxy.iss"
   - "cd ..\\"
   
-
-artifacts:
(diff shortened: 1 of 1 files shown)


peterbarker:
Still dying with `Maximum allowed artifact storage size of 50000 Mb will be exceeded.`

OXINARF:
I think we might need to contact AppVeyor to clean it up.

tridge:
i've also contacted appveyor, and they have increased our limit

PR #673, opened 2019-08-23
Parameditor: Categorical search of parameters

Akshath-Singhal:
Added categorical search of parameters with a predefined list of categories.

@tridge  Might need a work around for Flight Modes on Rover. Using only MODE displays extra parameters.

### DIFF
--- a/MAVProxy/modules/mavproxy_paramedit/checklisteditor.py
+++ b/MAVProxy/modules/mavproxy_paramedit/checklisteditor.py
@@ -155,7 +155,7 @@ class GridScrollEditor(gridlib.PyGridCellEditor):
         try:
             self._tc.SetValue(selected)
         except Exception as e:
-            print e
+            print (e)
 
     def SetSize(self, rect):
         self._tc.SetDimensions(rect.x, rect.y, rect.width+2, 30,
--- a/MAVProxy/modules/mavproxy_paramedit/param_editor.py
+++ b/MAVProxy/modules/mavproxy_paramedit/param_editor.py
@@ -48,7 +48,6 @@ class ParamEditorEventThread(threading.Thread):
                         self.mp_paramedit.gui_event_queue_lock.release()
 
                     elif event_type == ph_event.PEE_TIME_TO_QUIT:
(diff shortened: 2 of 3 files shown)


Akshath-Singhal:
@tridge The second commit basically deals with the special case of modes in ROVERS. I will squash the commits once we agree on using this method. Also you might want to list the categories to be added to the dropdown menu here.

Akshath-Singhal:
@tridge Both the search key and the category complements each other now. Seems ready to be merged to me.

tridge:
the categories work, but I think the drop-down should only have the bit before the ':', so "PosControl" not "PosControl:VEL_,POS_,WPNAV_,RTL_"
It is also using substring searches, so the POS_ in PosControl is finding EK2_POS_I_GATE. Substring searches are great for the search box, but not good in the dropdown.
I've also noticed that the param editor is failing completely under python3 (specifically 3.6.8 on ubuntu 18.04). The module loads but doesn't display anything.
I also see some bugs by running flake8 over the code. For example:

MAVProxy/modules/mavproxy_paramedit/param_editor.py:219:12: F821 undefined name 'ParamEditorModule'

Akshath-Singhal:
> the categories work, but I think the drop-down should only have the bit before the ':', so "PosControl" not "PosControl:VEL_,POS_,WPNAV_,RTL_"
> It is also using substring searches, so the POS_ in PosControl is finding EK2_POS_I_GATE. Substring searches are great for the search box, but not good in the dropdown.
> I've also noticed that the param editor is failing completely under python3 (specifically 3.6.8 on ubuntu 18.04). The module loads but doesn't display anything.
> I also see some bugs by running flake8 over the code. For example:
> 
> MAVProxy/modules/mavproxy_paramedit/param_editor.py:219:12: F821 undefined name 'ParamEditorModule'

Corrected the typo in Flake8. Though it had no effect.
Python3 error was due to unbuffered I/O for /dev/null, it has been corrected.
Displays only upto ':' for category now, substring search removed from categories.

tridge:
much better, thanks!

PR #674, opened 2019-08-24
Paramstatus: Display current status of received parameters to help with param fetch

Akshath-Singhal:
@tridge @peterbarker This displays parameter status both in console and in parameditor modules to prevent user from requesting parameter fetch repeatedly as discussed earlier.

PS : Three different commits for three different modules as suggested earlier.


tridge:
great! I'd like to see the commit msgs prefixed with module name, eg:
"console: display param status"

Akshath-Singhal:
Done

PR #682, opened 2019-09-04
MAVExplorer: added paramchange command

tridge:
useful for seeing changes in flight

### DIFF
--- a/MAVProxy/tools/MAVExplorer.py
+++ b/MAVProxy/tools/MAVExplorer.py
@@ -625,6 +625,34 @@ def cmd_param(args):
         if fnmatch.fnmatch(str(p).upper(), wildcard.upper()):
             print("%-16.16s %f" % (str(p), mestate.mlog.params[p]))
 
+def cmd_paramchange(args):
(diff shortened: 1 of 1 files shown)


MAVProxy/tools/MAVExplorer.py
```diff
@@ -625,6 +625,34 @@ def cmd_param(args):
...
+            break
+        if m.get_type() == 'PARM':
+            pname = m.Name
+            pvalue = m.Value
+        elif m.get_type() == 'PARAM_VALUE':
+            pname = m.param_id
+            pvalue = m.param_value
+        else:
+            continue
+        if pname.startswith('STAT_'):
```
peterbarker: We have a list of differently-uninteresting parameters here: https://github.com/ArduPilot/pymavlink/master/mavparm.py#L14 - there might be a few more to add here.

tridge: I really just want to filter out ones that change frequently in flight. STAT_ is good for that

PR #689, opened 2019-09-11
mavproxy_map: Fix issue on Windows where tiles and srtm were not cached

stephendade:
Turns out that srtm and tiles were not actually being cached in Windows. Also, fixed some srtm downloading errors.

### DIFF
--- a/MAVProxy/modules/mavproxy_map/srtm.py
+++ b/MAVProxy/modules/mavproxy_map/srtm.py
@@ -23,7 +23,6 @@ import array
 import math
 from MAVProxy.modules.lib import mp_util
 from MAVProxy.modules.lib import multiproc
-import tempfile
 
 childTileDownload = {}
 childFileListDownload = {}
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/mavproxy_map/srtm.py
```diff
@@ -75,7 +74,7 @@ def __init__(self, server="firmware.ardupilot.org",
             try:
                 cachedir = os.path.join(os.environ['HOME'], '.tilecache/SRTM')
             except Exception:
-                cachedir = os.path.join(tempfile.gettempdir(), 'MAVProxySRTM')
+                cachedir = os.path.join(os.environ['LOCALAPPDATA'], '.tilecache/SRTM')
```
tridge: I think we need to check if 'LOCALAPPDATA' exists in os.environ, otherwise we'll get another nested exception

stephendade: Fixed

PR #691, opened 2019-09-14
rc: allow override rate to be set

tridge:


MAVProxy/modules/mavproxy_rc.py
```diff
@@ -99,6 +113,9 @@ def cmd_rc(self, args):
         if args[0] == 'all':
             for i in range(self.count):
                 channels[i] = value
+        elif args[0] == 'clear':
```
peterbarker: Was this supposed to do something with `args[1]`?  Or should this just be removed entirely?

tridge: fixed

PR #693, opened 2019-09-29
mavproxy: radio setup now works under Python3

stephendade:
The ``--setup`` argument now works under Python3 - just needed some str->byte conversions.


tridge:
thanks Stephen!

PR #696, opened 2019-10-11
MAVExplorer: added time delay

yonhonng:
added a function while graphing by doing:
graph delay <TYPE> <SECONDS> <FIELDS...>


MAVProxy/modules/lib/grapher.py
```diff
@@ -364,16 +367,17 @@ def timestamp_to_days(self, timestamp):
...
             except ValueError:
                 # this can happen if the log is corrupt
                 # ValueError: year is out of range
                 return 0
         sec_to_days = 1.0 / (60*60*24)
-        return self.tday_base + (timestamp - self.tday_basetime) * sec_to_days
+        return self.tday_base + (timestamp - self.tday_basetime + self.timeshift) * sec_to_days
 
     def process_mav(self, mlog, flightmode_selections):
+        print("processing now...\n")
```
tridge: don't need this debug I think?

tridge:
I just tried this and it doesn't seem to work? 
I tried:
 delay NKF1.Roll 0.5
 graph ATT.Roll NKF1.Roll
I expected to see NKF1.Roll off by 0.5s

tridge:
ahh, works only at msg level, so this works "delay NKF1 0.5"

PR #697, opened 2019-10-11
Dataflash_logger: add setting to automatically rotate logs on disarm

RickReeser:
This will make the dataflash-over-mavlink module automatically rotate logs when the vehicle disarms if the ArduPilot parameter `LOG_FILE_DSRMROT` is set. Tested with ArduCopter 3.6.9.

My change is pretty naive, so let me know if there's a smarter way to do this.

### DIFF
--- a/MAVProxy/modules/mavproxy_dataflash_logger.py
+++ b/MAVProxy/modules/mavproxy_dataflash_logger.py
@@ -222,7 +222,7 @@ class dataflash_logger(mp_module.MPModule):
             self.armed = isarmed
             dsrmrot = self.get_mav_param('LOG_FILE_DSRMROT', 0)
             if not self.armed and dsrmrot == 1:
-                self.rotate_log()          
+                self.rotate_log()
                 
         if self.log_settings.verbose:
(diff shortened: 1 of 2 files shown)


peterbarker:
I do like the idea of having the option of rotating these logs on disarm, but perhaps this would be better done as a setting within the dataflash_logger module within mavproxy?

So

`dataflash_logger set rotate_on_disarm True`

I don't think there's a good reason to reuse the on-board configuration.

RickReeser:
Yeah, I was waffling between using the FC param or a module setting. Module setting is better. Changed and committed.

peterbarker:
@RickReeser Looks good, except for the commit list.  Could you squash the commits together into a singlet commit with a message starting with, "dataflash_logger: ", please?

RickReeser:
Can you merge-n-squash? Won't have access to terminal for a while, and I can't squash with the browser.

peterbarker:
@RickReeser rebased, squash, fixed minor whitespace issue, added good commit message and pushed in - thanks!

PR #705, opened 2019-11-27
log: skip already-downloaded logs

AndKe:
This change makes "log download all" skip existing files.
very useful when there are hundreds of files, and download needs to be resumed.


tridge:
good idea, thanks!

peterbarker:
Probably could have done a size match here.  OTOH, not overwriting existing files might also be a feature.

AndKe:
@peterbarker you are right, I deleted last partially downloaded file manually. I may consider to make a size comparsion in the future, and also a file start-number that lets user download only log numbers above some size.

PR #708, opened 2019-12-01
mavproxy_link: understand statustext.moretext

peterbarker:


meee1:
what is the purpose of mavlink_statustext_long_t then?

peterbarker:
On Sun, 1 Dec 2019, Michael Oborne wrote:

> what is the purpose of mavlink_statustext_long_t then?

Same thing.  It's mavlink2 only, 'though.

Sensibly we can only emit one message (STATUSTEXT_LONG or STATUSTEXT) per 
link.  Extending STATUSTEXT means the GCSs that haven't gotten around to 
implementing the extension still get the base text.

The ArduPilot side is marked as WIP and RFC, so if you think this is a bad 
idea, please say so.  It was just easier to explain in code....

peterbarker:
Closing in favour of https://github.com/ArduPilot/ardupilot/pull/13039

PR #711, opened 2019-12-05
MAVFlightView: Increase number of decimals in Lat/Lon and Distance Measurements

jmachuca77:
This PR increases the number of decimals in the lat/lon coordinate labels as well as adding more decimals to the distance measurement on the map plot. This allows for precise measurements for checking RTK navigation accuracy


peterbarker:
Merged, thanks!

PR #713, opened 2019-12-09
mavproxy_link: add statustext chunk amalgamation

peterbarker:
demonstration code for reassembling long statustext messages to go along with ArduPilot PR

https://github.com/ArduPilot/ardupilot/pull/13010


peterbarker:
We decided to go with the other implementation of this.

PR #715, opened 2019-12-09
Graph: shows error message instead of exception when called on array fields

EbinPhilip:
When graph is called on an array type field, without using an index eg: WHEEL_DISTANCE.distance, MAVProxy goes into a loop of exceptions. This patch checks if the data is of array(list) type and informs the user.

### DIFF
--- a/MAVProxy/modules/lib/live_graph_ui.py
+++ b/MAVProxy/modules/lib/live_graph_ui.py
@@ -167,6 +167,11 @@ class GraphFrame(wx.Frame):
         if self.paused:
             return
         for i in range(len(self.plot_data)):
+            if (type(state.values[i]) == list):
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/lib/live_graph_ui.py
```diff
@@ -166,6 +166,13 @@ def on_redraw_timer(self, event):
             state.values = state.child_pipe.recv()
         if self.paused:
             return
+
+        if (type(state.values[0]) == list):
```
peterbarker: Needs to work across all values, not just the first.

EbinPhilip: I've moved the check into the for loop below

peterbarker:
Merged, thanks!

PR #717, opened 2019-12-10
fixes failed import of rtcm3 module

petergerten:


MAVProxy/modules/lib/ntrip.py
```diff
@@ -10,7 +10,7 @@
 import base64
 import time
 import errno
-import rtcm3
+from . import rtcm3
```
tridge, line 13: I don't think this works with py3

PR #728, opened 2020-02-03
Add explicit closing of ports in mpstate.mav_outputs list

apechacek:
When UDP ports in the `mpstate.mav_output` list are not closed explicitly, they can hang for an extended period of time before they get cleaned up. This causes port already in use errors if MAVProxy is relaunched too quickly.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1223,6 +1223,13 @@ if __name__ == '__main__':
 
     def quit_handler(signum = None, frame = None):
         #print('Signal handler called with signal', signum)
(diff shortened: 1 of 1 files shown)


peterbarker:
@apechacek These should be being opened with `SO_REUSEADDR` - as soon as the original process exits the port should be bindable again.

How were you reproducing the problem?

apechacek:
@peterbarker I switched the option from --out to --sitl and it resolve the issue without this change. The `mavutil.mavlink_connection` was defaulting to TCP because it was unspecified. Thanks for your input. Closing this now.

PR #730, opened 2020-02-04
handle pymavlink without STATUSTEXT chunks

tridge:


peterbarker:
Pushed up another patch to work around the "id field already exists" problem.

PR #731, opened 2020-02-09
Fix GoPro module syntax

Pedals2Paddles:
I may be the only one to ever use this in the last several years, and probably the next several years to come. But it's fixed. Thx Tridge for getting me squared away with this.

- Use vehicle system ID since Solo gimbal is a private channel that ignores broadcast system and component messages.
- Use array for message param values.


peterbarker:
merged, thanks!

PR #737, opened 2020-03-09
Add TCP Connection to DGPS

jmachuca77:
Emlid Reach GPS can only publish the RTCM corrections using TCP or Serial connections. This adds the ability to select between UDP and TCP and allows configuration of the IP, PORT, CONNTYPE. It adds four commands (following the ntrip modules model):
start
stop
status
set

by default it is configure the same as it was before (UDP, 127.0.0.1, 13320)

I don't have a GPS that publishes corrections via UDP to test, but it should work. I did test the TCP connection


MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -1,24 +1,119 @@
...
-        self.inject_seq_nr = 0
         print("DGPS: Listening for RTCM packets on UDP://%s:%s" % ("127.0.0.1", self.portnum))
-    
+
+    def connect_tcp_rtcm_base(self, ip, port):
+        print "TCP Connection"
+        try:
+            self.port = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            self.port.connect((ip, port))
+            self.port.settimeout(1)
```
tridge: need to make this non-blocking

jmachuca77: I made the change and the lost packet issue did go away. Thanks for the suggestion Tridge. PR is updated.

MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -1,86 +1,208 @@
...
+        print ("TCP Connection")
+        try:
+            self.port = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            self.port.connect((ip, port))
+            self.port.settimeout(1)
+            self.port.setblocking(0)
+        except:
+            print ("ERROR: Failed to connect to RTCM base over TCP, retrying in 2.5s")
+            self.waiting = True
+            threading.Timer(2.5,self.connect_tcp_rtcm_base(ip, port)).start()
```
tridge, line 135: I don't think we want a timer here, instead use the idle task to check for a pending connection. See the start_pending code in the mavproxy_ntrip module

MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -1,86 +1,208 @@
...
         '''called in idle time'''
+        if self.port is None:
+            return
+
+        if self.waiting is True:
+            return
+
         try:
-            data = self.port.recv(1024) # Attempt to read up to 1024 bytes.
+            data = self.port.recv(1) # Attempt to read up to 1024 bytes.
```
tridge, line 149: comment doesn't match code, and why read one byte at a time?

jmachuca77: I am copying what is done in the NTRIP library:
https://github.com/ArduPilot/MAVProxy/blob/041dda21ddf9caf884abcef93396f7ba174dd913/MAVProxy/modules/lib/ntrip.py#L174

jmachuca77:
I have already added the status count for each packet, So that can be seen with the status command.

MAVProxy/modules/mavproxy_DGPS.py
jmachuca77: This is because the function https://github.com/ArduPilot/MAVProxy/blob/041dda21ddf9caf884abcef93396f7ba174dd913/MAVProxy/modules/lib/rtcm3.py#L60
takes one byte at a time.

jmachuca77:
@tridge This seems to be working pretty good now, let me know if you want me to squash the three commits into one, but I think its ready to merge.

jmachuca77:
Squashed

tridge:
please test with both py2 and py3

amilcarlucas:
This needs a rebase and some more testing.
Part of the functionality is a duplicate from #408

stw2nf:
do you have a build that I could test/run? Really interested in this feature. we have a Emlid Reach M2 we would like to automate RTK corrections with via Rpi, but having troubles rebasing this branch on master and actually testing functionality

PR #750, opened 2020-04-10
change deprecated isAlive() to is_alive()

acxz:
isAlive() has been deprecated in py3.8 and removed in py3.9. is_alive() has been available since python2.7.

See: https://docs.python.org/3.9/whatsnew/3.9.html (ctrl-f isAlive()) for `isAlive()` removal
and https://docs.python.org/2/library/threading.html for `is_alive()` in python2.6 and above.


peterbarker:
Merged, thanks!

PR #761, opened 2020-05-04
disable periodical setting stream rates if 'mavfwd_rate' is True

standmit:
While setting `mavfwd_rate` to `True` we assume that we will change stream rates or messages intervals. Therefore this settings shouldn't be overriden every few seconds.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -953,7 +953,8 @@ def periodic_tasks():
     if heartbeat_check_period.trigger():
         check_link_status()
 
-    set_stream_rates()
+    if mpstate.settings.mavfwd_rate == False:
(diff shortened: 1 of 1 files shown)


MAVProxy/mavproxy.py
```diff
@@ -953,7 +953,8 @@ def periodic_tasks():
     if heartbeat_check_period.trigger():
         check_link_status()
 
-    set_stream_rates()
+    if mpstate.settings.mavfwd_rate == False:
```
peterbarker, line 956: This should be `is`, not `==`

standmit:
> This is nothing you couldn't achieve by the traditional `set streamrate -1`, right?

Really. I overlooked it. Closing

PR #773, opened 2020-06-15
joystick: add Spektrum InterLink DX

muramura:
I have a Spektrum InterLink DX joystick available for use with Real Flight 9.
I want to use this joystick in SITL.

Spektrum InterLink DX:.
https://www.horizonhobby.com/interlink-dx-simulator-controller-usb-plug-spmrftx1


stephendade:
Thanks! Merged

PR #779, opened 2020-06-21
Add SteamVR module for tracking 

IamPete1:
This adds a module that reads in the pose from a SteamVR Tracker using [PYopenVR](https://github.com/cmbruns/pyopenvr). The tracker provides position, velocity, attitude and attitude rate at 144HZ. This is then dished out in GPS or VISION messages at a set rate.

https://youtu.be/P7J486V5__A

My first MAVProxy PR and also my first proper bash as Python so apologies for any mistakes.

I ended up using a OpenVR example code to go from the provided matrix to roll pitch and yaw. I couldn't workout the ordering to use the existing one.

I will do a better vid explaining how to setup in the next day or so.


MAVProxy/modules/mavproxy_steamvr.py
```diff
@@ -0,0 +1,304 @@
...
+use SteamVR tracker data to provide VISION_POSITION_ESTIMATE and GPS_INPUT data
+"""
+
+import math
+import threading
+import time
+
+from MAVProxy.modules.lib import mp_module
+from MAVProxy.modules.lib import mp_settings
+from MAVProxy.modules.lib import LowPassFilter2p
```
stephendade, line 11: I can't find this module in MAVProxy. Did you mean to add this module too?

IamPete1: Opps, that is left over from vicom, i used that module as a starting point

IamPete1:
https://discuss.ardupilot.org/t/aggressive-indoor-copter-flying-with-steamvr-tracking/58186

MAVProxy/modules/mavproxy_steamvr.py
```diff
@@ -0,0 +1,303 @@
...
+
+            mav = self.master
+            time_us = int(now * 1.0e6)
+            if now - last_origin_send > 1 and self.vr_settings.vision_rate > 0:
+                # send a heartbeat msg
+                mav.mav.heartbeat_send(mavutil.mavlink.MAV_TYPE_GCS, mavutil.mavlink.MAV_AUTOPILOT_GENERIC, 0, 0, 0)
+
+                # send origin at 1Hz
+                mav.mav.set_gps_global_origin_send(self.target_system,
+                                                   int(self.vr_settings.origin_lat*1.0e7),
```
stephendade, line 201: I thought this only had to be sent once?

MAVProxy/modules/mavproxy_steamvr.py
```diff
@@ -0,0 +1,303 @@
...
+                                                        pos_ned.y, pos_ned.x)
+                    gps_alt = self.vr_settings.origin_alt - pos_ned.z
+
+                    gps_week, gps_week_ms = get_gps_time(now)
+
+                    if self.vr_settings.gps_nsats >= 6:
+                        fix_type = 3
+                    else:
+                        fix_type = 1
+                    mav.mav.gps_input_send(time_us, 0, 0, gps_week_ms, gps_week, fix_type,
```
stephendade, line 227: Why does gps_input need to be sent if you're also sending global_vision_position_estimate?

MAVProxy/modules/mavproxy_steamvr.py
```diff
@@ -0,0 +1,303 @@
+"""
+use SteamVR tracker data to provide VISION_POSITION_ESTIMATE and GPS_INPUT data
+"""
```
stephendade, line 3: In the header comments: add in any extra packages required for this module (ie openvr) and links to any special install/setup instructions

MAVProxy/modules/mavproxy_steamvr.py
```diff
@@ -0,0 +1,303 @@
...
+                mav.mav.global_vision_position_estimate_send(time_us,
+                                                            pos_ned.x, pos_ned.y, pos_ned.z,
+                                                            roll, pitch, yaw, force_mavlink1=True)
+                vision_frame_count += 1
+
+                if self.vr_settings.vision_speed:
+                    # send VISION_SPEED_ESTIMATE
+                    mav.mav.vision_speed_estimate_send(time_us, vel_ned.x, vel_ned.y, vel_ned.z, force_mavlink1=True)
+
+    def cmd_start(self):
```
stephendade, line 250: Add something user-friendly if openvr fails to start, like "Unable to start OpenVR. Check OpenVR is installed and configured".

MAVProxy/modules/mavproxy_steamvr.py
IamPete1: It just gives the option, by default the gps stream rate is 0

MAVProxy/modules/mavproxy_steamvr.py
IamPete1: It only has to be received once, I guess this is easier than trying to tell. This is completely left over from the Vicon stuff.

MAVProxy/modules/mavproxy_steamvr.py
```diff
@@ -0,0 +1,303 @@
...
+             ('vision_speed', bool, 1),
+             ('gps_rate', int, 0),
+             ('gps_nsats', float, 16),
+             ('yaw_offset', float, 0.0)])
+        self.add_command('steamvr', self.cmd_steamvr, 'SteamVR control',
+                         ["<start>",
+                          "<stop>",
+                          "set (STEAMVRSETTING)"])
+        self.add_completion_function('(STEAMVRSETTING)',
+                                     self.settings.completion)
```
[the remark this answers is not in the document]
IamPete1: fixed and tested, to be honest I had never tested the settings

MAVProxy/modules/mavproxy_steamvr.py
IamPete1: I have added a try: exception: round this, it seems to work but I'm not sure it that is the best way

IamPete1:
I have made this a good deal more robust, stop and start work properly now, the only way I have found to break it is to shut steam vr while its being used. I think this is more due to issues with PyOpenVR than this. SteamVR does give a 'are you sure python still running' warning so I don't think its a disaster.

PR #781, opened 2020-07-01
Allow file dialog boxes to see MP param files

Hwurzburg:


MAVProxy/modules/mavproxy_paramedit/param_editor_frame.py
```diff
@@ -410,7 +410,7 @@ def getinfo(self, name):
 
     def Read_File(self, event):  # wxGlade: ParamEditor.<event_handler>
         fd = wx.FileDialog(self, "Open Parameter File", os.getcwd(), "",
-                           "*.parm|*", wx.FD_OPEN | wx.FD_FILE_MUST_EXIST)
```
peterbarker, line 413: Could this be `*.parm|*.param|*` instead?

Hwurzburg: not sure....right now the dialog only can show .parm files.....the expression seems to indicate that its that or anything...the "or" does not work....at least in Ubuntu18.04....there are no drop downs for"All files"  as you would expect in most systems and that the "|" implied....so I dont think it will work

peterbarker:
I've adjusted the wildcards here so they'll only match `*.parm` and `*.param`.

peterbarker:
Merged, thanks!

PR #782, opened 2020-07-01
Add correct types to all file dialogs 

Hwurzburg:
just set it to "\*.*"  files   MAVproxy itself stores fences as .txt files....Mission Planner as *.fen.....MAVProxy fence module only shows .fen... until now
built and tested


Hwurzburg:
reviewed and updated all file dialogs.
current dialogs are inconsistent with Mission
Planner file extensions and even MAVProxy's
own backup files!
.waypoints file extentions will need to be added to the dialogs , if/when the new rally point and complex fence capabilities are added
tested locally

peterbarker:
Merged, thanks!

Issue #783, opened 2020-07-15
MAVProxy Windows Failed to execute script pyi_rth_pkgres


joshferns:
OS: Windows 10
MAVProxy version : 1.8.20 (installed using MAVProxySetup-latest.exe)
On running mavproxy.exe from terminal, get the following error
```
Traceback (most recent call last):
  File "site-packages\PyInstaller\loader\rthooks\pyi_rth_pkgres.py", line 13, in <module>
  File "C:\Python35\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 623, in exec_module
  File "site-packages\pkg_resources\__init__.py", line 85, in <module>
ImportError: No module named 'pkg_resources.py2_warn'
[10700] Failed to execute script pyi_rth_pkgres
```
I have been able to run the previous available version  1.8.6  successfully.

stephendade:
Yes, I'm working on that issue (and other Windows compatibility issues). Should have a fix in master in the next few weeks.

stephendade:
Fixed in #786

PR #786, opened 2020-07-24
Make Windows work

stephendade:
This is a bunch of commits to get MAVProxy working natively on Windows again.

This includes:
-Updating the Appveyor build chain to Python 3.8
-Adding missing packages (see #783)
-Including parameter xml's in Windows installer
-Changing the text console to use Prompt-toolkit rather than readline.
-Make parameditor use threads rather than processes

The latter two changes make liberal use of ``if platform.system() == 'Windows':`` to ensure Linux behaviour remains the same.


stephendade:
@tridge - Are you OK with this patch? I'd like to get it merged in and a non-broken MAVPorxy build up on the ArduPilot firmware site.

MAVProxy/mavproxy.py
```diff
@@ -1364,7 +1360,7 @@ def quit_handler(signum = None, frame = None):
         start_script = os.path.join(os.environ['HOME'], ".mavinit.scr")
         start_scripts.append(start_script)
     if 'LOCALAPPDATA' in os.environ and not opts.setup:
-        start_script = os.path.join(os.environ['LOCALAPPDATA'], "MAVProxy", "mavinit.scr")
+        start_script = os.path.join(os.environ['LOCALAPPDATA'], ".mavproxy", "mavinit.scr")
```
peterbarker, line 1363: Is it appropriate to use a dot-directory in `LOCALAPPDATA`?

I see you've marked it as "consolidate" - but it doesn't seem right to use a dot-directory in there....

windows/MAVProxyWinBuild.bat
```diff
@@ -1,11 +1,11 @@
 rem build the standalone MAVProxy.exe for Windows.
-rem This assumes Python is installed in C:\Python36
+rem This assumes Python is installed in C:\Program Files (x86)\Python37-32
```
peterbarker, line 2: This is odd considering the 3.8 version mentioned earlier.

We're testing against a version we're not requiring in the intall?

MAVProxy/mavproxy.py
stephendade: The ``.mavproxy`` is used to store the parameter definitions from ``param download``. I was consolidating touse that folder, rather than have multiple place that MAVProxy stores data in the user home folder.

windows/MAVProxyWinBuild.bat
stephendade: Fixed - now just using the user's default python version.
Note this script is not used in the appveyor CI build. This is primarily for developers to build locally.

stephendade:
@peterbarker - Have now tested on Python 2 and 3 on both Windows10 and Ubuntu18. There were a few Python2/Windows10 issues, but I've now resolved them.
Testing = checking for prompt history and autocomplete, plus loading the parameditior module.

peterbarker:
I've run this locally a bit now and it didn't cause me problems.

Merging, thanks!

PR #788, opened 2020-07-30
modules: devop: add support for bank selection

bugobliterator:


tridge:
pending PR in ArduPilot, with xml change

PR #789, opened 2020-08-03
mavproxy_misseditor: save misseditor layout with layout save

sjw58194:
Since commands `layout save` and `layout load` only apply to console and map at the moment, I modified the mission editor module to be included when these commands are run. 

I also removed missionEditor.wxg since missionEditorFrame.py can generate the GUI by itself.


tridge:
please test both py2 and py3

sjw58194:
Slightly modified the code and tested it. It now works in both Python2 and Python3.

peterbarker:
Merged, thanks!

PR #790, opened 2020-08-04
Modules: fix help module errors when no Internet

stephendade:
Fix issue where the help module would produce errors if there was no Internet connection on startup.

### DIFF
--- a/MAVProxy/modules/mavproxy_help.py
+++ b/MAVProxy/modules/mavproxy_help.py
@@ -1,7 +1,7 @@
 """
     MAVProxy help/versioning module
 """
-import os, time, platform, re, sys
+import os, time, platform, re, sys, socket
 from pymavlink import mavwp, mavutil
 from MAVProxy.modules.lib import mp_util
 from MAVProxy.modules.lib import mp_module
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_help.py
```diff
@@ -45,13 +45,15 @@ def __init__(self, mpstate):
...
         pypi = xmlrpclib.ServerProxy('https://pypi.python.org/pypi')
-        available = pypi.package_releases('MAVProxy')
-        if not available:
+        try:
+            available = pypi.package_releases('MAVProxy')
+            if not available:
+                self.newversion = 'Error finding update'
+            else:
+                self.newversion = available[0]
+        except:
```
peterbarker, line 55: Please try to avoid catching Exception like this; try to catch the relevant subset of exceptions.

MAVProxy/modules/mavproxy_help.py
```diff
@@ -45,13 +45,15 @@ def __init__(self, mpstate):
...
 
         #check for updates, if able
-        import pip
         pypi = xmlrpclib.ServerProxy('https://pypi.python.org/pypi')
-        available = pypi.package_releases('MAVProxy')
-        if not available:
+        try:
+            available = pypi.package_releases('MAVProxy')
+            if not available:
+                self.newversion = 'Error finding update'
```
peterbarker, line 52: Setting this to `None` outside here would be better.

So `if available:` sort of thing.

Checking for `None` would then be done down in the "format the update string" step.

MAVProxy/modules/mavproxy_help.py
stephendade: Fixed

MAVProxy/modules/mavproxy_help.py
stephendade: Fixed

PR #799, opened 2020-08-17
Add ability to plot data from OBSTACLE_DISTANCE and DISTANCE_SENSOR

peterbarker:


rmackay9:
This is really great to see.  I wonder if it would be possible to display the output from the obstacle distance message using small arcs instead of circles?  I think the circles could be confused with the obstacle information we send using ADSB messages.

peterbarker:
I've now tested this on a real-world RPLidar2 and it works well.

PR #803, opened 2020-08-31
mavproxy_horizon: repair the status text in artificial horizon

aardvaark111:
unknown

Restore the status indicator in the horizon module to working condition.
Currently never stops showing UNKNOWN in grey.


peterbarker:
Merged, thanks!

PR #805, opened 2020-09-06
horizon: fix roll beyond +-90 deg

vierfuffzig:
fixes sky / ground polygon switching at +-90° roll


peterbarker:
Merged, thanks!

PR #811, opened 2020-09-14
mavproxy_DGPS: Python3 fixes

peterbarker:
I do not have hardware to test this on.

Tested by throwing random data at the port and checking for obvious Python errors.  This testing was done on both Python2 and Python3.


peterbarker:
Closes #808

PR #814, opened 2020-09-14
mavproxy.py: look for .mavinit.scr in HOME

peterbarker:
This is very different to Windows, where we only look in the .mavproxy
directory (or equivalent)


peterbarker:
@stephendade for reference

PR #819, opened 2020-09-21
Python 3.8 / Windows Fix

stephendade:
Fix and enhancement:

- Fixed a text console failure on Python 3.8 / Windows. This is due to the prompt_toolkit library using asyncio, which was changed in Python 3.8 (see https://github.com/tornadoweb/tornado/issues/2608)
- Removed tcl/tk from Windows build as we don't use in Windows MAVProxy.


peterbarker:
Merged, thanks!

PR #821, opened 2020-09-22
CI: Use less storage space

stephendade:
This removes one of the generated artifacts from Appveyor, saving 50% storage.

The removed artifact isn't required, as it's just a capture of part-way through the build process (for debugging purposes, though I've never used it).


peterbarker:
Merged, thanks!

PR #828, opened 2020-10-25
Modules: Update URL for User Alerts

stephendade:
The URL for User Alerts was updated to be more sensible.


peterbarker:
Merged, thanks!

PR #830, opened 2020-11-06
Add support for macOS Catalina / Python3.8+

srmainwaring:
On macOS use `multiprocessing` instead of `billiard` if the Python version >= 3.8.
The custom Queue is still required as macOS does not implement functions needed to support `multiprocessing.Queue.qsize()`.

### DIFF
--- a/MAVProxy/modules/lib/mp_image.py
+++ b/MAVProxy/modules/lib/mp_image.py
@@ -171,12 +171,12 @@ class MPImage():
 
     def poll(self):
         '''check for events, returning one event'''
-        if self.out_queue.qsize() <= 0:
+        if self.out_queue.empty():
             return None
         evt = self.out_queue.get()
         while isinstance(evt, win_layout.WinLayout):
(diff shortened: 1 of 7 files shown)


tridge:
we don't really need qsize(), we just need a function that determines if the queue is empty. Can macOS give that somehow?

srmainwaring:
Yes, multiprocessing.Queue.empty() works on macOS(). In the documentation is has the same reliability caveats as qsize().

srmainwaring:
I removed all references to qsize(), so the change is no longer local to macOS, but does bring the platforms in line for Python 3.8+.
The 'spawn` start method is available for macOS from Python 3.4, but is not set as default. It would be possible to force if we wanted to add a call to `multiprocessing.set_start_method('spawn', True)` in the main methods of each process. I thought this would overcomplicate things, so didn't implemented it.

tridge (approved):

tridge:
@srmainwaring nice work!

PR #838, opened 2020-11-24
Feature/sail module

srmainwaring:
This is a dashboard for displaying information relevant to sailing vehicles. The module name is `sail`.

It's an example of a minimal wxPython module that AFAICT works with no errors or warnings on macOS / Python 3.9. I'd wanted a baseline app to start investigations into why some of the other MAVProxy modules are not working on macOS. It borrows heavily from the horizon module in its structure, but does not assign the wxmodule to the mpstate attribute of MPModule. I'm not sure whether this matters or not? (it doesn't appear to).

It's operational but could do with further work to make it production ready - in particular the units are essentially hardcoded rather than driven by the MPSettings which would be better. It also depends on NAMED_VALUE_FLOAT messages for some of the wind data as a dedicated message format for sailing specific info is not yet available in mavlink - I'm not sure what the policy is on this.

It's best viewed using the sailboat sim:

```
sim_vehicle.py -v Rover -f sailboat
``` 

I'd appreciate feedback and any guidance on additional steps necessary to eventually have this PR merged.

wind_display_mockup

### DIFF
--- a/MAVProxy/modules/lib/wxsaildash.py
+++ b/MAVProxy/modules/lib/wxsaildash.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+"""
+  MAVProxy sailing dashboard
+"""
+
+from MAVProxy.modules.lib import multiproc
+import time
+
+class SailingDashboard(object):
+    '''
+    A sailing dashboard for MAVProxy
+    '''
+
+    def __init__(self, title="MAVProxy: Sailing Dashboard"):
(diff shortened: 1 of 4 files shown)


tridge (changes requested):
looks great! just squash the commits (or reword them not as wip)

srmainwaring:
Updated with commits squashed and comments cleaned up.

tridge (approved):

PR #840, opened 2020-11-26
MAVExplorer MagFitUI support for macOS / Windows

srmainwaring:
This change enables the MagFitUI to display on macOS. It is contingent on pickle support for DFReader https://github.com/ArduPilot/pymavlink/pull/480.

An additional function is provided for the UI child process when the platform is macOS which requires DFReader to be pickled. The alternative would be to retain the existing function and provide additional pickle support for the MagFitUI class.

### DIFF
--- a/MAVProxy/modules/lib/magfit.py
+++ b/MAVProxy/modules/lib/magfit.py
@@ -6,6 +6,7 @@ fit best estimate of magnetometer offsets, diagonals, off-diagonals, cmot and sc
 
 from MAVProxy.modules.lib import wx_processguard
 from MAVProxy.modules.lib.wx_loader import wx
+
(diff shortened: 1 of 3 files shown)


srmainwaring:
Rework the approach removing the need to modify pymavlink and applying the change consistently across platforms.

srmainwaring:
- Move custom pickle functions and class wrappers to new file in MAVProxy.modules.lib
- Add new class MagFit in magfit.py to manage multiprocessing and app lifetime
- Replace event loop in MagFitUI.show() with timer and idle events and corresponding callbacks

Tested on:
- macOS / Python 3.9
- Windows 10 / Python 3.9 (VM)

tridge (approved):
great work!

PR #844, opened 2020-12-01
MavExplorer: Stats for macOS / Windows

srmainwaring:
This PR updates enables MAVExplorer > Tools > Stats on macOS / Windows

It also factors out the common code to support child tasks into a small class hierarchy: 

MPChildTask <- MPDataLogChildTask

This reduces much of the duplication in the earlier versions.

Tested on:
- macOS / Python 3.9
- Windows10 / Python 3.9 (VM)


tridge (approved):
looks great!
I've tested py2 and py3 on linux, works well. Graphs are still synchronised properly, as is the map, and the tools work properly

PR #858, opened 2021-01-31
GUI for adding links

stephendade:
This PR adds a GUI for adding links by modifying the link module's menu to include a custom dialog.

Tested in Ubuntu20 and Windows10.

Screenshot from 2021-01-31 15-20-19

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1340,7 +1340,10 @@ if __name__ == '__main__':
           print("Connecting to %s" % serial_list[0])
           mpstate.module('link').link_add(serial_list[0].device)
     elif not opts.master and len(serial_list) > 1:
-          print("Warning: multiple possible serial ports. Use console GUI to add port")
+          print("Warning: multiple possible serial ports. Use console GUI or 'link add' to add port, or restart using --master to select a single port")
+          #if no display, assume running CLI mode and exit
+          if "DISPLAY" not in os.environ:
+              sys.exit(1)
     elif not opts.master:
           wifi_device = '0.0.0.0:14550'
           mpstate.module('link').link_add(wifi_device)
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1340,8 +1340,7 @@ if __name__ == '__main__':
           print("Connecting to %s" % serial_list[0])
           mpstate.module('link').link_add(serial_list[0].device)
     elif not opts.master and len(serial_list) > 1:
-          print("Error: multiple possible serial ports; use --master to select a single port")
-          sys.exit(1)
+          print("Warning: multiple possible serial ports. Use console GUI to add port")
     elif not opts.master:
           wifi_device = '0.0.0.0:14550'
           mpstate.module('link').link_add(wifi_device)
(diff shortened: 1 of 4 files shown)


MAVProxy/mavproxy.py
```diff
@@ -1340,8 +1340,7 @@ def quit_handler(signum = None, frame = None):
           print("Connecting to %s" % serial_list[0])
           mpstate.module('link').link_add(serial_list[0].device)
     elif not opts.master and len(serial_list) > 1:
-          print("Error: multiple possible serial ports; use --master to select a single port")
-          sys.exit(1)
+          print("Warning: multiple possible serial ports. Use console GUI to add port")
```
peterbarker, line 1343: The old advice here is still good.
```suggestion
          print("Warning: multiple possible serial ports. Use console GUI or 'link add'  to add port, or restart using --master to select a single port")
```

MAVProxy/modules/lib/wx_addlink.py
```diff
@@ -0,0 +1,168 @@
...
+    '*FTDI*',
+    "*Arduino_Mega_2560*",
+    "*3D*",
+    "*USB_to_UART*",
+    '*Ardu*',
+    '*PX4*',
+    '*Hex_*',
+    '*Holybro_*',
+    '*mRo*',
+    '*FMU*']
```
peterbarker, line 22: This is going to get out-of-date with the other list.  Perhaps we drop this into a state structure or something?

MAVProxy/modules/mavproxy_link.py
```diff
@@ -208,6 +207,11 @@ def link_add(self, descriptor, force_connected=False):
         '''add new link'''
         try:
             (device, optional_attributes) = self.parse_link_descriptor(descriptor)
+            # if there's only 1 colon in device, it must be serialdevice:baud
+            # As all network connections will be pre-pended with tcp:, udpin:, etc
```
peterbarker, line 211: That's not the case; `localhost:1234` is the same as `udpin:localhost:1234`

peterbarker (changes requested):
This tests out well for me.

The only interface niggle I found is that when creating a serial connection the default baudrate is 38,400 - useful to nobody :-)  (well, that and the "remove link" dropdown is completely busted, but that's not your problem here)

MAVProxy/mavproxy.py
stephendade: Fixed

MAVProxy/modules/lib/wx_addlink.py
stephendade: Removed and now importing preferred_ports from ``MAVProxy/modules/mavproxy_link.py``

MAVProxy/modules/mavproxy_link.py
stephendade: Ahh, did not did consider that. Have fixed up such to check for that case

stephendade:
@peterbarker: I've updated the PR and it should be good to merge now.

peterbarker:
Looks OK to me, but this is a significant behavioural change on startup so @tridge is going to have to be good with it.

peterbarker (approved):

tridge:
if no DISPLAY env variable I think we should fail as before if no devices

tridge (changes requested):
need to check DISPLAY on linux

stephendade:
Have added check for DISPLAY

tridge (approved):

PR #860, opened 2021-02-06
mp_util: fixed param download in py2

tridge:
py2 gzip doesn't have decompress()


peterbarker:
Merged, thanks.

We're probably going to see more of these creep in; its 50/50 whether I'll be running Python2 or Python3 at any particular time.

tridge:
> We're probably going to see more of these creep in; its 50/50 whether I'll be running Python2 or Python3 at any particular time.

it's good to test PRs with both. Very easy to build with each

PR #861, opened 2021-02-06
gimbal: add missing RC option in gimbal mode usage

BTPankow:
The usage string in the `gimbal mode` command was missing the `RC` option even though it's supported. This just corrects that and adds the `RC` option into the usage string.


peterbarker:
Merged, thanks!

Issue #867, opened 2021-02-17
MAVProxy graphs not working for macOS


srmainwaring:
The MAVProxy graph module is not working correctly on macOS.

Steps to replicate:

1. Run SITL:

```bash
$ sim_vehicle.py -v Rover -f rover --map --console
```

2. Load the graph module:

```console
MANUAL> module load graph
MANUAL> Loaded module graph
```

3. Attempt to graph live data in MAVProxy following the [wiki example](https://ardupilot.org/mavproxy/docs/modules/graph.html):

```console
MANUAL> graph VFR_HUD.alt VFR_HUD.airspeed:2
MANUAL> Adding graph: ['VFR_HUD.alt', 'VFR_HUD.airspeed:2']
Process Process-3:
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.1_7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/local/Cellar/python@3.9/3.9.1_7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/live_graph.py", line 56, in child_task
    app.frame = live_graph_ui.GraphFrame(state=self)
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/live_graph_ui.py", line 17, in __init__
    self.create_main_panel()
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/live_graph_ui.py", line 30, in create_main_panel
    from MAVProxy.modules.lib.MacOS import backend_wxagg
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/MacOS/backend_wxagg.py", line 5, in <module>
    from backend_agg import FigureCanvasAgg
ModuleNotFoundError: No module named 'backend_agg'
```

srmainwaring:
Closed by #870

PR #870, opened 2021-03-02
live_graph: fix for #867 on macOS

srmainwaring:
This PR is to fix an error in the graph module on macOS raised in issue #867

The change is macOS specific and affects `lib/MacOS/backend_wxagg.py`. It updates import directives to account for interface changes in the `matplotlib.backends` module.

These changes tested on macOS Catalina 10.15.7, Python 3.9.1, matplotlib 3.3.3.

macos_live_graph_example

My main concern would be support for earlier versions of matplotlib, but we could make it clear in the documentation that a later version is expected and perhaps add matplotlib to `requirements.txt`.


srmainwaring:
Scanning through https://matplotlib.org/stable/api/api_changes_old.html it appears that `DEBUG_MSG` is also deprecated, so remove to prevent graphs breaking when using later versions of matplotlib.

Issue #868, opened 2021-02-17
MAVProxy console settings not working for macOS


srmainwaring:
The The MAVProxy > Settings menu for the console module is not working correctly on macOS.

Steps to replicate:

1. Run SITL:

```bash
$ sim_vehicle.py -v Rover -f rover --map --console
```

2. Select the Console window then select MAVProxy > Settings from the menu bar:

```console
MANUAL> Process Process-5:
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.1_7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/local/Cellar/python@3.9/3.9.1_7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/wxsettings.py", line 32, in child_task
    dlg = SettingsDlg(self.settings)
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/wxsettings_ui.py", line 174, in __init__
    TabbedDialog.__init__(self, tabs, title)
  File "/Volumes/MacPro2_DV1/Code/ardupilot/MAVProxy/MAVProxy/modules/lib/wxsettings_ui.py", line 31, in __init__
    self.dialog_sizer.Add(button_box, 0, wx.GROW|wx.ALIGN_CENTER_VERTICAL|wx.ALL, 5)
wx._core.wxAssertionError: C++ assertion "!(flags & wxALIGN_CENTRE_VERTICAL)" failed at /private/var/folders/s4/4blpy32d4b9_v7_5llt47r7m0000gn/T/pip-install-9ssuf7g3/wxpython/ext/wxWidgets/src/common/sizer.cpp(2077) in DoInsert(): Vertical alignment flags are ignored in vertical sizers
```

srmainwaring:
Removing the wx.ALIGN_CENTER_VERTICAL from the sizers fixes the fault preventing the settings window from opening:

[MAVProxy > Settings] opens the window as expected:

![image](https://user-images.githubusercontent.com/24916364/109688205-eea64500-7b7b-11eb-85ae-488b90e87d25.png)

However there remains a residual warning:

```bash
[...]

PR #871, opened 2021-03-02
wxsettings: fix for #868 on macOS

srmainwaring:
This PR is to fix an error in the console settings module (wxsettings) on macOS raised in issue #868.

The main issue was the use of `wx ALIGN_CENTER_VERTICAL` in vertical sizers which raised an exception in wxPython that prevented the window from displaying.

There are two other minor changes to fix issues that surfaced once the window was working: the first is to replace a deprecated event binding call and the second deals with the Python 2 / 3 treatment of unicode. 

These changes tested on:
- macOS Catalina 10.15.7, Python 3.9.1, wxPython 4.1.0
- Ubuntu 18.04, Python 3.6.9, wxPython 4.1.1
- Ubuntu 18.04, Python 2.7.17, wxPython 4.1.1
- Windows 10, Python 3.9.0, wxPython 4.1.1 (successful when run from PowerShell)
  - (would benefit from further testing on Windows as tests are not successful on all command shells (cmd and cygwin are failing probably due to env set up, MAVProxy run in PowerShell is working) 

image

### DIFF
--- a/MAVProxy/modules/lib/wxsettings_ui.py
+++ b/MAVProxy/modules/lib/wxsettings_ui.py
@@ -1,3 +1,4 @@
+import sys
 from MAVProxy.modules.lib.wx_loader import wx
 
 class TabbedDialog(wx.Dialog):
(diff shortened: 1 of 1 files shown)


jmachuca77:
tested also on Big Sur, python 3.9.1

PR #887, opened 2021-05-07
Vicon: Fix coordinate frame conventions, add object_name option

jackhenderson101:
This commit adds the object_name option to the vicon module. If there are multiple objects being tracked by the vicon system, then you can select the object that corresponds to the UAV. If no option is selected, it will select the first object found.

I have also fixed the coordinate frame conventions, so that the vicon system reports the coordinates in NED, which is what ardupilot expects. There are also some fixes to how quaternions are handled, so that the roll pitch are yaw are correctly reported. I have removed the yaw_offset parameter since it is no-longer necessary.

There is an accompanying pull request on the ardupilot wiki ArduPilot/ardupilot_wiki#3510 to document the correct setup for the vicon system.

### DIFF
--- a/MAVProxy/modules/mavproxy_vicon.py
+++ b/MAVProxy/modules/mavproxy_vicon.py
@@ -10,15 +10,15 @@ from MAVProxy.modules.lib import mp_module
 from MAVProxy.modules.lib import mp_settings
 from MAVProxy.modules.lib import LowPassFilter2p
 from pymavlink.rotmat import Vector3
+from pymavlink.quaternion import Quaternion
 from pymavlink import mavutil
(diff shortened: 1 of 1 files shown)


jackhenderson101:
ArduPilot/ardupilot_wiki#3510

jackhenderson101:
https://cloudstor.aarnet.edu.au/plus/s/ABcR9U1pVnkJ4Q6
Flight log from a quick test with the new code.

tridge (approved):

tridge:
testing the vision msgs as well would be nice, but this is a nice cleanup and logs look good, so merging

PR #905, opened 2021-07-13
Swarming fixes and enhancements

stephendade:
A few patches to make swarming easier:

- Moving the list-of-vehicles from the ``link`` module to ``mpstate``, to make access easier for other modules
- Fixing a bug where ``request_data_stream`` was only sent to 1 vehicle on each link (did not consider the case of multiple vehicles on a single link). Now sends to all vehicles on all links.
- Added a ``swarm`` module, with commands to make swarm management easier

Ping @auturgy


stephendade:
Didn’t want it accidentally merged while it’s still very much a WIP. It’s still in my remote branch.

Sent from my iPhone

On 8 Oct 2021, at 6:12 pm, auturgy ***@***.***> wrote:

﻿

This makes me nervous. Opening a new one?

PR #914, opened 2021-08-11
Add dump_attr command

thanedowning:
The dump_attr command dumps a specific attribute of a given message from the log.

### DIFF
--- a/MAVProxy/tools/MAVExplorer.py
+++ b/MAVProxy/tools/MAVExplorer.py
@@ -477,6 +477,38 @@ def cmd_dump(args):
             continue
         print("%s %s" % (timestring(msg), msg))
     mlog.rewind()
+    
+def cmd_dump_attr(args):
+    '''dump an attribute of a specific message from log'''
+    global xlimits
+    if len(args) < 2:
+        print("Usage: dump_attr MESSAGE attribute")
+        return
+    else:
+        arg_msg = args[0]
+        arg_attribute = args[1]
+    mlog = mestate.mlog
+    mlog.rewind()
+    types = []
+    attribute_types = []
+    for t in mlog.name_to_id.keys():
+        if fnmatch.fnmatch(t, arg_msg):
+            types.extend([t])
+
+    if types == []:
+        print("No message found matching", arg_msg)
+
+    while True:
+        msg = mlog.recv_match(type=types)
+        if msg is None:
+            break
+        in_range = xlimits.timestamp_in_range(msg._timestamp)
+        if in_range < 0:
+            continue
+        if in_range > 0:
+            continue
+        print("%s" % getattr(msg, arg_attribute))
+    mlog.rewind()
 
 mfit_tool = None
 
@@ -963,6 +995,7 @@ command_map = {
     'stats'      : (cmd_stats,     'show statistics on the log'),
     'magfit'     : (cmd_magfit,    'fit mag parameters to WMM'),
     'dump'       : (cmd_dump,      'dump messages from log'),
+    'dump_attr'  : (cmd_dump_attr, 'dump an attribute of a specific message from log'),
     }
 
 def progress_bar(pct):


peterbarker:
@thanedowning you're familiar with the `--verbose` option to `dump`?

thanedowning:
I am not; is there documentation about its usage I could review?  I've tried both  `dump --verbose MESSAGE` and `dump MESSAGE --verbose` with no luck.

peterbarker:
On Wed, 11 Aug 2021, thanedowning wrote:

> I am not; is there documentation about its usage I could review? I've tried both dump --verbose MESSAGE and dump MESSAGE --verbose with no luck.

Probably just as well, I appear to have made it up completely.

I was actually thinking of "status --verbose" which uses a method on the 
message to dump everything we can possibly think of about the message.

It's a more general solution - but would it do what you wanted this PR to 
do?

thanedowning:
Gotcha!  

When I run `status --verbose` from the MAVExplorer console, I get `Unknown command 'status --verbose'`.

There is a `stats` command that's defined as:

`def cmd_stats(args):
    '''show status on log'''

    from MAVProxy.modules.lib import msgstats
    global msgstats_tool
    msgstats_tool = msgstats.MPMsgStats(mlog=mestate.mlog)
    msgstats_tool.start()
`
But when I run `stats --verbose`, I'm greeted by `ERROR in command ['--verbose']: 'mavmmaplog' object has no attribute 'filehandle'`, which seems to be propagating from line 278 of [multiproc_util.py](https://github.com/ArduPilot/MAVProxy/blob/cc641b55fa95069472c8eaf439c7294b3cf5dd16/MAVProxy/modules/lib/multiproc_util.py).

What am I missing here?

I'm happy to defer to a more general solution though.  I only sat down and wrote a new implementation after reading all the documentation and trying all the available commands to no avail.  

I appreciate your input on this and totally understand if you're thinking my approach is a little too niche.  Thanks!

peterbarker:
On Thu, 12 Aug 2021, thanedowning wrote:

> When I run status --verbose from the MAVExplorer console, I get Unknown command 'status --verbose'.

mavproxy, not MAVExplorer :-)

> I appreciate your input on this and totally understand if you're thinking my approach is a little too niche. Thanks!

Probably a good idea to show you what it does here:

***@***.***:~/muck/junkcode-tridge/python/MicMute(master)$ mavlogdump.py --verbose --t HEARTBEAT $TLOG | head -20
2017-10-29 12:44:09.76: HEARTBEAT (id=0) (link=None) (signed=False) 
(seq=0) (src=255/0)
     type: 6 (MAV_TYPE_GCS)
     autopilot: 8 (MAV_AUTOPILOT_INVALID)
     base_mode: 0
       ! MAV_MODE_FLAG_CUSTOM_MODE_ENABLED
       ! MAV_MODE_FLAG_TEST_ENABLED
       ! MAV_MODE_FLAG_AUTO_ENABLED
       ! MAV_MODE_FLAG_GUIDED_ENABLED
       ! MAV_MODE_FLAG_STABILIZE_ENABLED
       ! MAV_MODE_FLAG_HIL_ENABLED
       ! MAV_MODE_FLAG_MANUAL_INPUT_ENABLED
       ! MAV_MODE_FLAG_SAFETY_ARMED
     custom_mode: 0
     system_status: 0 (MAV_STATE_UNINIT)
     mavlink_version: 3





STABILIZE> status --verbose HEARTBEAT
STABILIZE> 2021-08-13 22:58:59.40: HEARTBEAT (id=0) (link=None) 
(signed=False) (seq=42) (src=1/1)
     type: 2 (MAV_TYPE_QUADROTOR)
     autopilot: 3 (MAV_AUTOPILOT_ARDUPILOTMEGA)
     base_mode: 81
         MAV_MODE_FLAG_CUSTOM_MODE_ENABLED
       ! MAV_MODE_FLAG_TEST_ENABLED
       ! MAV_MODE_FLAG_AUTO_ENABLED
       ! MAV_MODE_FLAG_GUIDED_ENABLED
         MAV_MODE_FLAG_STABILIZE_ENABLED
       ! MAV_MODE_FLAG_HIL_ENABLED
         MAV_MODE_FLAG_MANUAL_INPUT_ENABLED
       ! MAV_MODE_FLAG_SAFETY_ARMED
     custom_mode: 0
     system_status: 3 (MAV_STATE_STANDBY)
     mavlink_version: 3

peterbarker:
Huh.  My apologies, I think I've been talking garbage here.  This is for simploifying the output from dump rather than what I was doing - making it more verbose!

Could I suggest dumping an expression instead?

When graphing expressions we grab out messages that we parse out from the expression and emit a data point whenever that message changes.

So instead of `dump_attr ATTITUDE roll` which will give you a crapload of radian values you could `dumpexpr degrees(ATTITUDE.roll)` for example.

And also
`dumpexpr IMU[0].AccZ` (I think) will be more useful than `dump_attr IMU AccZ` as you wouldn't need to first set a selection condition on the messages to select which instance of IMU you want to see data for.

Code to grab messages to pass to `recv_match` is here: https://github.com/ArduPilot/pymavlink/blob/master/tools/mavgraph.py#L241 - and evaluate_expression is used further down.

Could reasonably move into `mavexpression.py` (or just be copied with an apology in the first instance)

PR #917, opened 2021-08-23
Map: remove artefacts from heli icon and update the image.

Gone4Dirt:
This PR removes the image processing artefacts that currently appear on one of the heli images as shown below:
old_heli

I couldn't figure out what had been done to add non-alpha channel regions outside of the clipped image (at least this is what I believe the issue to be due to) so ended up making a new icon.  I decided to go for a more civilian looking looking heli logo.  The below shows the new image relative to multirotor copter and plane for scale comparison:
NewHeliRelative

I created a vector image file for the new heli icon so that we have a scalable image.  Let me know if it is desirable for me to PR the vector file as well so that others have a scale able icon to work with.


peterbarker:
Merged, thanks!  Adding the svg couldn't hurt.

PR #919, opened 2021-08-25
Map: Add heli icon *.svg

Gone4Dirt:
Follow on from #917.  Adding the .svg file for the heli icon.


peterbarker:
Merged, thanks!

PR #921, opened 2021-08-31
Add icons for console, graph and mavexplorer

khancyr:
This add a simple icon.py script that allow to generate icons for mavproxy with the help of Matplotlib.
There is a two steps loading as we cannot use the wx function before the application is created.
It doesn't bring more dependencies.

Rendering on Ubuntu with xorg. (tried on wayland, but icon aren't displayed, another wayland weirdness)
Capture d’écran de 2021-08-31 16-13-59
Capture d’écran de 2021-08-31 16-13-28

### DIFF
--- a/MAVProxy/modules/lib/wxconsole_ui.py
+++ b/MAVProxy/modules/lib/wxconsole_ui.py
@@ -8,9 +8,11 @@ from MAVProxy.modules.lib import win_layout
 class ConsoleFrame(wx.Frame):
     """ The main frame of the console"""
 
-    def __init__(self, state, title):
+    def __init__(self, state, title, ico=None):
         self.state = state
         wx.Frame.__init__(self, None, title=title, size=(800,300))
+        if ico is not None:
+            self.SetIcon(ico.get_ico())
         self.panel = wx.Panel(self)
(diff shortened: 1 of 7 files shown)


MAVProxy/modules/lib/wxconsole_ui.py
```diff
@@ -8,9 +8,11 @@
 class ConsoleFrame(wx.Frame):
     """ The main frame of the console"""
 
-    def __init__(self, state, title):
+    def __init__(self, state, title, ico):
```
peterbarker, line 11:
```suggestion
    def __init__(self, state, title, ico=None):
```

peterbarker (changes requested):
Tested locally.  I like it.

MAVProxy/modules/lib/wxconsole_ui.py
khancyr: done

khancyr:
Capture d’écran de 2021-09-01 00-37-24
update with map icon !

khancyr:
For the Why those colors , I tried to get closed to Mavproxy Windows icons

tridge (approved):

PR #923, opened 2021-09-05
Add units to to FFT axis

Gone4Dirt:
Sadly we don't get units from the log data on the batch sampler so this adds a crude way to label the y axis.

Results look like this:
image

### DIFF
--- a/MAVProxy/modules/lib/mav_fft.py
+++ b/MAVProxy/modules/lib/mav_fft.py
@@ -81,6 +81,14 @@ def mavfft_display(mlog, timestamp_in_range):
             else:
                 return "?Unknown Sensor Type?"
 
+        def units(self):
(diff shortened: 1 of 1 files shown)


peterbarker:
@andyp1per fixed up the units for `mavfft_isb.py` - those units look roughly like these, but we could potentially be consistent.

Gone4Dirt:
Yes, you are right, plus I made a mistake in my units... doh!
just looking at the functionality of the mavfft_isb tool, is there value in me just importing the fft tool from pymavlink into MAVProxy and adding a few ticky boxes for the options?  That way we would get all of the windowing functions as well.  Seems a shame to have duplication.  What do you think?

andyp1per:
Yes that would be my preference - I spent quite a lot of time make mavfft_isb do something sensible :)

PR #924, opened 2021-09-06
Add Linux shortcuts for MAVProxy and MAVExplorer

Hwurzburg:
Tested on Ubuntu 18.04 and 20


tridge:
fails:
image
also, the desktop files should be marked executable in git (chmod +x *.desktop)

Hwurzburg:
so you can't start MAVExplorer.py from any terminal window...in this case the /home directory?

Hwurzburg:
@tridge pushed the improvements we talked about....

tridge (approved):

PR #930, opened 2021-09-23
MAVExplorer: Add custom labels to xml graphs

Gone4Dirt:
This PR adds the custom legend label override functionality to the XML graphs.

Sometimes I prefer a 'prettier' label over the sometimes long expressions in the ledged label.  So this adds the ability to the predfined xml graphs.  There a few examples added as well.  E.g:

image

comes from:

image

I also snuck a few extra comments in there as a seperate commit :-)

### DIFF
--- a/MAVProxy/modules/lib/graphdefinition.py
+++ b/MAVProxy/modules/lib/graphdefinition.py
@@ -4,9 +4,10 @@ GraphDefinition class
 
 class GraphDefinition(object):
     '''a pre-defined graph'''
-    def __init__(self, name, expression, description, expressions, filename):
+    def __init__(self, name, expression, description, expressions, filename, label_or=None):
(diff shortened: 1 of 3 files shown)


tridge:
the existing <SomeLabel> syntax didn't work?

Gone4Dirt:
Correct, adding the label syntax (the one available to us when we overide the label from the command line) to the <expression> in the xml did not work, hence adding this.
I don't recall exactly why that didn't work (was a while ago that I did this), but I do remember testing that first and finding it didn't work.

tridge:
even if existing syntax does work, this is clearer. @Gone4Dirt can you rebase?

Gone4Dirt:
Yes, no problem.  Shall get it rebased over the weekend

PR #932, opened 2021-09-26
Cmdlong: Use target_component instead of MAV_COMP_ID_SYSTEM_CONTROL

stephendade:
Using MAV_COMP_ID_SYSTEM_CONTROL breaks some forms of mavlink routing (like the mavlink-router package). More accurate to use the actual target_component instead.


peterbarker (approved):

peterbarker:
Merged, thanks!

PR #933, opened 2021-09-27
Remove Cesium submodule

khancyr:
The cesium module is old and doesn't work anymore. I think it best to remove it as it is useless for now

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "MAVProxy/modules/mavproxy_cesium"]
-	path = MAVProxy/modules/mavproxy_cesium
-	url = https://github.com/SamuelDudley/MAVCesium.git


auturgy:
I think it does work, just needs documentation (users need their own cesium keys)

tridge:
I'd rather try and fix the issues.

khancyr:
This is a significant effort that nobody want to do for years and in the same time people keep pulling this and waste bandwidth and storage with this.

This is also showing down a little CI

stephendade:
I did have a go a few months ago in trying to update this module.

Due to the Cesium API having changed significantly over the past 3 years (_everything_ the current module uses has been deprecated - most importantly the terrain and tile data formats), this web-facing parts of the module would need to be written almost from the ground-up to work again. It's nowhere near a simple patch or two.

I agree with @khancyr in that I don't see a reason for keeping this module if there's no clear path or volunteer to get it working again.

PR #938, opened 2021-10-02
Graph: Catch exception in Windows

stephendade:
This patch fixes an ``BrokenPipeError`` exception when running under Windows.


stephendade:
Tested on Windows and Linux. Should be good to go in!

tridge (approved):

peterbarker:
This PR has caused a regression on Linux under Python2.

Fix PR is here: https://github.com/ArduPilot/MAVProxy/pull/946

PR #942, opened 2021-10-10
Allow MAVExplorer Tools to be used in Windows

stephendade:
This patch allows the programs in the "Tools" menu of MAVExplorer to be used in Windows.

Previously, MAVExplorer would crash when starting any of these tools (see https://discuss.ardupilot.org/t/magfit-mavexplorer-on-windows/75138).

The root cause of the issue is that Python is unable to pickle a mlog object when passing the mlog to a new process (in Windows. Linux is fine).

I was unable to find a method to make mlog pickle-able, so decided to go with making all Tools use threads instead of processes.

### DIFF
--- a/MAVProxy/modules/lib/wxconsole_ui.py
+++ b/MAVProxy/modules/lib/wxconsole_ui.py
@@ -1,5 +1,6 @@
 import time
 import os
+import socket
 from MAVProxy.modules.lib import mp_menu
 from MAVProxy.modules.lib.wxconsole_util import Value, Text
 from MAVProxy.modules.lib.wx_loader import wx
@@ -93,14 +94,23 @@ class ConsoleFrame(wx.Frame):
             self.timer.Stop()
             self.Destroy()
             return
-        while state.child_pipe_recv.poll():
+
+        while True:
+            try:
+                poll_success = state.child_pipe_recv.poll()
+                if not poll_success:
+                    return
+            except socket.error as e:
+                if e.errno == errno.EPIPE:
+                    break
+                else:
+                    raise e
+
             try:
                 obj = state.child_pipe_recv.recv()
             except EOFError:
-                self.timer.Stop()
-                self.Destroy()
-                return
-            
+                break
+                
             if isinstance(obj, Value):
                 # request to set a status field
                 if not obj.name in self.values:
@@ -142,3 +152,7 @@ class ConsoleFrame(wx.Frame):
                 self.Update()
             elif isinstance(obj, win_layout.WinLayout):
                 win_layout.set_wx_window_layout(self, obj)
+
+        self.timer.Stop()
+        self.Destroy()
+        return


MAVProxy/modules/lib/multiproc_util.py
```diff
@@ -10,6 +10,8 @@
 import mmap
 import platform
 import struct
+import threading
+import os
```
peterbarker, line 14: Put in alphabetical order

MAVProxy/modules/lib/multiproc_util.py
```diff
@@ -198,10 +200,16 @@ def start(self):
         with mutex:
             self.wrap()
             try:
-                self._child = multiproc.Process(target=self._child_task)
+                if os.name == 'nt':
+                    # Windows can't pickle the mavlog object, thus can't spin off
+                    #
```
peterbarker, line 205:
```suggestion
```

MAVProxy/modules/lib/wxconsole_ui.py
```diff
@@ -93,52 +93,52 @@ def on_timer(self, event):
...
+                    if obj is not None:
+                        self.SetMenuBar(None)
+                        self.menu = obj
+                        self.SetMenuBar(self.menu.wx_menu())
+                        self.Bind(wx.EVT_MENU, self.on_menu)
+                    self.Refresh()
+                    self.Update()
+                elif isinstance(obj, win_layout.WinLayout):
+                    win_layout.set_wx_window_layout(self, obj)
+        except (EOFError, BrokenPipeError):
```
peterbarker, line 141:
```suggestion
        except EOFError, BrokenPipeError:
```

MAVProxy/modules/lib/wxconsole_ui.py
```diff
@@ -93,52 +93,52 @@ def on_timer(self, event):
             self.timer.Stop()
             self.Destroy()
             return
-        while state.child_pipe_recv.poll():
-            try:
+        try:
```
peterbarker, line 96: Why are you shifting the `try` here?

It increases the code covered by the try/except block - when the exception being caught should really only come from the line that's currently covered by the try/except block, shouldn't it?

peterbarker (changes requested):

MAVProxy/modules/lib/wxconsole_ui.py
stephendade: Similar to https://github.com/ArduPilot/MAVProxy/commit/4898e10555f22a0db5a3e36be7243b568ca6c732, the ``BrokenPipeException`` occurs when ``poll()`` is called. Not sure if there's a cleaner way to do this...

stephendade:
Updated as per you comments

MAVProxy/modules/lib/wxconsole_ui.py
```diff
@@ -93,52 +93,52 @@ def on_timer(self, event):
...
+                    if obj is not None:
+                        self.SetMenuBar(None)
+                        self.menu = obj
+                        self.SetMenuBar(self.menu.wx_menu())
+                        self.Bind(wx.EVT_MENU, self.on_menu)
+                    self.Refresh()
+                    self.Update()
+                elif isinstance(obj, win_layout.WinLayout):
+                    win_layout.set_wx_window_layout(self, obj)
+        except EOFError, BrokenPipeError:
```
peterbarker, line 141: See https://github.com/ArduPilot/MAVProxy/pull/946 - `BrokenPipeError` doesn't exist under Python2

peterbarker (changes requested):

MAVProxy/modules/lib/wxconsole_ui.py
stephendade: Cool - i've incorporated that change now

PR #944, opened 2021-10-11
Create --udp-timeout option

Williangalvani:
This allows connecting multiple clients to an udp server.
If there is more than one client connected, clients with no communication after `udp-timeout` seconds get removed.

Depends on https://github.com/ArduPilot/pymavlink/pull/574

With timeout=0 (the default), the communication is kept even if the GCS is silent (`set heartbeat 0`)
Screenshot from 2021-10-11 19-30-08

With timeout higher than zero, the clients that stop talking get removed:
Screenshot from 2021-10-11 19-36-09

I'm guessing the "Heartbeat OK" is some background task of MAVProxy restoring the communication by sending something (apparently mavproxy sending REQUEST_DATA_STREAM)


Williangalvani:
should this wait for a new release of pymavlink and update requirements.txt?

tridge:
> should requirements.txt be updated to require pymavlink 2.4.18?

i tried this recently, and it broke CI testing of ArduPilot on older branches, like 4.2 releases.
I think we'll need to only add the udp_timeout option if it is non-zero, sorry

tridge (approved):

PR #945, opened 2021-10-14
remove menus when unloaded

jinchengde:
remove the console menu when the module unload, that could help if module console reload and add the menu again


peterbarker:
Merged, thanks!

PR #947, opened 2021-10-17
MAVProxy: Icons

auturgy:
I had these done a while ago and completely forgot to push them.
Likely overtaken by events, but provided in case they're useful - I think they're quite good.


nkruzan:
They look good to me

tridge:
merging these in, then we need to work out where to use them

PR #950, opened 2021-10-20
Asterix: Python 3 compatibility

stephendade:
It's been a while since we last used this module, fixed a few Python3 issues.


stephendade:
> Does this still work under Python2?

Yes, I've checked.

peterbarker:
Merged, thanks!

PR #951, opened 2021-10-20
Add missing errno imports

peterbarker:


peterbarker:
I've tested this on Linux - the import works fine.

I don't currently have a Windows install of MAVProxy which is why my testing on the first PR didn't pick this up.

peterbarker:
On Wed, 20 Oct 2021, Henry Wurzburg wrote:

> any way I can test this for you? replace current file with these? but I dont
> know where in Windows they reside...

Me neither! :-)

I will get MAVProxy onto my Windows box in the next couple of days in any 
case.

PR #952, opened 2021-10-22
ADSB: Multi-vehicle compatibility and revamp

stephendade:
Modernisation of the ADSB Module:

- Able to take in ADSB feeds from all connected vehicles
- Will print on console if any vehicles are too close (defined by threat_radius) to ADSB objects
- "In range" detection changed to being the same as ArduPilot (separate horizontal and vertical thresholds)


stephendade:
I've fixed it up - now using a similar approach to the AIS module.

peterbarker (changes requested):
This looks OK but I'll want to test it a chunk locally.

peterbarker:
Testing went well, but I have (re)discovered that you really don't want multiple vehicles that are able to see each other's `ADSB_VEHICLE` messages.

stephendade:
I've fixed up the variable naming and re-added threat_radius. If threat_radius is 0, then the xy/z overrides will be used instead.

PR #956, opened 2021-10-29
mavproxy_console:unload console module after close window

jinchengde:
unload console module after close window

### DIFF
--- a/MAVProxy/modules/mavproxy_console.py
+++ b/MAVProxy/modules/mavproxy_console.py
@@ -36,6 +36,8 @@ class ConsoleModule(mp_module.MPModule):
         self.last_sys_status_errors_announce = 0
         self.user_added = {}
         self.safety_on = False
+        self.unload_check_interval = 5 # seconds
+        self.last_unload_check_time = time.time()
         self.add_command('console', self.cmd_console, "console module", ['add','list','remove'])
         mpstate.console = wxconsole.MessageConsole(title='Console')
 
@@ -652,6 +654,13 @@ class ConsoleModule(mp_module.MPModule):
                 except Exception as ex:
                     print(ex)
                     pass
+
+    def idle_task(self):
+        now = time.time()
+        if self.last_unload_check_time + self.unload_check_interval < now:
+            self.last_unload_check_time = now
+            if not self.console.is_alive():
+                self.needs_unloading = True
 
 def init(mpstate):
     '''initialise module'''


MAVProxy/modules/mavproxy_console.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, mpstate):
         self.last_sys_status_errors_announce = 0
         self.user_added = {}
         self.safety_on = False
+        self.unload_check_interval = 0.1 # seconds
```
tridge, line 39: on slow systems (eg. windows python) I suspect this will trigger while loading

tridge (changes requested):

MAVProxy/modules/mavproxy_console.py
jinchengde: how about 0.5 seconds?

jinchengde: and this parameter also used in mavproxy_map/mission_editor/param_editor

jinchengde:
I will have another PR for other unload_check_interval parameter

MAVProxy/modules/mavproxy_console.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, mpstate):
         self.last_sys_status_errors_announce = 0
         self.user_added = {}
         self.safety_on = False
+        self.unload_check_interval = 0.5 # seconds
```
tridge, line 39: please add to 5s. On some systems the thread creation is very slow. With this 0.5s change it will close the window before it opens

tridge (changes requested):

MAVProxy/modules/mavproxy_console.py
jinchengde: done

jinchengde: @tridge done

MAVProxy/modules/mavproxy_console.py
```diff
@@ -653,6 +655,13 @@ def mavlink_packet(self, msg):
                     print(ex)
                     pass
 
+    def idle_task(self):
+        now = time.time()
+        if self.last_unload_check_time + self.unload_check_interval < now:
```
peterbarker, line 660: It's usually better to subtract times as a habit rather than adding them.  Doesn't really matter here, but sometimes it does.

MAVProxy/modules/mavproxy_console.py
peterbarker: @tridge This is an *unload check interval", not the same timeout as we have on some module's load time!

I think I might merge this one anyway - but it really ought to be 0.5s.

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #958, opened 2021-11-01
Modules: Do not list ADSB system as a vehicle

stephendade:
In flight controllers with mavlink-based ADSB receivers (such as the ADSB Cube carrier board), the receiver will show up as a separate vehicle in MAVProxy (so 2 separate vehicles in the MAVProxy console's "Vehicle" menu), leading to some confusion as to which is the correct vehicle.

This is extra confusing if the flight controller has a different SYSID to the ADSB receiver.

### DIFF
--- a/MAVProxy/modules/mavproxy_console.py
+++ b/MAVProxy/modules/mavproxy_console.py
@@ -199,6 +199,8 @@ class ConsoleModule(mp_module.MPModule):
             return "Tracker"
         if hb.type == mavutil.mavlink.MAV_TYPE_AIRSHIP:
             return "Blimp"
+        elif hb.type == mavutil.mavlink.MAV_TYPE_ADSB:
+            return "ADSB"
         return "UNKNOWN(%u)" % hb.type
 
     def component_type_string(self, hb):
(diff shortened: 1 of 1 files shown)


peterbarker:
How would a user set a parameter on an ADSB device if we merged this?

Perhaps we need a setting to include everything, including things that aren't vehicles?

stephendade:
Ok, I've re-done the PR to just list the vehicle as "ADSB". Should make things a bit clearer for users.

>How would a user set a parameter on an ADSB device if we merged this?

I've not been able to grab any parameters or otherwise get anything out of the ADSB device except the heartbeat and ADSB messages. I was under the impression these devices were not configurable. Is that not the case?

peterbarker:
On Sat, 6 Nov 2021, Stephen Dade wrote:

> Ok, I've re-done the PR to just list the vehicle as "ADSB". Should make things a bit clearer for users.
>
>       How would a user set a parameter on an ADSB device if we merged this?
> 
> I've not been able to grab any parameters or otherwise get anything out of the ADSB device except the heartbeat and ADSB messages. I was under the impression these devices were not configurable. Is that not the case?

Some of them are.

peterbarker:
I've fixed up the commit message a bit.

Note that ordinarily you'd set bit 10 in SERIALn_OPTIONS (`SERIAL5_OPTIONS` on a CubeOrange carrier board) to avoid this device being seen outside the autopilot itself.

Tested, merging, thanks!

PR #966, opened 2021-11-13
module:graph - Plot time interval between messages - WIP

crashmatt:
Provides graph debugging of intervals between messages.  For telemetry links with limited bandwidth or robustness.
Uses its own timestamping mechanism.  Will only work on live data or playback at normal speed.

Uses keyword ".interval" to add interval between messages to message data and then plot it.
Will collide with any mavlink message with a variable named "interval".  Did not find any messages configured this way.

### DIFF
--- a/MAVProxy/modules/mavproxy_graph.py
+++ b/MAVProxy/modules/mavproxy_graph.py
@@ -6,6 +6,7 @@
 
 from pymavlink import mavutil
 import re, os, sys
+import time
 
 from MAVProxy.modules.lib import live_graph
 
@@ -35,7 +36,7 @@ class GraphModule(mp_module.MPModule):
             return
 
         elif args[0] == "help":
-            print("graph <timespan|tickresolution|expression>")
+            print("graph <timespan|tickresolution|expression|frequency>")
         elif args[0] == "timespan":
             if len(args) == 1:
                 print("timespan: %.1f" % self.timespan)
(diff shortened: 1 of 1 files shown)


peterbarker:
On Sat, 13 Nov 2021, Matthew Coleman wrote:

> Provides graph debugging of intervals between messages. For telemetry links with limited bandwidth or robustness.
> Uses its own timestamping mechanism. Will only work on live data or playback at normal speed.
> 
> Uses keyword ".interval" to add interval between messages to message data and then plot it.
> Will collide with any mavlink message with a variable named "interval". Did not find any messages configured this way.

Usually we plot "diff(ATT._timestamp,0)" for this.  Does this add anything 
over that?

crashmatt:
Maybe not.  I can't find anywhere in mavproxy that explicitly sets timestamps for incoming messages other than in data logging.

crashmatt:
and I did not know about diff() either.

Need to do a bit of testing to prove to myself if this works.

I can see that the magic happens at evaluateexpression.  How to use this is not obvious without better examples.  Will look to add some examples to the [graph module documentation](https://ardupilot.org/mavproxy/docs/modules/graph.html)

Found msg._timestamp in mavutil.post_message which is used in mavlink [here](https://github.com/ArduPilot/MAVProxy/blob/917f68803b1abf0587c824bb45b13e8609de4c6d/MAVProxy/mavproxy.py#L802)

Which still doesn't make it clear if timestamps are always added to messages before sending to modules.
I didn't find how messages are sent to modules.  If they are treated as mavlink masters then I missed how modules get added to the list of masters.

Again, I need to test.

crashmatt:
Found mavlink_message for modules in [module_link](https://github.com/ArduPilot/MAVProxy/blob/0479f47c97dc1dcaf621ce75fe1ba41e79c7b234/MAVProxy/modules/mavproxy_link.py#L870)

and now I think I see that _timestamp is always added.

Now the next thing I need is message size

crashmatt:
Confirmed. Closing PR.

crashmatt:
@peterbarker  New problem.  The diff(ATT._timestamp,0) method only works for one value.  If you do more than one diff then they interfere with each other.  This happens within single or multiple graphs.
My aim is to have multiple message intervals shown on one plot so diff() will not work.

peterbarker:
On Mon, 15 Nov 2021, Matthew Coleman wrote:

> @peterbarker New problem. The diff(ATT._timestamp,0) method only works for one value. If you do more than one diff then they interfere with each other. This happens within single or multiple graphs.
> My aim is to have multiple message intervals shown on one plot so diff() will not work.

That second field is a key into a hash.  So if you want a second diff 
you'd diff(POS._timestamp,1)

crashmatt:
Another case for updating the documentation.  Closing PR.

PR #967, opened 2021-11-14
module:live_graph_ui - YAxis log scale option/button

crashmatt:
Allows user to show higher dynamic range of data on the same plot.

**Linear mode**
image

**Log mode with range change**
image


crashmatt:
To be combined with standalone module.  Closing PR.

PR #968, opened 2021-11-15
module:graph - plot all messages over real time - WIP

crashmatt:
Plots all message values exactly on time axis instead of latest message values sampled at fixed interval.
Allows the user to see the relative time separation of all the received messages

Implemented with child classes of originals which needs improvement.  Either as a completely separate entity or different  organization of the original.

Example with two messages sent at two different rates.
image


crashmatt:
Making a standalone module.  Closing PR.

PR #969, opened 2021-11-16
extend unload_check_interval for slow system

jinchengde:
reference https://github.com/ArduPilot/MAVProxy/pull/956


peterbarker:
This is an interval at which we check things.  AFAICS this will just delay the time it takes us to unload a module.

Is there some sort of race condition this makes less likely or something?

jinchengde:
@peterbarker @tridge 
I also think 5s is a little long for check status, 0.5s ~ 1s is better

PR #975, opened 2021-11-29
mavlfightview: fixed faulthandler have no register in windows

jinchengde:
faulthandler have no register attribute in windows, fixed that startup issue

### DIFF
--- a/MAVProxy/tools/mavflightview.py
+++ b/MAVProxy/tools/mavflightview.py
@@ -664,7 +664,10 @@ if __name__ == "__main__":
 
     try:
         import faulthandler, signal
-        faulthandler.register(signal.SIGUSR1)
+        try:
+            faulthandler.register(signal.SIGUSR1)
+        except AttributeError as e:
+            pass
     except ImportError:
         pass
 


MAVProxy/tools/mavflightview.py
```diff
@@ -665,7 +665,7 @@ def __init__(self):
     try:
         import faulthandler, signal
         faulthandler.register(signal.SIGUSR1)
-    except ImportError:
+    except Exception as e:
```
peterbarker, line 668: This is way too generic an error to be catching.

What's the error you get in the original code?

peterbarker: ... I'll also note that we use `ImportError` in many other places so if it is a problem it's a big problem :-)

jinchengde: the problem is that is you haven't install faulthandler in system when you import that package, you will have importError, that's right, but if you had install faulthandler but this module doesn't support register function in windows, they will have another error.
pls reference: https://docs.python.org/3/library/faulthandler.html

peterbarker: What is that other error?  My guess is another exception type - we can add it to the list of Exceptions caught - rather than just catching the base class exception and heedlessly going on.

jinchengde: fixed

peterbarker:
Merged, thanks @jinchengde 

We should probably have a comment explain *why* - but meh - someone will find my comment here in the future.

PR #983, opened 2021-12-19
default to verbose param display

tridge:


MAVProxy/modules/lib/param_help.py
```diff
@@ -31,26 +31,30 @@ def param_help_download(self):
...
+                    print("Unknown vehicle type")
                 return None
             path = mp_util.dot_mavproxy("%s.xml" % self.vehicle_name)
             if not os.path.exists(path):
                 if self.vehicle_name == 'APMrover2':
                     path = mp_util.dot_mavproxy("%s.xml" % "Rover")
             if not os.path.exists(path):
-                print("Please run 'param download' first (vehicle_name=%s)" % self.vehicle_name)
+                if verbose:
+                    print("Please run 'param download' first (vehicle_name=%s)" % self.vehicle_name)
```
peterbarker, line 53: This is also essentially an error and should be emitted all the time - verbose or not

MAVProxy/modules/lib/param_help.py
```diff
@@ -31,26 +31,30 @@ def param_help_download(self):
...
                 return None
         if not os.path.exists(path):
-            print("Param XML (%s) does not exist" % path)
+            if verbose:
+                print("Param XML (%s) does not exist" % path)
```
peterbarker, line 57: Also an error and should be displayed verbose-or-not

peterbarker (commented):
Pretty sure we want to display out-and-out errors even if we're not verbose.

peterbarker:
Nudge.  Want this feature, but also want the diagnostics even if we're not verbose.

MAVProxy/modules/lib/param_help.py
```diff
@@ -31,26 +31,30 @@ def param_help_download(self):
...
         if self.last_pair == (self.xml_filepath, self.vehicle_name):
             return self.last_htree
         if self.xml_filepath is not None:
-            print("param: using xml_filepath=%s" % self.xml_filepath)
+            if verbose:
+                print("param: using xml_filepath=%s" % self.xml_filepath)
             path = self.xml_filepath
         else:
             if self.vehicle_name is None:
-                print("Unknown vehicle type")
```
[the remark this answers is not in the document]
tridge: nope, this does happen and would be very annoying in set and show

MAVProxy/modules/lib/param_help.py
tridge: nope, it would be extremely annoying.

MAVProxy/modules/lib/param_help.py
tridge: nope

PR #984, opened 2021-12-21
Add special handling for SiK RADIO_STATUS info for graph module

pap9ck:
The graph module cannot display the RADIO_STATUS items for a SiK radio (together with other status items for the primary target vehicle). The reason for this is that the SiK radio firmware injects RADIO_STATUS mavlink packets having a hard coded sysid/componentid set to '3'/'D' (see mavlink.c macro definition in ardupilot/SiK repository). The graph module is multi_vehicle=false, so it will only get the data for the current primary vehicle and thus the RADIO_STATUS messages with sysid/"vehicle" id 51 don't reach the graph module. This legacy situation warrants special exception handling for the RADIO_STATUS frames.

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -862,10 +862,17 @@ class LinkModule(mp_module.MPModule):
             for (mod,pm) in self.mpstate.modules:
                 if not hasattr(mod, 'mavlink_packet'):
                     continue
-                if not mod.multi_vehicle and sysid != target_sysid:
-                    # only pass packets not from our target to modules that
-                    # have marked themselves as being multi-vehicle capable
-                    continue
+                # sysid 51/'3' is used by SiK radio for the injected RADIO_STATUS mavlink frames.
+                # In order to be able to pass these to the graph module, which is not multi-vehicle,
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_link.py
```diff
@@ -862,10 +862,17 @@ def master_callback(self, m, master):
...
-                    # only pass packets not from our target to modules that
-                    # have marked themselves as being multi-vehicle capable
-                    continue
+                # sysid 51/'3' is used by SiK radio for the injected RADIO_STATUS mavlink frames.
+                # In order to be able to pass these to the graph module, which is not multi-vehicle,
+                # special handling is needed to have graph display both RADIO_STATUS and target 
+                # vehicle information.
+                # Since only graph module is tested getting the additional sysid frames, condition
+                # for this is added to the exception as well.
+                if not(sysid == 51 and mod.name == "graph"):
```
tridge, line 871: I think we should make the exception based on mtype=='RADIO_STATUS' (perhaps include 'RADIO' as well, the older form?) rather than the mod.name. I think piping RADIO_STATUS to all modules will be ok, and could be useful for some modules that want to do RSSI monitoring

tridge (changes requested):

tridge (approved):

Issue #987, opened 2021-12-21
MAVExplorer: Deprecated matplotlib calls


WickedShell:
I've forgotten the invocation to trigger this, but the stack trace lets you look at it:

```
/home/wickedshell/.local/bin/MAVExplorer.py:1024: MatplotlibDeprecationWarning: 
The num2epoch function was deprecated in Matplotlib 3.5 and will be removed two minor releases later. Use mdates.num2date(e).timestamp() instead.
  xlimits.xlim_high = matplotlib.dates.num2epoch(xlim[1]) - tzofs
```

``` shell
[wickedshell@lackey ~]$ pip show matplotlib
Name: matplotlib
Version: 3.5.1
Summary: Python plotting package
Home-page: https://matplotlib.org
Author: John D. Hunter, Michael Droettboom
Author-email: matplotlib-users@python.org
License: PSF
Location: /home/wickedshell/.local/lib/python3.10/site-packages
Requires: kiwisolver, fonttools, pyparsing, python-dateutil, numpy, packaging, cycler, pillow
Required-by: 
[wickedshell@lackey ~]$ python --version
Python 3.10.1
```

hendjoshsr71:
Looks like this can be closed.

PR #991, opened 2021-12-28
MAVExplorer: fix deprecated-function warning from matplotlib

peterbarker:


peterbarker:
Closes #987 

@WickedShell this doesn't kill older versions AFAICT.  Can you test it fixes your problem, please?

WickedShell (commented):
It looks to me like `num2date` existed as far back as at least [matplotlib 1.3.1 ](https://matplotlib.org/1.3.0/api/dates_api.html#matplotlib.dates.num2date) which would have been 2013. Do we still need the fallback path?

peterbarker:
On Tue, 28 Dec 2021, WickedShell wrote:

> @WickedShell commented on this pull request.
> 
> It looks to me like num2date existed as far back as at least matplotlib 1.3.1 which would have been 2013. Do we still need the fallback path?

Probably not.  But this is the least likely to break any existing setups.

WickedShell (approved):
Fixes the issue, I still think we should drop the backwards compat path, as it shouldn't be an issue for anyone.

PR #994, opened 2022-01-03
sitl_autospeedup: module to help with SITL testing

peterbarker:
Twiddles SIM_SPEEDUP parameter at simulation startup


peterbarker:
image

peterbarker:
I think notionally we could build this into SITL instead, with a `--prep-for-arming` command-line option.  We would attach to (e.g. port 5760) ourselves and do what the MAVProxy script currently does, then drop the connection and start MAVProxy on the same connection.  Drawback is that you don't get MAVProxy at all until the vehicle is ready to go.

peterbarker:
Didn't ever get around to making this 100%

PR #998, opened 2022-01-17
Corrected major/minor/patch decoding of flight_sw_version in AUTOPILOT_STATUS

pap9ck:
The useralerts module retrieves the major/minor/patch version (to compare with the alerts) from the AUTOPILOT_STATUS message's flight_sw_version field. But the decoding of this 32bit integer is incorrectly done with the python hex function which array index iterates through octets not bytes of the flight_sw_version field whereas the encoding in the ardupilot is using encoding as below /1/. The decoding function has been separated out as this might be useful in other places, e.g. the version command in mavproxy doesn't produce any output today (the triggered AUTOPILOT_STATUS message just goes into the log), but that will be for another PR.

/1/
    flight_sw_version = version.major << (8 * 3) | \
                        version.minor << (8 * 2) | \
                        version.patch << (8 * 1) | \
                        (uint32_t)(version.fw_type) << (8 * 0);

### DIFF
--- a/MAVProxy/modules/lib/mp_util.py
+++ b/MAVProxy/modules/lib/mp_util.py
@@ -490,3 +490,24 @@ def decode_devid(devid, pname):
         bustypes.get(bus_type,"UNKNOWN"), bus_type,
         bus, address, address, devtype, devtype, decoded_devname,
         devid))
+
+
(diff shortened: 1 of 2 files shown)


stephendade (approved):
Thanks! Will merge

PR #999, opened 2022-01-25
Add FeiYing Model KMODEL Simulator joystick

Jaaaky:
A famous one used by hobbyists

I've tested it myself for a friend and confirmed working successfully


peterbarker (approved):
LGTM

peterbarker:
Merged, thanks!

I wouldn't be surprised if we do need to narrow that pattern into the future, 'though :-)

PR #1015, opened 2022-02-19
catch an exception in mavexplorer on windows

tridge:


peterbarker (approved):
Would be nice to narrow that exception.

PR #1016, opened 2022-02-24
map: allow for override of vehicle type

tridge:
useful in swarms to make vehicles distinct


peterbarker (approved):
Looks like it will do the job.

Better would be a `map vehiclecolour SYSID (r,g,b,a)` - a lot more work, but should be doable if we assume you can take a red-copter and change all red pixels to the new colour.

PR #1017, opened 2022-02-25
allow wildcards in --master

tridge:
useful to connect to many links at once
tested with 5 boards like this:
  mavproxy.py --console --master /dev/serial/by-id/usb\\*0

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -17,6 +17,7 @@ import math
 import platform
 import json
 import struct
+import glob
 
 try:
     reload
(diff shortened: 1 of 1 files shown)


MAVProxy/mavproxy.py
```diff
@@ -1360,7 +1361,11 @@ def quit_handler(signum = None, frame = None):
 
     # open master link
     for mdev in opts.master:
-        if not mpstate.module('link').link_add(mdev, force_connected=opts.force_connected):
+        if mdev.find('?') != -1 or mdev.find('*') != -1:
```
peterbarker, line 1364: I don't get it - why not just always glob?

If those characters don't exist then the result will be the same.

tridge: > I don't get it - why not just always glob?

it would break platforms like windows

PR #1022, opened 2022-03-15
MAVProxy: default to using the 'all' dialect

peterbarker:
This would allow MAVProxy to see messages in other inclusions in `all.xml` - so including the ones in `storm32.xml` for example.

I've run this for a while locally and it seems to run just fine.


peterbarker:
pymavlink equivalent is here: https://github.com/ArduPilot/pymavlink/pull/650

PR #1025, opened 2022-03-17
try to handle windows exceptions

tridge:


peterbarker (approved):
Born from frustration...

PR #1033, opened 2022-03-27
ntrip: make connection non-blocking

tridge:
this prevents blocking main thread when there is no internet link

### DIFF
--- a/MAVProxy/modules/lib/ntrip.py
+++ b/MAVProxy/modules/lib/ntrip.py
@@ -47,6 +47,7 @@ class NtripClient(object):
         self.user = base64.b64encode(user)
         self.port = port
         self.caster = caster
+        self.caster_ip = None
         self.mountpoint = mountpoint
         if not self.mountpoint.startswith("/"):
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/lib/ntrip.py
```diff
@@ -121,7 +123,8 @@ def get_ID(self):
 
     def read(self):
         if self.socket is None:
-            time.sleep(0.1)
+            if self.socket_pending is None:
+                time.sleep(0.1)
```
peterbarker, line 127: This structure will introduce a 100ms sleep every other call.

Scribbling down a `last_connect_attempt = time.time()` and using it to determine when to try again would avoid this.

peterbarker (commented):
Looks good except for the 100ms pause in the main loop.

... looks like an infrastructure problem with CI.

MAVProxy/modules/lib/ntrip.py
tridge: fixed

PR #1034, opened 2022-03-30
added mqtt module

Towflos:
added a module to publish mavlink messages to mqtt

### DIFF
--- a/MAVProxy/modules/mavproxy_mqtt.py
+++ b/MAVProxy/modules/mavproxy_mqtt.py
@@ -1,0 +1,70 @@
+from paho.mqtt import MQTTException
+from MAVProxy.modules.lib import mp_settings
+from MAVProxy.modules.lib import mp_module
+import paho.mqtt.client as mqtt
+import json
+import numbers
+
+
+class MqttModule(mp_module.MPModule):
+
+    def __init__(self, mpstate):
+        super(MqttModule, self).__init__(mpstate, "mqtt", "mqtt publisher")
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_mqtt.py
```diff
@@ -0,0 +1,70 @@
...
+        elif args[0] == 'connect':
+            self.connect()
+
+    def usage(self):
+        """show help on command line options"""
+        return "Usage: mqtt <set|connect>"
+
+    def convert_to_dict(self, message):
+        """converts mavlink message to python dict"""
+        if hasattr(message, '_fieldnames'):
```
peterbarker, line 58: when is this not the case?

peterbarker (approved):
This looks interesting.

I'm going to merge this - it's not going to break anyone who doesn't load it.

peterbarker:
Merged, thanks!

MAVProxy/modules/mavproxy_mqtt.py
Towflos: I was not sure if there could be messages that contain another message as field that's why the function is recursive. This checks if the function gets called on a field with a message or a value.

peterbarker:
On Thu, 31 Mar 2022, Towflos wrote:

> I was not sure if there could be messages that contain another message as field that's why the function is recursive. This checks if the function gets called on a field with a message or a value.

Yeah, that doesn't happen :-)

Feel free to simplify.

Issue #1069, opened 2022-06-09
WP Editor corrupts command's Param5 if not Latitude


rmackay9:
Some mission commands attempt to hold values in Param5 but they become corrupted after saving probably because the editor is converting between a human readible lattitude (e.g. -35.36...) and what must be sent (e.g. -35.36 * 1e7)

After "Write WP" (e.g. what the user entered)
![image](https://user-images.githubusercontent.com/1498098/172749391-bca40ab3-89a9-4af1-96dc-c1536c6cf203.png)

After "Read WP" (e.g. what the user gets)
![image](https://user-images.githubusercontent.com/1498098/172749438-b64d1cd3-7ae4-408c-8cc4-14b5bdb405e2.png)

[The same issue exists for Mission Planner](https://github.com/ArduPilot/MissionPlanner/issues/2871)

I do not believe this is a flight code issue because I added debug to AP_Mission and the mavlink message's packet.x seems to have a very large number it in.

I think the best way to fix this is to check the xml's "isDestination" tag.  It is false for the commands that have this issue and true for all others.

tridge:
fixed by #1110

PR #1076, opened 2022-06-24
mavproxy_wp: check has_location attribute before multiplying x and y

peterbarker:
Closes https://github.com/ArduPilot/MAVProxy/issues/1069

Requires https://github.com/ArduPilot/pymavlink/pull/697


rmackay9:
Thanks very much for this!  I'm not sure I can personally test this because I don't normally run the dev version of MAVProxy but the issue linked shows the method to replicate the problem (and you've probably already tested).

PR #1077, opened 2022-06-27
param: support downloading param defaults from vehicle

tridge:
to go with https://github.com/ArduPilot/ardupilot/pull/20972


tridge:
replaced by #1082

PR #1081, opened 2022-07-18
movinghome: Accept position information from other constellations too

booo:
Position information can come from other constellations too as discribed
in another ArduPilot related ticket:

https://github.com/ArduPilot/ardupilot/issues/1241

I did not test this with actual hardware. I can test this in the next days if necessary.


booo:
As stated in the referenced ticket the first to characters of refer to the constellation in use. The GGA suffix is same for all constellations and the content should be too because it is a GGA message. But I did not test this so far.

booo:
I think we should rewrite the moving home example to use the `mavproxy_nmeagps.py` module.

PR #1085, opened 2022-07-31
added a wx ID server

tridge:
this ensures we have unique IDs across multiple processes

darn, doesn't work on windows


tridge:
doesn't work on windows, closing

PR #1088, opened 2022-08-01
map: mp_elevation: take and pass cachedir through in ElevationModel init

peterbarker:
This allows a user to point to special places they may be storing their data

### DIFF
--- a/MAVProxy/modules/mavproxy_map/mp_elevation.py
+++ b/MAVProxy/modules/mavproxy_map/mp_elevation.py
@@ -23,7 +23,7 @@ TERRAIN_SERVICES = {
 class ElevationModel():
     '''Elevation Model. Only SRTM for now'''
 
-    def __init__(self, database='SRTM3', cachedir=None, offline=0, debug=False):
+    def __init__(self, database='SRTM3', offline=0, debug=False, cachedir=None):
         '''Use offline=1 to disable any downloading of tiles, regardless of whether the
         tile exists'''
         if database.lower() == 'srtm':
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_map/mp_elevation.py
```diff
@@ -23,15 +23,15 @@
 class ElevationModel():
     '''Elevation Model. Only SRTM for now'''
 
-    def __init__(self, database='SRTM3', offline=0, debug=False):
+    def __init__(self, database='SRTM3', cachedir=None, offline=0, debug=False):
```
tridge, line 26: please add the new parameter at the end so existing users that do EvelationModel("SRTM", True) still work

tridge (changes requested):

MAVProxy/modules/mavproxy_map/mp_elevation.py
```diff
@@ -23,15 +23,15 @@
 class ElevationModel():
     '''Elevation Model. Only SRTM for now'''
 
-    def __init__(self, database='SRTM3', offline=0, debug=False):
+    def __init__(self, offline=0, database='SRTM3', cachedir=None, debug=False):
```
tridge, line 26: still changes order

tridge (approved):

peterbarker:
On Mon, 1 Aug 2022, Andrew Tridgell wrote:

> -    def __init__(self, database='SRTM3', offline=0, debug=False):
> +    def __init__(self, database='SRTM3', cachedir=None, offline=0, debug=False):
> 
> please add the new parameter at the end so existing users that do EvelationModel("SRTM", True) still work

That wouldn't work - "database" is the first positional argument there.

I've moved "offline" to be the first positional argument, which I think is 
what you wanted.

PR #1092, opened 2022-08-08
OpenDroneID: added support module

tridge:


davidbuzz:
How does this module get the operator's ground position?  maybe a usb-based GPS module ?

PR #1099, opened 2022-08-16
mavproxy_misc: add dfu_boot command

bugobliterator:


peterbarker:
On the ArduPIlot side - I tried to do the reboot-to-DFU-mode thing, but the images that were then flashed ended up corrupt  Hope you have better luck!

https://github.com/ArduPilot/ardupilot/pull/13633

PR #1100, opened 2022-08-17
map: remove reference to OviHybrid tile service

lopsided98:
Change the default value of the `--service` argument from OviHybrid to MicrosoftSat. OviHybrid was removed a few years ago, so attempting to run `mp_tile.py` as a script with no arguments results in the following error:

```
  File "/nix/store/zs9s8ncp22dpyvnhn858rpvpsch921g3-MAVProxy-1.8.53/bin/.mp_tile.py-wrapped", line 644, in <module>
    mt = MPTile(debug=opts.debug, service=opts.service,
  File "/nix/store/zs9s8ncp22dpyvnhn858rpvpsch921g3-MAVProxy-1.8.53/bin/.mp_tile.py-wrapped", line 208, in __init__
    raise TileException('unknown tile service %s' % service)
__main__.TileException: unknown tile service OviHybrid
```

MicrosoftSat was chosen as the default to match the default parameters of `MPTile`


stephendade:
Thanks! merging

PR #1101, opened 2022-09-02
SecureCommand: added module for secure bootloader updates

tridge:


davidbuzz:
Ive had a bit of a look, looks coherent so far, will re-read again tomorrow

PR #1102, opened 2022-09-15
mavproxy_param: fix --continue mode

peterbarker:


peterbarker:
--continue still doesn't work, but this is closer to it

PR #1116, opened 2022-10-31
Better filtering of MAVLink components

stephendade:
This adds the appropriate filters so that ``MAV_COMP_ID_PERIPHERAL`` devices can be used without confusing MAVProxy.

EDIT:
This PR now allows any MAVLink accessories (same sysid as flight controller, different component id) to:
a) Not confuse MAVProxy if the accessory is sending a heartbeat
b) Selecting the accessory in the console->vehicle->select menu correctly filters packets from the accessory to non-multi-vehicle-aware modules.

### DIFF
--- a/MAVProxy/modules/lib/mp_module.py
+++ b/MAVProxy/modules/lib/mp_module.py
@@ -205,9 +205,11 @@ class MPModule(object):
             label = str(link.linknum+1)
         return label
 
-    def is_primary_vehicle(self, msg):
+    def message_is_from_primary_vehicle(self, msg):
         '''see if a msg is from our primary vehicle'''
         sysid = msg.get_srcSystem()
-        if self.target_system == 0 or self.target_system == sysid:
+        compid = msg.get_srcComponent()
+        if ((self.target_system == 0 or self.target_system == sysid) and
+           (self.target_component == 0 or self.target_component == compid)):
             return True
         return False
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -615,7 +615,7 @@ class LinkModule(mp_module.MPModule):
 
     def master_msg_handling(self, m, master):
         '''link message handling for an upstream link'''
-        if self.settings.target_system != 0 and m.get_srcSystem() != self.settings.target_system:
+        if not self.message_is_from_primary_vehicle(m):
             # don't process messages not from our target
             if m.get_type() == "BAD_DATA":
                 if self.mpstate.settings.shownoise and mavutil.all_printable(m.data):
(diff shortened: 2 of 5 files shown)


stephendade:
PR has now been re-engineered for the general case:
- ``self.is_primary_vehicle(m)`` now checks for component id matching, instead of using a hardcoded component id blacklist.

MAVProxy/modules/lib/mp_module.py
```diff
@@ -208,6 +208,8 @@ def link_label(link):
     def is_primary_vehicle(self, msg):
```
peterbarker, line 208:
```suggestion
    def message_is_from_primary_vehicle(self, msg):
```

MAVProxy/modules/mavproxy_link.py
```diff
@@ -853,12 +846,8 @@ def master_callback(self, m, master):
                 if mtype_instance not in self.status.msg_count:
                     self.status.msg_count[mtype_instance] = 0
                 self.status.msg_count[mtype_instance] += 1
-        
-        if m.get_srcComponent() == mavutil.mavlink.MAV_COMP_ID_GIMBAL and mtype == 'HEARTBEAT':
-            # silence gimbal heartbeat packets for now
-            return
 
-        if getattr(m, 'time_boot_ms', None) is not None and self.settings.target_system == m.get_srcSystem():
+        if getattr(m, 'time_boot_ms', None) is not None and self.is_primary_vehicle(m):
```
peterbarker, line 850: This is a good change - components won't necessarily have the same timestamp as the parent vehicle.

MAVProxy/modules/mavproxy_link.py
```diff
@@ -893,6 +882,9 @@ def master_callback(self, m, master):
             for (mod,pm) in self.mpstate.modules:
                 if not hasattr(mod, 'mavlink_packet'):
                     continue
+                # Do not send other-system-or-component heartbeat packets to non-multi-vehicle modules
```
peterbarker, line 885: Will this break `mavproxy_gimbal.py`?  That's the first module I thought of...

MAVProxy/modules/mavproxy_link.py
```diff
@@ -541,7 +541,7 @@ def emit_accumulated_statustext(self, key, id, pending):
 
     def master_msg_handling(self, m, master):
         '''link message handling for an upstream link'''
-        if self.settings.target_system != 0 and m.get_srcSystem() != self.settings.target_system:
+        if not self.is_primary_vehicle(m):
```
peterbarker, line 544: This will break multi-vehicle-capable modules, won't it?

MAVProxy/modules/mavproxy_link.py
stephendade: Checked - gimbal still works. I'm only _not_ sending heartbeats here

MAVProxy/modules/lib/mp_module.py
stephendade: Fixed

MAVProxy/modules/mavproxy_link.py
stephendade: Tested in SITL with 10-copters and the "swarm" module. All works as expected

stephendade:
Using ``sim_vehicle.py --vehicle=Copter --console --map --count=10 --auto-sysid`` works fine

The changes in ``master_callback()`` still allow packets from the non-target vehicle to be passed through to multi-vehicle-aware modules (line 886 of mavproxy_link)

peterbarker (approved):
fence module still says "fence removed" / "fence present" when moving between vehicles

Testing went well.

peterbarker:
Merged, thanks!

PR #1117, opened 2022-11-03
cope with recent pymavlink string changes

tridge:
ping @alehed

### DIFF
--- a/MAVProxy/modules/mavproxy_swarm.py
+++ b/MAVProxy/modules/mavproxy_swarm.py
@@ -727,8 +727,8 @@ class SwarmFrame(wx.Frame):
                 elif mtype == "STATUSTEXT" and msg.severity <= mavutil.mavlink.MAV_SEVERITY_WARNING:
                     # Only pass on warning messages or more severe
                     widget.addstatustext(msg.text)
-                elif mtype == 'PARAM_VALUE' and msg.param_id in self.parmsToShow:
-                    widget.updateOffset(msg.param_id, int(msg.param_value))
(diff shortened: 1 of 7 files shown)


MAVProxy/modules/mavproxy_swarm.py
```diff
@@ -939,7 +939,7 @@ def mavlink_packet(self, m):
                 self.vehicleLastHB[(sysid, compid)] = time.time()
 
             # updated leader information from vehicle
-            elif mtype == 'PARAM_VALUE' and m.param_id == "FOLL_SYSID":
+            elif mtype == 'PARAM_VALUE' and m.param_id.decode('ascii') == "FOLL_SYSID":
```
alehed, line 942: Here I would compare to a bytes literal.

alehed (commented):
Looks good. I made one comment about avoiding a decode call. Does mavproxy specify which version of pymavlink to use with it?

If not, we probably need to stay compatible with older versions of pymavlink, in which case param_id can also be a string. Calling decode on a string fails, so in that case there would have to be a type check in the code.

alehed:
There are probably more changes needed. For instance in [MAVProxy/modules/mavproxy_ais.py‎](https://cs.github.com/ArduPilot/MAVProxy/blob/22ffc05170df048b9465ae2f664d272e19209b22/MAVProxy/modules/mavproxy_ais.py?q=owner%3AArduPilot+language%3Apython+callsign) on line 102, there is a str() which should be a decode. I will try to find a way to exhaustively find these cases.

tridge:
not needed any more

PR #1118, opened 2022-11-08
map: support ahrs3 vehicle position (copter)

joshanne:
This adds an orange vehicle variant of the copter and rover to the repo. This is used when mapping ahrs2/ahrs3 for these vehicle types.

Without these, attempting to show ahrs3 on the map for copter/rover causes the map module to error.


joshanne:
@tridge looks like I can't add the DevCall labels to the PR.

tridge (approved):

tridge:
thanks!

PR #1119, opened 2022-11-10
Support MAV_CMD_AIRFRAME_CONFIGURATION

LupusTheCanine:
Adds support for [MAV_CMD_AIRFRAME_CONFIGURATION](https://mavlink.io/en/messages/common.html#MAV_CMD_AIRFRAME_CONFIGURATION) this command controls landing gear. 
I would consider shorting "landinggear" to something shorter like "gear" as long as it is still easy to be found.

### DIFF
--- a/MAVProxy/modules/mavproxy_misc.py
+++ b/MAVProxy/modules/mavproxy_misc.py
@@ -106,6 +106,8 @@ class MiscModule(mp_module.MPModule):
         self.add_command('formatsdcard', self.cmd_formatsdcard, "format SD card")
         self.add_command('canforward', self.cmd_canforward, "enable CAN forwarding")
 
+        self.add_command('gear', self.cmd_landing_gear, "landing gear control")
+        
         self.repeats = []
 
     def altitude_difference(self, pressure1, pressure2, ground_temp):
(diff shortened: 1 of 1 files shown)


peterbarker (commented):
Do you think we should just add a `gear` command to `mavproxy_misc.py` instead of a full-blown module?

Also, aliases (or just rename) of `up` and `down` would seem to make sense to me...

LupusTheCanine:
I think that would be good. While I like single responsibility modules as it makes it easier to find things later it will be much more convenient to have it in misc.

`gear up` and `gear down` with `retract` and `extend` as aliases sounds good to me.

It would be good to add documentation page for misc too I think.

peterbarker:
@LupusTheCanine looks like you committed a `mav.parm` in here :-)

Please remove that and squash everything into one commit.  Commit comment should be something like `'mavproxy_misc: add gear command for controlling landing gear"

This has been tested?

LupusTheCanine:
I tested it with Ardupilot SITL. 
I will test with hardware (Arduplane updatetd to support the message)
May I add *.param to gitignore or should it be a separate pull request @peterbarker?

peterbarker:
On Tue, 15 Nov 2022, Lupus the Canine wrote:

> I tested it with Ardupilot SITL.
> I will test with hardware (Arduplane updatetd to support the message)
> May I add *.param to gitignore or should it be a separate pull request @peterbarker?

I'm not sure the ignore would be accepted.

I'd suggest in this case "git add -p" would be what you wanted to use to 
get things ready to commit :-)

LupusTheCanine:
@peterbarker 
Finished testing with Arduplane (modified to support the command).
Cleaned up and squashed.
Should be ready.

peterbarker (approved):

peterbarker:
merged, thanks!

PR #1124, opened 2022-11-23
movinghome: Rewrite movinghome module to use NMEAGPS module

booo:
The NMEAGPS if activated receives position updates from a NMEA source. We can use the position information in other modules such as this one. This change reduces code duplication and ensures multiple modules can use one NMEA source.

This is still subject to change and testing. But I would like to get some comments on the idea.

### DIFF
--- a/MAVProxy/modules/mavproxy_movinghome.py
+++ b/MAVProxy/modules/mavproxy_movinghome.py
@@ -15,11 +15,7 @@ from pymavlink import mavutil
 import errno
 import time
 import math
-import serial
-try:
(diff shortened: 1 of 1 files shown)


julled:
if this now has a  hard dependency on an running NMEAGPS module, this should be checked in init and told to the user if not the case.

booo:
I'm not sure if we are able to check if a module is loaded. And there could be different modules providing position information.

We can check if no position update arrives and inform the user about that from time to time. The old module did something similar.

PR #1133, opened 2022-12-14
Modules: Serial tunnel print and timeout fix

stephendade:
Requires https://github.com/ArduPilot/ardupilot/pull/22397

Needed to reduce timeout to avoid a ``main_loop_stk`` error in ArduPilot

### DIFF
--- a/MAVProxy/modules/mavproxy_serial.py
+++ b/MAVProxy/modules/mavproxy_serial.py
@@ -16,7 +16,7 @@ class SerialModule(mp_module.MPModule):
         self.serial_settings = mp_settings.MPSettings(
             [ ('port', int, 0),
               ('baudrate', int, 57600),
-              ('timeout', int, 500)
+              ('timeout', int, 350)
               ]
             )
         self.add_completion_function('(SERIALSETTING)', self.serial_settings.completion)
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_serial.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, mpstate):
         self.serial_settings = mp_settings.MPSettings(
             [ ('port', int, 0),
               ('baudrate', int, 57600),
-              ('timeout', int, 500)
```
peterbarker, line 19: Shouldn't this fix be in ArduPilot itself?  That would usually be by flagging to the scheduler that we're going to be off doing other things, and is important because the hardware watchdog probably won't be being patted either without a fix in AP.

MAVProxy/modules/mavproxy_serial.py
```diff
@@ -28,6 +28,7 @@ def mavlink_packet(self, m):
             data = m.data[:m.count]
             s = ''.join(str(chr(x)) for x in data)
             sys.stdout.write(s)
+            print("")
```
peterbarker, line 31: This may break some people's workflows.

peterbarker (commented):
How does one replicate the timeout problem?

Issue #700, opened 2019-11-01
Waypoints indefinitely re-requested when using MAVProxy with --out option


mkunzBGC:
This is similar to #402, but SITL is not involved.

I run mavproxy in passthrough mode using
```
mavproxy.py --master=/dev/serial/by-id/usb-FTDI_FT230X_Basic_UART_DN02RM7W-if00-port0\
            --out=udp:127.0.0.1:14550
```
and let QGroundControl connect to that udp port. The specified serial device is a radio connected to a Cube autopilot running APM:Copter V3.6.11.

Once the parameters are transmitted, I see a new message `re-requesting WPs [0, 1, 2, 3, 4]` every other second in MAVProxy's terminal output.

The workaround of using `wp list` stops the re-requesting of waypoints.

I am using MAVProxy 1.8.15.

magicrub:
should be fixed via https://github.com/ArduPilot/MAVProxy/pull/1152

Issue #918, opened 2021-08-24
MAVproxy infinitely re-requesting WPs after mission list by other GCS


domino-pl:
We use MAVProxy as relay between UAV and GCS. 

I found an issue in MAVProxy. After requesting mission download from UAV to GCS, MAVProxy in the middle starts send mission requests periodically. It often disturbs firstly started mission download process. 

I did some research in file `mavproxy_wp.py` and discovered that:

Every MISSION_COUNT message is process and value of expected mission count is set, but operation mode self.wp_op is not changed [(code)](https://github.com/ArduPilot/MAVProxy/blob/2dfcea92597f6f37448d4c76a8b5f4b58956b37e/MAVProxy/modules/mavproxy_wp.py#L149):
```
if mtype in ['WAYPOINT_COUNT','MISSION_COUNT']:
    self.wploader.expected_count = m.count
    [...]
```

However in [(code)](https://github.com/ArduPilot/MAVProxy/blob/2dfcea92597f6f37448d4c76a8b5f4b58956b37e/MAVProxy/modules/mavproxy_wp.py#L214) timeout handler will always send new mission request if number of downloaded item is less then just received `self.wploader.expected_count`, even `if self.wp_op != "list"`:
```
def idle_task(self):
    '''handle missing waypoints'''
    if self.wp_period.trigger():
        # cope with packet loss fetching mission
        if self.master is not None and self.master.time_since('MISSION_ITEM') >= 2 and self.wploader.count() < getattr(self.wploader,'expected_count',0):
            wps = self.missing_wps_to_request();
            print("re-requesting WPs %s" % str(wps))
            self.send_wp_requests(wps)
```

UAV will send mission list to MAVProxy, but due to  `self.wp_op != list` this items will never be processed.

I added additional condition `and self.wp_op == "list"` in `idle_task` to workaround problem. 

Maybe also target system/component ids and mission type fields in received messages ought to be checked?

karthikdesai:
I got around it by issuing `wp list` in the console. The wp list is then re-requested and the console isn't flooded

magicrub:
should be fixed via https://github.com/ArduPilot/MAVProxy/pull/1152

PR #1152, opened 2023-02-16
fix re-request forever bug

magicrub:
If partial WP re-request fails then request all. This happens where there's mission transactions done by other GCSs using the same link and MAVProxy gets confused. This simply checks to see if it happened 5 times and it so just starts a whole new wp fetch to start over it's state machine.

Users like me have been suffering from this for YEARS. Luckily it's open source so I can fix it myself :) Open-source FTW!!!

Closes https://github.com/ArduPilot/MAVProxy/issues/700 and https://github.com/ArduPilot/MAVProxy/issues/918

### DIFF
--- a/MAVProxy/modules/mavproxy_wp.py
+++ b/MAVProxy/modules/mavproxy_wp.py
@@ -32,6 +32,7 @@ class WPModule(mp_module.MPModule):
         self.undo_wp_idx = -1
         self.upload_start = None
         self.wploader.expected_count = 0
+        self.wploader.retry_count = 0
         self.last_get_home = time.time()
         self.add_command('wp', self.cmd_wp,       'waypoint management',
                          ["<list|clear|move|remove|loop|set|undo|movemulti|moverelhome|changealt|param|status|slope|ftp|add_takeoff|add_landing|add_dls|add_rtl>",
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_wp.py
```diff
@@ -190,9 +191,16 @@ def idle_task(self):
         if self.wp_period.trigger():
             # cope with packet loss fetching mission
             if self.master is not None and self.master.time_since('MISSION_ITEM') >= 2 and self.wploader.count() < getattr(self.wploader,'expected_count',0):
-                wps = self.missing_wps_to_request();
-                print("re-requesting WPs %s" % str(wps))
-                self.send_wp_requests(wps)
+                self.wploader.retry_count = self.wploader.retry_count + 1 
+                if self.wploader.retry_count > 5:
```
tridge, line 195: this will avoid that annoying retry forever, but it will mean that if you are on a lossy link with lots of WPs then you'll never complete as it will keep restarting

tridge (changes requested):

davidbuzz:
I tend to 'module unload WP' in those circumstances, as the downstream gcs is the one the operator sits in front of... ie MP.  would giving up and auto-unloading the module be ok?

magicrub:
I'm open to suggestions.. maybe multiple retries in a row with perfect link quality? Maybe check for a packet of sometime requesting the list?

rmackay9:
I've tested and confirmed that this is not an issue of conflicting ground station system ids.

I've easily reproduced the problem on my machine using WSL2.

1. run sim_vehicle for copter using master (I ran this under WSL2)
2. start Mission Planner (with GCS's sysid = 249), it should connect automatically
3. On MP's Plan screen add one waypoint, push Write then Read

MAVProxy produces a "re-requesting WPs [0,1] endlessly

peterbarker:
We've merged  an alternative fix to this one.  Thanks for prompting us to move forward on this.

rmackay9:
fantastic!

PR #1153, opened 2023-02-23
help: fixed help for parameters with both range and values

tridge:


peterbarker (approved):
Looks likely.  I thought we were eliminating these from AP, 'though.

Looking at an attribute might have been nicer than the stringification, maybe.

PR #1154, opened 2023-03-01
map: use command_int when sending ROI

peterbarker:
ArduPilot interprets the command_long frame as relative-alt - we are sending through an AMSL alt of the terrain for the lat/lon.

Using command_int means we remove this problem by actually specifying our frame.

This work sponsored by Harris Aerial


peterbarker:
.... we've supported this since Feb 2016 in ArduPilot....

tridge (approved):

PR #1156, opened 2023-03-02
Use mavlink DO_MOUNT_CONTROL in place of old MOUNT_CONTROL message

peterbarker:
ArduPIlot has started to send deprecation warnings for the old mount messages and they will be removed at some point.

This work sponsored by Harris Aerial

### DIFF
--- a/MAVProxy/modules/mavproxy_gimbal.py
+++ b/MAVProxy/modules/mavproxy_gimbal.py
@@ -90,14 +90,20 @@ class GimbalModule(mp_module.MPModule):
             print("No map click position available and no parameters set")
             return
         elif latlon is None:
-            print("usage: gimbal roi [LAT LON ALT]")
+            print("usage: gimbal roi [LAT LON RELHOMEALT]")
             return
-        self.master.mav.mount_control_send(self.target_system,
-                                           self.target_component,
-                                           int(latlon[0]*1e7),
-                                           int(latlon[1]*1e7),
-                                           int(alt),
-                                           0)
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):

MAVProxy/modules/mavproxy_gimbal.py
```diff
@@ -90,14 +90,20 @@ def cmd_gimbal_roi(self, args):
...
-            print("usage: gimbal roi [LAT LON ALT]")
+            print("usage: gimbal roi [LAT LON RELHOMEALT]")
             return
-        self.master.mav.mount_control_send(self.target_system,
-                                           self.target_component,
-                                           int(latlon[0]*1e7),
-                                           int(latlon[1]*1e7),
-                                           int(alt),
-                                           0)
+        self.master.mav.command_long_send(
```
rmackay9, line 95: should this be a command_int instead perhaps?

peterbarker: Funny you should mention that - I thought just the same thing.

I brought up making ArduPilot understanding these as `COMMAND_INT` messages at the most recent DevCall - but got shot down.

So it's `COMMAND_LONG` for now.

rmackay9: I think I may have heard the tailend of this but was occupied so I couldn't comment.  Definitely we should be trying to be as accurate as possible especially when it costs us nothing as far as I can tell.

PR #1159, opened 2023-03-08
mavproxy_ftp: ignore FTP packets not targetted at me

peterbarker:
stop us spewing lots of warnings when someone else is doing a transfer


tridge:
a bit of testing would be good

peterbarker:
Tested.

PR #1165, opened 2023-03-19
speech: added support for "@" prefix for speech output

tridge:
used by aerobatics scripting

### DIFF
--- a/MAVProxy/modules/mavproxy_speech.py
+++ b/MAVProxy/modules/mavproxy_speech.py
@@ -116,8 +116,10 @@ class SpeechModule(mp_module.MPModule):
         type = msg.get_type()
         if type == "STATUSTEXT":
             # say some statustext values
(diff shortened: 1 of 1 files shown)


peterbarker:
Wondering if @hamishwillee has input on this.

The basic premise being any statustext received with `@` as the first character will be read out loud by a GCS.  The standard staustext levels really aren't the right thing for this.

I can't think of any simple alternative to this.  Best I've got is to send the statustext from a different component ID and configure the GCS to read out any message coming from that component ID.  There's a *lot* of work required to make that sort of thing work in ArduPilot.

tridge:
this is so simple, and easy for lua scripts.

hamishwillee:
The STATUSTEXT content is entirely up to the flight stack. It would be nice to standardize on standard prefixes so that GCS can have common handling. But that's a bigger conversation.

PR #1166, opened 2023-03-21
MAVExplorer: add mission file save and help

Hwurzburg:
needs the header to be loadable
needs some extra tabs for 0,0 lat long case to prevent columns with no space between


tridge:
I think ideally we'd use "mission save FILENAME" and that would have the header

Hwurzburg:
@trdige done and tested

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #1167, opened 2023-03-23
mavproxy_wp: use MAV_CMD_DO_SET_MISSION_CURRENT to set waypoints

peterbarker:
Sends both the old message and the new command until we know for certain if the autopilot is going to take the command (if it doesn't then we should receive a `MAV_CMD_UNSUPPORTED`)


peterbarker:
@hamishwillee @meee1 tridge point out that we do open ourselves to some race conditions if we send both commands.  So now we just send the mavlink command unless a setting says otherwise (or we detect that the vehicle doesn't do the command)

tridge (approved):

hamishwillee:
> @hamishwillee  tridge point out that we do open ourselves to some race conditions if we send both commands. So now we just send the mavlink command unless a setting says otherwise (or we detect that the vehicle doesn't do the command)

Thanks @peterbarker - I think I pointed out I prefer this approach too. Either way, I do.

PR #1168, opened 2023-03-23
mavproxy_link: MAVProxy locks onto both sysid and compid based on heartbeat

peterbarker:
prior to this PR MAVProxy only locked onto a system ID.

This leads to the situation that if you send a command you send it to all components of the vehicle.  This is probably not desired behaviour - by default you should probably only target the primary "autopilot" the vehicle.  An example of where this could be bad is MAV_CMD_DO_SET_MODE - it's entirely reasonable for a component to have its own set of custom modes, and if it processes MAVProxy's DO_SET_MODE intended for the autopilot may have unintended consequences

This PR changes things so we set both system ID and component ID when locking onto the vehicle based off heartbeat, so commands will be targetted at e.g. 1/1 rather than 1/0

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -539,6 +539,56 @@ class LinkModule(mp_module.MPModule):
             self.status.last_apm_msg_time = time.time()
         del self.status.statustexts_by_sysidcompid[key][id]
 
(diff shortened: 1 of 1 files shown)


peterbarker:
I've tested this a chunk in SITL.

tridge (approved):

PR #1171, opened 2023-04-13
speech: move speech to a subprocess

tridge:
this should avoid lockups caused by speech backends


peterbarker (commented):
speech stops working if you unload then load the module again

Apart from that, looks good!

tridge:
> speech stops working if you unload then load the module again

which backend do you get?

PR #1175, opened 2023-04-24
grapher: attempt to fix windows close graph error

tridge:
maybe using plotlib timer instead of threading timer will help


tridge:
confirmed that this does fix the issue on windows!

PR #1182, opened 2023-05-20
Dial control type

landswellsong:
Made a new control type that uses the hat controls to update a value in the direction of the control. Can either center the output value at the startup (default) or start at the low value.

Also fixed some typos in the hat description and an apparent bug with `outhigh`/`outlow` not being supplied to its constructor.


landswellsong:
I realized I opened this a while ago :O 
Rebased and repushed, please consider

PR #1193, opened 2023-07-12
mavproxy_map: add context-menu option for keeping window focus

mtekman:
The map module tends to steal focus from the terminal when being
  invoked which makes it hard to issue custom commands when the map
  module is live.  This commit creates a new Menu option "Focus Map"
  that toggles this behaviour on and off.


tridge:
@mtekman this turns out to be a new issue with the latest wx version. I've removed the set focus call in the lastest builds, so this is no longer an issue.
thanks for the contrib though!

mtekman:
oh nice, thanks for fixing this!

PR #1195, opened 2023-07-13
console: give consistent ordering of user added items

tridge:
display a "" initially to lock in location


peterbarker (approved):
Perhaps "?" rather than "" would be nicer.

tridge:
> Perhaps "?" rather than "" would be nicer.

this is closer to what people get now, it doesn't display till they get telem

Issue #1215, opened 2023-08-05
BUG: Mavproxy from Pip is missing dependencies causing Import Errors.


Ryanf55:
# Information

* OS: Ubuntu 23
* Python version: Python `3.11.4`
* Pip version: `pip-23.2.1`
* Method to install mavproxy: pip

## Steps to reproduce

```bash
python3 -m venv .mavproxy
. .mavproxy/bin/activate
python3 -m pip install mavproxy
which mavproxy.py 
>>> /home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/bin/mavproxy.py
```
```
(.mavproxy) ryan@ryan-B650-970:~/Development/ardu_ws/src/MAVProxy$ mavproxy.py --console --map --aircraft test --master=:14550 --moddebug=3
Connect :14550 source_system=255
No module named 'wx'
Traceback (most recent call last):
  File "/home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/bin/mavproxy.py", line 679, in import_package
    mod = __import__(name)
          ^^^^^^^^^^^^^^^^
  File "/home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/lib/python3.11/site-packages/MAVProxy/modules/mavproxy_console.py", line 9, in <module>
    from MAVProxy.modules.lib import wxconsole
  File "/home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/lib/python3.11/site-packages/MAVProxy/modules/lib/wxconsole.py", line 11, in <module>
    from MAVProxy.modules.lib import win_layout
  File "/home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/lib/python3.11/site-packages/MAVProxy/modules/lib/win_layout.py", line 4, in <module>
    import os, wx, pickle
ModuleNotFoundError: No module named 'wx'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/bin/mavproxy.py", line 526, in load_module
    m = import_package(modpath)
        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ryan/Development/ardu_ws/src/MAVProxy/.mavproxy/bin/mavproxy.py", line 682, in import_package
    mod = __import__(name)
          ^^^^^^^^^^^^^^^^
[...]

PR #1216, opened 2023-08-05
Setup: Add missing dependencies…

Ryanf55:
Adds matplotlib, opencv-python and wxPython to all distributions. Works on Linux using a virtual environment. I also converted the requirements from a list to set to prevent duplicates. 

Note: There is comment in the source code saying that requiring dependencies breaks things:
```
# note that we do not include all the real dependencies here (like matplotlib etc)
# as that breaks the pip install. It seems that pip is not smart enough to
# use the system versions of these dependencies, so it tries to download and install
# large numbers of modules like numpy etc which may be already installed
```

I tried building and installing, and it didn't break my pip install. I used a virtual environment. This comment may no longer be valid. 

Solves #1215


tridge (changes requested):
this would break the way mavproxy is often installed on headless companion computers. We could add them only on windows though

PR #1217, opened 2023-08-07
Add manipulation of return point location

peterbarker:
- can right-click-menu-item add a return point
 - can remove that return point with a right-click on the point


peterbarker:
image

.... vehicle does fly back to that if you configure the vehicle appropriately.

PR #1220, opened 2023-08-10
map: fixed menu on map with multiple objects

tridge:
when you right click and the position matches multiple objects then check each of the objects rather than just the first one

example with both polygon and wp matching the click point
image


peterbarker (approved):
Works for me.

PR #1222, opened 2023-08-10
mavproxy_mode: use MAV_CMD_DO_REPOSITION for 'Fly To'

peterbarker:
this has the advantage that you don't need to be in guided mode to start the command.  Prevents some clunk interactions between pilot and GCS operator as the vehicle does not need to be somewhere safe to loiter before the command can be safely issued.

Tested Copter, Plane and Rover in SITL.  Altitudes appear to be correct.


tridge:
my main concern is an accidental click when not in GUIDED mode. We could use confirm, like we now do for set home, but maybe we should confirm only if not currently in GUIDED?

peterbarker:
> my main concern is an accidental click when not in GUIDED mode. We could use confirm, like we now do for set home, but maybe we should confirm only if not currently in GUIDED?

"Fly To" pops up an altitude box - if you cancel that dialogue box rather than giving it a value then the operation does nothing - the vehicle doesn't change modes or anything.  I considered that confirmation enough :-)

tridge (approved):

PR #1225, opened 2023-08-14
ftp: warn of incorrect component ID

tridge:
allow user to see bug https://github.com/ArduPilot/ardupilot/pull/24633
note that you can work around the bug with "set source_component 190" to match MissionPlanner


peterbarker (approved):
Hopefully only as an interim measure.

PR #1230, opened 2023-08-26
added SIYI camera module

tridge:
for basic control of a SIYI camera


rmackay9:
I wonder if we can use some of this for a more general purpose custom camera/gimbal screen for MAVProxy?  We're pretty stuck with MP in particular not making any progress so I hope that showing how it can be done with MAVProxy will help.

Issue #1162, opened 2023-03-16
Mission Editor: scroll down no longer works


rmackay9:
The Mission Editor has a very long list of mission commands (far more than any of the vehicles support but this is a separate issue) and I seem to be unable to use the mouse scroll button to scroll down the list meaning that only the top of the list is usable.

Below is an example of what I can see on my screen, there's no way as far as I can tell to select items below DO_MOUNT_CONFIGURE
![image](https://user-images.githubusercontent.com/1498098/225532409-669375ef-8bfe-464f-8207-98ac16b14288.png)

rmackay9:
Hi, I really struggled with this today because I was trying to build a mission on my WSL2 box.  This bug wastes a lot of time for me actually.

stephendade:
What version of wxpython are you using? Use ``pip3 list`` to get the version.

I'm using 4.0.7

On my Ubuntu box, I'm getting scroll buttons:
![Screenshot from 2023-06-27 11-30-47](https://github.com/ArduPilot/MAVProxy/assets/1731976/134742e2-9c15-4435-8dcf-7c149066ff4a)

rmackay9:
Hi @stephendade,

I'm apparently also using WsPython 4.0.7..

stephendade:
I've confirmed the issue on WSL.

Interestingly, if I run MAVProxy from Windows directly (not via WSL) I do get the scrollbars. Maybe be some issue with the display or layout manager in WSL?

PR #1233, opened 2023-08-29
misseditor: Better mission command dropdown

stephendade:
Fix for #1162.

Modifies the mission editor's "Command" dropdown to use a custom control (ListCtrl), which has a better layout and visible scrollbars.

Tested on WSL and Ubuntu 22:

Screenshot 2023-08-29 161219

@rmackay9 - I'd be interested if this works better for you.

### DIFF
--- a/MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
+++ b/MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
@@ -39,6 +39,140 @@ ME_DOWN_COL = 11
 ME_DIST_COL = 12
 ME_ANGLE_COL = 13
 
+
+class ListCtrlComboPopup(wx.ComboPopup):
+
+    def __init__(self):
+        wx.ComboPopup.__init__(self)
+        self.lc = None
+
+    def AddItem(self, txt):
+        self.lc.InsertItem(self.lc.GetItemCount(), txt)
+
+    def OnMotion(self, evt):
+        item, flags = self.lc.HitTest(evt.GetPosition())
+        if item >= 0:
+            self.lc.Select(item)
+            self.curitem = item
+
+    def OnLeftDown(self, evt):
+        self.value = self.curitem
+        self.Dismiss()
+
+    # This is called immediately after construction finishes.  You can
+    # use self.GetCombo if needed to get to the ComboCtrl instance.
+    def Init(self):
+        self.value = -1
+        self.curitem = -1
+
+    # Create the popup child control.  Return true for success.
+    def Create(self, parent):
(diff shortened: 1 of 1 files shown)


rmackay9:
Thank you so much!

stephendade:
@rmackay9, would you be able to give this a test-drive? Mostly just to see if the GUI changes work for you.

I'll get @tridge or @peterbarker to do the actual code review.

rmackay9:
Hi @stephendade,

Yes, I'll give this a test.  I'm not very confident about building MAVproxy from source but I'll try.

rmackay9:
Hi @stephendade,

I gave this a test and it is certainly better than master because I can actually get to the commands I'd like to add.

It is a bit weird though, sometimes when I click on an existing command and try to change it, the drop-down appears behind the wp editor.
image

stephendade:
> It is a bit weird though, sometimes when I click on an existing command and try to change it, the drop-down appears behind the wp editor. It also appeared to me that at least once the list of commands was not sorted alphabetically although I'm now having troubles reproducing this.

I've managed to replicate this. It's a x11 vs Wayland issue again. @MichelleRos had similar issues and fixed it via putting ``GDK_BACKEND="x11"`` in ``~\.bashrc`` and restarting.

I've confirmed if you do the same, it should fix the window focus issue.

rmackay9 (approved):
@stephendade,

Yes, that's fixed it for me as well.  I had to add "**export** GDK_BACKEND="x11"".

peterbarker (changes requested):
@stephendade I can reliably produce this assertion:
```
wx._core.wxAssertionError: C++ assertion "firstHandler != this" failed at ../src/common/wincmn.cpp(1461) in PopEventHandler(): cannot pop the wxWindow itself

The above exception was the direct cause of the following exception:

SystemError: _PyEval_EvalFrameDefault returned a result with an exception set

```

... by selecting a type which ArduPilot doesn't undertstand, writing waypoints then reading waypoints.

stephendade:
> @stephendade I can reliably produce this assertion:
> 
> ```
> wx._core.wxAssertionError: C++ assertion "firstHandler != this" failed at ../src/common/wincmn.cpp(1461) in PopEventHandler(): cannot pop the wxWindow itself
> 
> The above exception was the direct cause of the following exception:
> 
> SystemError: _PyEval_EvalFrameDefault returned a result with an exception set
> ```
> 
> ... by selecting a type which ArduPilot doesn't undertstand, writing waypoints then reading waypoints.

Now fixed. It was a reference counting issue (see https://github.com/wxWidgets/Phoenix/issues/627)

peterbarker:
Merged, thanks!

PR #1278, opened 2023-12-09
Add OpenAI compatible Chat module

rmackay9:
This adds a "chat" module that can interface with an OpenAI chatGPT-4 Assistant and control an Ardupilot vehicle (Copter, Plane, Rover, Sub).  Note that running this requires either using my OpenAI API key or you must have your own and then run the provided setup_assistant.py script.

Below are some screenshots showing a typical interaction
chat-screen
rtl-home
console

There are some known issue we may want to resolve before merging:

- [x] add MAVProxy package dependency on "openai"
- [x] how should we deal with the need for the OpenAI API key?  [ArduPilot will likely fund at least the core devs for now](https://discuss.ardupilot.org/t/openai-business-account/110388).
- [ ] allow more permanent storing of OpenAI API key
- [x] reduce debug output to console
- [x] remove mavlink xml files from "assistant_setup" directory
- [ ] improve voice recording (currently always 5sec, should always "listen" or use threshold to turn off)

Corresponding MAVProxy wiki PR https://github.com/ArduPilot/ardupilot_wiki/pull/5651

Some references:

- [OpenAI Assistant API](https://platform.openai.com/docs/api-reference/assistants)
- [OpenAI Assistant Playground](https://platform.openai.com/playground)


stephendade:
Very cool! I'd be interested if it could help users diagnose any issues with their configuration. Something like "why is my copter not arming"

tpwrules:
I think this is a neat demo (though it does have its own problems) but I don't think the additional requirements justify it. (I think you forgot some too). This would mean that every SITL run, every distro package, every CI invocation, etc. will need the additional packages.

Is there a supported framework for external modules?

requirements.txt
```diff
@@ -1,3 +1,4 @@
 pymavlink>=2.4.14
 pyserial>=3.0
 numpy
+openai
```
tridge, line 4: instead of in requirements I'd prefer it give a message on load "need to install openai with 'python -m pip install openai'

tridge (changes requested):

tridge:
> Is there a supported framework for external modules?

there is, but its a bit clunky

requirements.txt
rmackay9: ok, removed openai from requirements.

peterbarker (approved):
So cool!

ShiauweiZhao:
cool

Issue #1281, opened 2023-12-13
Chat module issues & enhancements list


rmackay9:
This is a consolidated list of issues and enhancements for the MAVProxy chat module added with PR https://github.com/ArduPilot/MAVProxy/pull/1278

Bugs:

- [ ] typing "chat show" into the MAVProxy produces an innocuous error.  We should either remove that completely or make it re-show the chat GUI in case the user has closed it
- [ ] Under Windows the user input text is grey instead of red meaning it is slightly more difficult to see what is your text and what is the chatbot's reply.
- [ ] Voice-to-text recording is currently hard-coded to 5sec.  We should change this to a "Listen".  So when the user pushes the button it will stay depressed and record when the volume is over a given threshold.  Once it detects the end of a sentence (e.g. volume drops below the threshold for maybe 2 or 3 seconds) it sends to openai.

Enhancements:

- [ ] Improve reliability of takeoff command and "please fly here".  -- Randy has as WIP change for this
- [ ] Add function to allow AI to retrieve the latest mavlink messages sent by the vehicle ()
- [ ] Add function to allow AI to send arbitrary mavlink messages to the vehicle
- [ ] support "streaming" once it's availble from OpenAI (this will make the long lags more bearable because text will appear like it does on the ChatGPT web page.

For real-time discussion there is an ArduPilot "ai" discord channel here.

YousefYaghoobi:
Hi Randy.
Thanks for such an interesting project.

I have tried to follow along with this project.
After installation and using the ''CHAT" module:
 - when I ask "What vehicle am I using?" or "What is the vehicle location? the answers are correct;
 - when I try to "ARM" the vehicle or change the vehicle "MODE" I get the below errors:
     -  **"chat: handle_function_call: unrecognised function call: get_mode_mapping"**
     -  **"Got COMMAND_ACK: COMPONENT_ARM_DISARM: UNSUPPORTED"** (when I do the force arming)
 
I have tried for both SITL (ARDUCOPTER & ROVER) and real ROVER which is running on PIXHAWK 6C.

I was not sure whether or not to open a new issue.
Your help is highly appreciated.

rmackay9:
Hi @YousefYaghoobi,

[...]

PR #1283, opened 2023-12-14
Chat: improve takeoff and fly-to commands

rmackay9:
This PR includes two small enhancements to the chat module:

1. Instructions to the assistant have been updated to increase the chance that it will set the "frame" field correctly.
2. Add a new get_location_plus_offset function to help the assistant more quickly and easily respond to commands like, "fly 100m North".  Without this the assistant tries to work out the calculation itself each time and this leads to some failures

Both these changes have been tested and work.  Below are some screen shots from a recent test.
image

I should add that during testing I saw the chat window apparently freeze up twice probably waiting for responses from the Assistant that never arrived.  I haven't seen these freezes before but I assume they are unrelated to these changes and have added this to [the issues list](https://github.com/ArduPilot/MAVProxy/issues/1281).


rmackay9:
Thanks!

PR #1285, opened 2023-12-19
Chat: improved API key setting method

hfujikawa77:
Modified chat_voice_to_text.py to use the API Key set in the GUI's MENU in order to unify the API Key setting method.


rmackay9 (approved):
LGTM, this is consistent with what we do for the "chat_openai.py" file.

hfujikawa77:
@rmackay9 
Thanks for the review. If there are no problems, please merge.

tridge (approved):

PR #1288, opened 2023-12-21
Chat: minor GUI and assistant setup enhanecments

rmackay9:
This PR includes 3 enhancements and 1 bug fix:

Chat window improvements:

1. Reply window colours are fixed on Windows (user input now correctly appears in red so it can be more easily distinguished from the replies)
2. Blank text is no longer sent to the assistant in cases where the audio-to-text fails or the user simply presses the Send button with no text in the input box

Assistant setup script improvements

3. The assistant setup script re-uses previously uploaded files instead of creating duplicates.  This behaviour can be overridden with the "--upgrade" argument
4. an assistant_file_checker.py script helps find and delete files that were uploaded to OpenAI but never associated with any Assistant.  So far we haven't found any particular problem with having these unused files hanging around but there are known [issues with assistants occasionally not being able to find files](https://community.openai.com/t/assistant-not-able-to-access-uploaded-file/524495/13)... perhaps keeping the upload area orderly will help somehow.

These changes have all been tested on Windows and Ubuntu and it seems to be working OK.

BTW, I initially had a cool "listen" feature for the audio-to-text conversion but I wasn't able to get it reliable enough so I'll push that off to a future PR

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/assistant_setup/assistant_file_checker.py
+++ b/MAVProxy/modules/mavproxy_chat/assistant_setup/assistant_file_checker.py
@@ -1,0 +1,111 @@
+'''
+AI Chat Module Assistant File Checker
+Randy Mackay, December 2023
+
+This script is used to check for inconsistencies in the Assistants and Files.
+E.g. files which are not used by any Assistants or Assistants which use files which do not exist.
+
+OpenAI Assistant API: https://platform.openai.com/docs/api-reference/assistants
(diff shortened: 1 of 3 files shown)


rmackay9:
It's a bit long but here's a little interaction from testing on Windows.  They key points are that the area around the buttons is the correct colour now.  The user's input appears in red and there are more status updates appearing in the bottom left of the window.
image

PR #1289, opened 2023-12-26
Chat: reliability enhancements

rmackay9:
This improves the reliability in a few ways:

1. COMMAND_ACK messages are used to tell the assistant if the command was successful or not.
2. Assistant instructions are updated in a few ways:
     - Emphasize that it must check the vehicle type to protect against using the wrong vehicle's flight-mode-number when changing modes
     - Tell it to use CONDITION_YAW to change the vehicle's yaw
     - Emphasize takeoffs commands should only be sent after the vehicle is confirmed to be in Guided mode and armed

I could squash down the assistant instructions to a single commit if this is preferred.

I've tested this and it certain improves reliability.  Below is a typical test where the user asked it to takeoff before the arming checks cleared and it still managed to do it.

image

image

image


MAVProxy/modules/mavproxy_chat/__init__.py
```diff
@@ -26,6 +28,11 @@ def __init__(self, mpstate):
         # keep reference to mpstate
         self.mpstate = mpstate
 
+        # a dictionary of command_ack mavcmds we are waiting for
+        # key is the mavcmd, value is -1 or the MAV_RESULT (e.g. 0 to 9)
```
peterbarker, line 32: `None` can be a useful flag value.

MAVProxy/modules/mavproxy_chat/__init__.py
```diff
@@ -26,6 +28,11 @@ def __init__(self, mpstate):
...
+        # we assume we will never be waiting for two of the same mavcmds at the same time
```
peterbarker, line 33: You can't dis-ambiguate the results anyway!  I've pushed for the addition of an "opaque ID" supplied in the `COMMAND_INT` message and returned in a `COMMAND_ACK` to fix that problem, but didn't ever get traction on it...

You *can* ensure you never send a second command of a specific type while one is outstanding based on your dictionary, 'though.  You would then need a timeout as the ACK may get lost.

MAVProxy/modules/mavproxy_chat/__init__.py
```diff
@@ -55,6 +62,53 @@ def cmd_chat(self, args):
     def show(self):
         self.chat_window.show()
 
+    # handle mavlink packet
+    def mavlink_packet(self, m):
+        if m.get_type() == 'COMMAND_ACK':
```
peterbarker, line 67: Often worth creating an `m_type` in these straight-up.  `get_type` can be surprisingly expensive when called multiple times in `mavlink_packet`, and `elif` statements tend to accrete in `mavlink_packet` calls.

MAVProxy/modules/mavproxy_chat/__init__.py
```diff
@@ -55,6 +62,53 @@ def cmd_chat(self, args):
...
+        if mav_cmd in self.command_ack_waiting:
+            print("chat: already waiting for command ack for mavcmd:" + str(mav_cmd))
+            return False
+
+        # add to list of commands we are waiting for (-1 indicates we don't know result yet)
+        self.command_ack_waiting[mav_cmd] = -1
+
+        # wait for ack, checking for a response every 0.1 seconds
+        start_time = time.time()
+        while time.time() - start_time < timeout:
```
peterbarker, line 98: You can't do this in a lot of contexts as you will kill MAVProxy's main loop.

You should (a) test what happens if you sleep for 10 minutes in here and (b) add to the comment at the top of this method indicating that it's only OK to call this from the chat module's thread or whatever.

peterbarker (changes requested):
I'm really only fussed by that sleeping.

MAVProxy/modules/mavproxy_chat/__init__.py
rmackay9: ok, replaced -1 with None

MAVProxy/modules/mavproxy_chat/__init__.py
rmackay9: ok thanks.  The way the chat module works at the moment, the commands are only sent one-by-one and after sending it immediately waits for the ACK so there's actually no way for it to be waiting for two command_acks.  It could still pickup the wrong ACK though if that ACK is in response to a similar command from some other sender... and your proposed change would help with that so I'm all for it.

BTW I could have simplified the dictionary down to just a single name-value pair but in python a dictionary is so easy to use I went with it.

MAVProxy/modules/mavproxy_chat/__init__.py
rmackay9: OK noted.  I did notice that in other modules but in this case it only checks it once so I avoided the local variable for now anyway.

MAVProxy/modules/mavproxy_chat/__init__.py
rmackay9: Hi, in this case the chat window runs in a separate thread so MAVProxy's main loop is not affected.  I've added a comment to the top of the function to specify that it blocks and shouldn't be called from the main thread.

`# Note: this should not be called from the main thread because it blocks for up to "timeout" seconds`

I've also tested the long sleep in this function and it doesn't freeze up MAVProxy.  The chat module window can even still be resized and moved around but it gets stuck for a long time before it provides an update to the user.

peterbarker:
Merged, thanks!

rmackay9:
Thanks!!

PR #1290, opened 2023-12-27
Chat: add wakeup timer support

rmackay9:
This PR adds three new functions that allow the Assistant to set, get or delete wakeup timers.

Having timer support allows the Assistants to more easily handle request to perform a sequence of tasks.  For example, fly North 100m, then west 100m and then RTL.

This has been tested, below are some screen shots of the basic timer get, set and delete functions.
image

Below is a screenshot of a test of a sequence of tasks.  In fact during this test it sadly switch to Drift instead of RTL at the end but that's a different issue that I'm looking into.
image


peterbarker:
I believe this was merged as part of another PR.  Ping me if I'm wrong, please.

PR #1291, opened 2023-12-27
Chat: improve flight mode reliability

rmackay9:
This PR is built on https://github.com/ArduPilot/MAVProxy/pull/1290 and greatly improves the reliability of the chat modules ability to change modes by adding a get_mode_mapping() function that uses MAVProxy's existing list of flight mode names and numbers.

This has been tested in the simulator on Copters and Rovers and so far I have not seen it set the mode incorrectly.

BTW, the previous method relied on the Assistant referring to four text files (one for each vehicle) which included the mapping from flight mode name to number.  Unfortunately OpenAI's assistant seems to have a bug which leads to it sometimes being unable to access the reference files ([see this discussion](https://community.openai.com/t/assistant-not-able-to-access-uploaded-file/524495/14)).  Without the correct mapping the results could be catastrophic (e.g. how do you feel about flying in Drift or AutoTune when you wanted RTL?)


MAVProxy/modules/mavproxy_chat/assistant_setup/get_mode_mapping.json
```diff
@@ -0,0 +1,15 @@
+{
+  "type": "function",
+  "function": {
+    "name": "get_mode_mapping",
```
peterbarker, line 4: I can't help but wonder if it would be nicer to generate all of this from comments in the Python code rather than explicit documentation like this.

Advantages to human programmers in that.  Especially when your editor is smart enough to use those comments to prompt in the user interface.  Additional advantage that this documentation won't get out-of-date with the code, then.

Still, carry on :-)

MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -87,71 +96,74 @@ def set_api_key(self, api_key_str):
...
-        if self.run is None:
-            return "chat: failed to create run"
-
-        # wait for run to complete
-        run_done = False
-        while not run_done:
-            # wait for one second
-            time.sleep(0.1)
+        # get lock
+        with self.send_lock:
```
peterbarker, line 100: You could queue instead of lock - all callers have to handle the async nature of the connection with OpenAI anyway.

MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -327,6 +379,49 @@ def get_vehicle_type(self):
             "vehicle_type": vehicle_type_str
         }
 
+    # get mode mapping
```
peterbarker, line 382: Comment could be better... I exepct the "as a json string" is implied in here somehow.

peterbarker (approved):

MAVProxy/modules/mavproxy_chat/assistant_setup/get_mode_mapping.json
rmackay9: OK, that would be cool.  The json files are slightly difficult to get right the first time you make them but after that it gets easier.  What's odd about them is that you only need to include the list of arguments that the assistant must fill in.  The return values can be whatever you want.. somehow the assistant just deals with whatever it gets.

MAVProxy/modules/mavproxy_chat/chat_openai.py
rmackay9: ok interesting, I'll look into that.

PR #1292, opened 2023-12-28
chat: use vehicle's mavlink sysid

rmackay9:
This changes the receiving and sending of mavlink messages so that we only communicate with the vehicle.

This allows us to ignore traffic from other GCSs which would otherwise confuse the chat module.

This doesn't go far enough to actually allow the control of multiple vehicles but it at least allows us to work in an environment where there are multiple GCSs connected to the vehicle.

Actually after some initial testing this doesn't quite seem to work in all cases.


rmackay9:
This is not quite right so closing.

Issue #1271, opened 2023-11-13
Horizon module crashes with `RegularPolygon.__init__() takes 3 positional arguments but 4 were given`


walchko:
I am running: 

- macOS 
- python 3.11.6 with these modules:
```
(drone)  kevin@Logan ~ % pip list
Package         Version
--------------- -------
billiard        4.1.0
contourpy       1.2.0
cycler          0.12.1
dronekit        2.9.2
dronekit-sitl   3.3.0
fonttools       4.44.0
future          0.18.3
gnureadline     8.1.2
kiwisolver      1.4.5
lxml            4.9.3
matplotlib      3.8.1
MAVProxy        1.8.67
monotonic       1.6
numpy           1.26.1
packaging       23.2
Pillow          10.1.0
pip             23.3.1
psutil          5.9.6
pymavlink       2.4.40
pyparsing       3.1.1
pyserial        3.5
python-dateutil 2.8.2
setuptools      68.2.2
six             1.16.0
wheel           0.41.3
wxPython        4.2.1
``` 

This piece of code seems to crash where `self.headingTri = patches.RegularPolygon()`:

```python
def createHeadingPointer(self):
    '''Creates the pointer for the current heading.'''
    self.headingTri = patches.RegularPolygon((0.0,0.80),3,0.05,color='k',zorder=4)
    self.axes.add_patch(self.headingTri)
    self.headingText = self.axes.text(0.0,0.675,'0',color='k',size=self.fontSize,horizontalalignment='center',verticalalignment='center',zorder=4)
```

Here is the traceback:

```python
(drone)  kevin@Logan ~ % mavproxy.py  --master=udp:localhost:14550 --console  
Connect udp:localhost:14550 source_system=255
Loaded module console
Log Directory: 
Telemetry log: mav.tlog
Waiting for heartbeat from localhost:14550
MAV> 2023-11-12 22:19:42.627 Python[33510:1225035] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to /var/folders/8m/htkq_jv500b_1yqyk33q6tzc0000gn/T/org.python.python.savedState
Detected vehicle 255:0 on link 0
module load horizon
UNKNOWN> UNKNOWN> Loaded module horizon
[...]

PR #1302, opened 2024-01-04
horizon: use keyword arg for radius

srmainwaring:
Fixes: https://github.com/ArduPilot/MAVProxy/issues/1271

The `matplotlib.patches.RegularPolygon` initialiser requires a keyword arg rather than positional arg for the radius:

https://github.com/matplotlib/matplotlib/blob/39582424f8ce5e042e4e39eeac94893f222c9f8b/lib/matplotlib/patches.py#L895-L897

Tested with `matplotlib==3.7.4` for backwards compatibility and `matplotlib==3.8.2`.


peterbarker:
Merged, thanks!

PR #1306, opened 2024-01-12
Chat: relative move improvements and other enhancements

rmackay9:
This includes these enhancements:

1. chat GUI's text input field is moved to the bottom
2. chat GUI's focus is restored back to the text input field after the text is sent to the Assistant
3. takeoff reliability improved by slightly adjusting the assistant's instructions and adding sanity checks to the chat module to ensure the assistant filled in the Z field.  Sometimes the assistant was filling in "param1" instead of "z".
4. set-position-target-global-int function accepts lat and lon in degrees (previously it was integers, degrees * 1e7).  Occasionally the assistant was making calculation errors when converting from degrees to integers (it would drop a digit) leading to the target being many 1000s of km away.
5. The chat module also gets sanity checks the inputs to set-position-target-global-int to ensure if position is used, lat, lon and alt are provided.  In rare cases the assistant was neglecting to send the altitude and we were default to zero alt which would cause a crash.
6. get-location-plus-dist-at-bearing function added to make it easier for the assistant to accept relative movement commands (e.g. "fly vehicle 50m to its right".  Frankly speaking the assistant is still very unreliable but the remaining issues can hopefully be resolved with more adjustments to the assistant's instructions
7. get-parameter-description function added to allow the assistant to get parameter units, valid values, directly from the chat module instead of using [OpenAI's File Retrieval which is unreliable](https://community.openai.com/t/assistant-not-able-to-access-uploaded-file/524495/20).  Despite this the assistant is still not reliably checking the units before setting parameters but hopefully the assistant instructions can be modified to fix this (or maybe AP will move to all SI units!)

I've tested all this fairly thoroughly in SITL.  Below is a screenshot of the parameter description feature.
param-descriptions-test


MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -449,6 +467,17 @@ def send_mavlink_set_position_target_global_int(self, arguments):
         afz = arguments.get("afz", 0)
         yaw = arguments.get("yaw", 0)
         yaw_rate = arguments.get("yaw_rate", 0)
+
+        # sanity check arguments
+        if type_mask == 3576:
+            # if position is specified check lat, lon, alt are provided
+            if "latitude" not in arguments.keys():
```
peterbarker, line 474: This could iterate over a list; duplicated code

peterbarker (approved):

rmackay9:
Thanks very much!

PR #1307, opened 2024-01-12
MAVExplorer: don't show instances with dump tab completion

shancock884:
I very much like PR #1303 from @peterbarker - I have often typed dump and pressed tab to be met with nothing!
However, when using the code from this update, I spotted a couple of things that were not quite as I expected...

- It looks like you meant to link the dump command to new the (MESSAGETYPE) completion that you created, but used (VARIABLE) by mistake, so I changed this.
- The completion includes message instances as well as messages (i.e. both IMU and IMU[1]), but dump only works with the name without instance, so I updated the filter to remove them.

And nothing to do with the previous PR:
- When pressing tab on something like "graph BAT[" it was getting an error, so I updated the partially open array regex to use * instead of + for the text after "[" (i.e. allowed to be empty).
- The regex strings use "\\[" and "\\]", but this generates python warnings, as it tries to apply the escape sequence before passing to regex. Could have either make them raw strings r"...", but instead decided to keep as regular string and escape the "\\".

### DIFF
--- a/MAVProxy/modules/lib/rline.py
+++ b/MAVProxy/modules/lib/rline.py
@@ -189,7 +189,7 @@ def complete_variable(text):
     else:
         suffix = ''
 
-    m1 = re.match("^(.*?)([A-Z0-9][A-Z0-9_]*(\[[0-9A-Z_]+\])?)[.]([A-Za-z0-9_]*)$", text)
+    m1 = re.match("^(.*?)([A-Z0-9][A-Z0-9_]*(\\[[0-9A-Z_]+\\])?)[.]([A-Za-z0-9_]*)$", text)
     if m1 is not None:
         prefix = m1.group(1)
         mtype = m1.group(2)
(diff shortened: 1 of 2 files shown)


peterbarker:
Merged, thanks!

Did find that we can't tab-complete another message name once you have one on the line... 

i.e. `graph XKF[0].TimeUS <tab><tab>` doesn't complete anything

shancock884:
> Did find that we can't tab-complete another message name once you have one on the line...
> 
> i.e. `graph XKF[0].TimeUS <tab><tab>` doesn't complete anything

I think it might be because XKF[0] message may not exist in your log..... should it be something like XKF**1**[0]? At least this was the case in my logs. If I do `graph XKF1[0].TimeUS <tab><tab>` it works.
i.e. if you enter anything that doesn't match a param (like `graph blah <tab><tab>`), then following arguments wont autocomplete.

PR #1309, opened 2024-01-19
MAVExplorer: Fix stats command for DF text and TLOG logs

shancock884:
When looking at something else, I noticed that when using the 'stats' command in MAVExplorer with either DF text logs or TLOGs it raises an Exception for different reasons.
* DF Text logs: `KeyError: 128`
* TLOGs: `ERROR in command []: 'mavmmaplog' object has no attribute 'filehandle'`

It looks like TLOGs were not supported, but should at least have given a nicer error message. But while here, I have added the support for them.

Summary of changes needed:
For DF text logs:
* message name rather than message id is used to index the counts dict, and it is unset for messages that are never seen.
* needs: https://github.com/ArduPilot/pymavlink/pull/906

For TLOG:
* As there isn't a formats attribute on the mavmmaplog class (as used for Tlogs), I am counting message instances rather than size in bytes. Output formatting is adjusted to take account of the longest name.
* The 'filehandle' attribute is named 'f' instead in mavmmaplog class (as used for Tlogs), so updated MPDataLogChildTask class of multiproc_util.py to handle either, using 'hasattr'.

The new output for TLogs looks something like this (abbreviated):
```
Total number of messages: 96576
REMOTE_LOG_BLOCK_STATUS    0.00%
AUTOPILOT_VERSION          0.00%
...
VIBRATION                  3.17%
MISSION_CURRENT            3.17%
HEARTBEAT                  3.50%
NAMED_VALUE_FLOAT          6.25%

@OTHER 100.00%
```

I've tested files of all three types (bin, log, tlog), and checked stats command works in each case.

### DIFF
--- a/MAVProxy/modules/lib/msgstats.py
+++ b/MAVProxy/modules/lib/msgstats.py
@@ -42,31 +42,52 @@ class MPMsgStats(MPDataLogChildTask):
 
 def show_stats(mlog):
     '''show stats on a file'''
-    if not hasattr(mlog, 'formats'):
-        print("Must be DF log")
-        return
     sizes = {}
     total_size = 0
     names = mlog.name_to_id.keys()
     pairs = []
+    maxnamelen = 4
 
     for name in names:
         sizes[name] = 0
 
     for name in names:
         mid = mlog.name_to_id[name]
-        count = mlog.counts[mid]
-        mlen = mlog.formats[mid].len
-        size = count * mlen
-        total_size += size
-        sizes[name] += size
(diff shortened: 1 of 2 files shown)


peterbarker (approved):
Would be nice to get tlog usages in bytes - useful for fixing link bandwidth issues.

We might just be able to do a `len` on the byte-content of the message which we stash away in the Python ovject somewhere, I think.  OTOH, it might be zero-pad-extended or it might not have timestamps, so delving reuqured!

Anyway, thanks, this has bugged me for a while, merging it!

Issue #1131, opened 2022-12-13
NTRIP client : Is this dependent on a specific GNSS manufacturer?


joohyuk95:
I am using NTRIP client code [(ntrip.py)](https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/lib/ntrip.py) to use VRS-RTK.

I live in Korea. There are two NTRIP servers in Korea. 
One is the solution of the US company **Trimble** and the other is the German company **Geo++** solution.

A = **Trimble**: PIVOT
B = **Geo++** : GNSMART2

---
The issue is as follow.

If I use VRS-RTK through server B, it works without any problem.
But if I use server A, I can connect to the server, but I couldn't receive the data.
I checked if it was a server problem, but it wasn't.

To solve the problem, I first changed the NMEA_GGA message generated by the **ntrip.py** internal function 'getGGAstring' to the message generated by [NMEA generator](https://www.nmeagen.org/) and sent it to the server.
Then I could receive the data from the server A.
However, this data was not reliable. Rather, it only worsened the accuracy of standalone-gps.
This means that data is being received, but something is wrong.
I checked again if there was a problem with the server, but it was not.
And using this method, the opposite happens. (I can receive data from server A, but I cannot receive data from server B.)

The NMEA_GGA message formats are as follows.
I think the number of digits is different.
* getGGAstring func :  GPGGA,060644.00,5005.40000000,N,00839.60000000,E,1,05,0.19,+00400,M,1200.000,M,,
* NMEA generator  :  GPGGA,045338.171,3553.287,N,12836.319,E,1,12,1.0 ,0.0,M,0.0,M,,*68

In summary, if I use the ntrip.py code as it is, I can use VRS-RTK through server B, but not server A.
I don't know why this phenomenon occurs when the same protocol is used. Is the method of data parsing on the server different because the manufacturer is different? So is the ntrip.py code dependent on a specific manufacturer?

I earnestly ask for your help.

richahert:
I have the same problem with topcon rtk corretion services. It works fine for SAPOS RTK2GO and many others.

One interesting thing I noticed is that [pygpsclient](https://github.com/semuconsulting/PyGPSClient) only worked with topcon in a newer Version because pyubx2 threw something like a `KeyError: 41 ` in the old version. Havent't understood what the problem is so far

PR #1323, opened 2024-02-15
NTRIP: use pynmeagps for correct gga generation

richahert:
The MAVproxy NTRIP client does not work with several NTRIP casters, because the GGA messages are sent in an incompatible format. this is described in #1131 

This PR solves that problem, by using the same library for gga generation that [pygpsclient](https://github.com/semuconsulting/PyGPSClient) uses ([pynmeagps](https://github.com/semuconsulting/pynmeagps)), which I experienced to be working with all casters. 

The ntrip client of pygpsclient works quite well, so I might open an Issue in the future to request that the MAVproxy ntrip could be refactored as a wraper of that NTRIP client. It has more active developers than the built-in NTRIP Client of MAVproxy.

For now, this PR should provide a solution that is tested with skylark and I am confident that it also solves the problem with topcon, but I can't test it atm, because I don't have an active subscription.

### DIFF
--- a/MAVProxy/modules/lib/ntrip.py
+++ b/MAVProxy/modules/lib/ntrip.py
@@ -14,8 +14,9 @@ import select
 from MAVProxy.modules.lib import rtcm3
 import ssl
 from optparse import OptionParser
+from pynmeagps import NMEAMessage, GET
 
-version = 0.1
+version = 0.2
 useragent = "NTRIP MAVProxy/%.1f" % version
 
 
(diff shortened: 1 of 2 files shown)


rmackay9:
Hi @richahert,

Could you rebase this on master to remove the merge commits?

richahert:
Hi @rmackay9 ,
I rebased the branch so the PR has one clean commit.

rmackay9:
Hi @richahert,

Thanks for that.  You might want to also prefix the commit message with "ntrip:" or "NTRIP:" to be consistent with other commits in this repo but again, I'm not a maintainer so this is just a suggestion and I'm not sure how strict Tridge and others are.

If this gets stuck and not merged then it might need to be brought up on one of the dev calls to get merged.  The best choice is probably the small EU Dev Call which is later today.

tridge (approved):

PR #1325, opened 2024-02-25
MAVExplorer: Autocompletion on dump --verbose command

shancock884:
This PR allows autocompletion in MAVExplorer when using "dump --verbose \<tab\>".
Previously the autocompletion was added for "dump \<tab\>", but with "--verbose" argument it did not work.
Tested on my local version.


peterbarker (approved):
I'm pretty sure this isn't the right way to do it, but good for now :-)

PR #1332, opened 2024-03-06
Mission Editor-extras

AndKe:
Adding these buttons: 

"Add Yaw for each WP":
 This add (optionally, starting from with selected waypoint) a YAW command for each waypoint. 
This is useful for copters flying a grid/mission with a preset heading (in order for a sensor to face it's target )

"Deleye Yaw" : 
Clean up mission, removing all Yaw commands.

"Change WP Altitude": 
Allows to enter a offset for all WP altitudes in mission, useful when changing the mission slightly up/down

"Change Frame":
Let the user quickly change the whole missions altitude frame - for example from REL to ABS or AGL.

### DIFF
--- a/MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
+++ b/MAVProxy/modules/mavproxy_misseditor/missionEditorFrame.py
@@ -202,6 +202,11 @@ class MissionEditorFrame(wx.Frame):
         self.grid_mission = wx.grid.Grid(self, wx.ID_ANY, size=(1, 1))
         self.button_add_wp = wx.Button(self, wx.ID_ANY, "Add Below")
         self.button_split = wx.Button(self, wx.ID_ANY, "Split")
+        self.button_add_yaw_foreach = wx.Button(self, wx.ID_ANY, "Add Yaw for each wp")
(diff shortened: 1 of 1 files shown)


peterbarker:
image

tridge recently added some commands to change the Frame.

We'll probably add something to change a arbitrary field somehow.

One problem with this PR as-written is that it will affect all waypoints rather than a range.  At least on Linux you can select a range of rows.  I think it would be a far better interface to be able to select a range and *only* affect those cells.

image

tridge will need to be happy with the extra screen real-estate taken up by the new buttons.

AndKe:
Yes, the "from" behavior is by design, as one often want to keep flying nose-first to the data gathering area, and then fly sideways.  
You mention selection is an option in Linux, - yes, but I have no clue is that has some reason to be different in win/mac . 
Of course, I could have made it "for selected waypoints"

The real-estate has been vacant for years, and it is time to use it for something useful :)

AndKe:
@tridge The current way it works is (as described) all, or - from selected (if a row is selected)  This is working fine for us (who are the only users of this that I am aware of)  :) 
Would you say changing the way it works to "selected cells" is required? Or is that just a slight preference you have?
Thank you.

PR #1334, opened 2024-03-06
module Joystick: adding Sparkfun pro micro as joystick

AndKe:
"Sparkfun pro micro" is a cheap 32u4 microcontroller that using Arduino joystick library is easy to act as a USB joystick.
It is easy to use for PPM input, or analog joystick interface.


peterbarker (approved):
Commit message should be "joystick: " or something

PR #1335, opened 2024-03-07
Show per-cell voltage when number of cells is given.

MichelleRos:
Shows the per-cell voltage when the user has set the number of battery cells.

Not sure if maybe it should still show the full battery voltage when this is set?
Maybe an option?

Tested in SITL but not with a real vehicle (yet).

### DIFF
--- a/MAVProxy/modules/mavproxy_battery.py
+++ b/MAVProxy/modules/mavproxy_battery.py
@@ -69,7 +69,19 @@ class BatteryModule(mp_module.MPModule):
             if batt_mon == 0:
                 continue
 
-            #report voltage level only
+            # show per-cell voltage if number of cells is set
+            if self.numcells(id) != 0:
+                #this is a workaround for the race condition since self.per_cell only gets filled in once the setting is set
+                if not id in self.per_cell:
+                    continue
+                if batt_mon == 3:
+                    battery_string += 'Batt%u: %.2fV*%uS ' % (id+1, self.per_cell[id], self.numcells(id))
+                    continue
+                if batt_mon >= 4:
+                    battery_string += 'Batt%u: %u%%/%.2fV*%uS %.1fA ' % (id+1, self.battery_level[id], self.per_cell[id], self.numcells(id), self.current_battery[id])
+                    continue
+
+            #show full battery voltage when number of cells is not set
             if batt_mon == 3:
                 battery_string += 'Batt%u: %.2fV ' % (id+1, self.voltage_level[id])
             elif batt_mon >= 4:


MAVProxy/modules/mavproxy_battery.py
```diff
@@ -69,14 +69,26 @@ def battery_report(self):
...
-            #report voltage level only
-            if batt_mon == 3:
-                battery_string += 'Batt%u: %.2fV ' % (id+1, self.voltage_level[id])
-            elif batt_mon >= 4:
-                battery_string += 'Batt%u: %u%%/%.2fV %.1fA ' % (id+1,
-                                                                   self.battery_level[id],
-                                                                   self.voltage_level[id],
-                                                                   self.current_battery[id])
+            #show full battery voltage when number of cells is not set
+            if self.numcells(id) == 0:
```
peterbarker, line 73: This patch can be simplified by continuing the `continue` pattern that's already in here.  So you would add a new block looking something like this, about line 72:

```
          # show per-cell voltage if number of cells is set:
          if self.numcells(id) != 0:
               #this is a workaround for the race condition since self.per_cell only gets filled in once the setting is set
                if not id in self.per_cell:
                    continue
                if batt_mon == 3:
                    battery_string += 'Batt%u: %.2fV*%ucells ' % (id+1, self.per_cell[id], self.numcells(id))
                    continue
                if batt_mon >= 4:
                    battery_string += 'Batt%u: %u%%/%.2fV*%ucells %.1fA ' % (id+1, self.battery_level[id], self.per_cell[id], self.numcells(id), self.current_battery[id])
                continue
```

The chief advantage of this is that your patch is simply added lines, not changing existing lines.

tridge:
@MichelleRos update on this?

MAVProxy/modules/mavproxy_battery.py
MichelleRos: Done.

MichelleRos:
Updated to incorporate @peterbarker's suggestion.

MichelleRos:
@tridge Could you have another look at it now that I've updated it?

MichelleRos:
@tridge @peterbarker Any thoughts on this?

Do we want the full battery voltage to still be displayed somewhere when num_cells is set? Maybe when the `bat` command is used?

MAVProxy/modules/mavproxy_battery.py
```diff
@@ -69,7 +69,19 @@ def battery_report(self):
             if batt_mon == 0:
                 continue
 
-            #report voltage level only
+            # show per-cell voltage if number of cells is set
+            if self.numcells(id) != 0:
+                #this is a workaround for the race condition since self.per_cell only gets filled in once the setting is set
+                if not id in self.per_cell:
```
peterbarker, line 75:
```suggestion
                if id not in self.per_cell:
```

peterbarker:
> @tridge @peterbarker Any thoughts on this?
> 
> Do we want the full battery voltage to still be displayed somewhere when num_cells is set? Maybe when the `bat` command is used?

That'd be good.

peterbarker:
> > @tridge @peterbarker Any thoughts on this?
> > Do we want the full battery voltage to still be displayed somewhere when num_cells is set? Maybe when the `bat` command is used?
> 
> That'd be good.



> @tridge @peterbarker Any thoughts on this?
> 
> Do we want the full battery voltage to still be displayed somewhere when num_cells is set? Maybe when the `bat` command is used?

It would also be nice to show the values of each cell under `bat` - that can actually mean something on something like the Angel with its per-cell monitor!

peterbarker (approved):
This is now very easy to understand.

The only thing that might be warranted is a MAVProxy setting to turn the behaviour on/off - but *I* don't think it is necessary.

tridge (approved):

PR #1337, opened 2024-03-10
MAVExplorer: Add logmessage download and help commands, using meta-data from autotest web site

shancock884:
The aim of this PR is to start making use of the LogMessage xml files in MAVExplorer.

This PR adds the commands "logmessage download" and "logmessage help <msg>" to MAVExplorer.
The download command will download all LogMessage XMLs to the user's .mavproxy folder.
The help command calls  mlog.metadata.print_help(msg) method on the new DFMetaData class (links to PR https://github.com/ArduPilot/pymavlink/pull/929).

The DFMetaData class is implemented in DFReader.py file, within pymavlink, so that other tools, such as mavlogdump can access the meta data on a loaded DF file without a dependency on MAVProxy. For example, to provide the enum strings and bitmask information when doing "mavlogdump --verbose".
Any call to mlog.metadata is protected with hasattr(mlog,'metadata') in case mlog is a TLOG or if the the pymavlink in use is older, and does not yet have the attribute.
I also updated load_graphs function to avoid XML warnings when it tries to load *.xml from .mavproxy folder.

An example of the output is as follows:
```
MAV> logmessage help ORGN
Log Message: ORGN
Vehicle navigation origin or other notable position

TimeUS [μs] : Time since system startup
Type [instance] : Position type
Lat [deglatitude] : Position latitude
Lng [deglongitude] : Position longitude
Alt [m] : Position altitude
```

### DIFF
--- a/MAVProxy/tools/MAVExplorer.py
+++ b/MAVProxy/tools/MAVExplorer.py
@@ -30,6 +30,7 @@ from MAVProxy.modules.lib.mp_menu import *
 import MAVProxy.modules.lib.mp_util as mp_util
 from pymavlink import mavutil
 from pymavlink import mavwp
+from pymavlink import DFReader
 from MAVProxy.modules.lib.mp_settings import MPSettings, MPSetting
 from MAVProxy.modules.lib import wxsettings
 from MAVProxy.modules.lib.graphdefinition import GraphDefinition
@@ -128,6 +129,7 @@ class MEState(object):
             "dump"      : ['(MESSAGETYPE)', '--verbose (MESSAGETYPE)'],
             "map"       : ['(VARIABLE) (VARIABLE) (VARIABLE) (VARIABLE) (VARIABLE)'],
             "param"     : ['download', 'check', 'help (PARAMETER)'],
+            "logmessage": ['download', 'help (MESSAGETYPE)'],
             }
         self.aliases = {}
         self.graphs = []
@@ -332,6 +334,9 @@ def load_graphs():
     mestate.graphs = []
     gfiles = ['mavgraphs.xml']
     for dirname, dirnames, filenames in os.walk(mp_util.dot_mavproxy()):
+        # Skip XML files in the LogMessages subfolder
+        if os.path.basename(dirname) == "LogMessages":
+            continue
         for filename in filenames:
             if filename.lower().endswith('.xml'):
                 gfiles.append(os.path.join(dirname, filename))
@@ -1176,6 +1181,44 @@ def cmd_paramchange(args):
         vmap[pname] = pvalue
     mestate.mlog.rewind()
 
+
+def cmd_logmessage(args):
+    '''show log message information'''
+    mlog = mestate.mlog
+    usage = "Usage: logmessage <help|download>"
+    # Print usage and return, if we have no arguments
+    if len(args) <= 0:
+        print(usage)
+        return
+    # help: print help for the requested log message
+    if args[0] == 'help':
+        if len(args) < 2:
+            print(usage)
+            return
+        if hasattr(mlog, 'metadata'):
+            mlog.metadata.print_help(args[1])
+        elif isinstance(mlog, mavutil.mavlogfile):
+            print("logmessage help is not supported for telemetry log files")
+        else:
+            print("Incompatible pymavlink; upgrade pymavlink?")
+        return
+    # download: download XML files for log messages
+    if args[0] == 'download':
+        if not hasattr(DFReader, 'DFMetaData'):
+            print("Incompatible pymavlink; upgrade pymavlink?")
+            return
+        try:
+            child = multiproc.Process(target=DFReader.DFMetaData.download)
+            child.start()
+        except Exception as e:
+            print(e)
+        if hasattr(mlog, 'metadata'):
+            mlog.metadata.reset()
+        return
+    # Print usage if we've dropped through the ifs
+    print(usage)
+
+
 def cmd_mission(args):
     '''show mission'''
     if (len(args) == 1):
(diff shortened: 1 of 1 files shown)


MAVProxy/tools/MAVExplorer.py
```diff
@@ -336,12 +337,16 @@ def load_graphs():
...
     for file in gfiles:
         if not os.path.exists(file):
             continue
-        # skip parameter files.  They specify an encoding, and under
-        # Python3 this leads to a warning from etree
-        if os.path.basename(file) in ["ArduSub.xml", "ArduPlane.xml", "APMrover2.xml", "ArduCopter.xml", "AntennaTracker.xml", "Blimp.xml", "Rover.xml"]:
+        # skip parameter and log message files.  They specify an encoding,
+        # and under Python3 this leads to a warning from etree
+        basename = os.path.basename(file)
+        if (basename in paramfiles) or basename.startswith("LogMessages_"):
```
peterbarker, line 349: Perhaps we should create a subdirectory instead of extending the misery?

Possibly mirroring the `LogMessages` name we use on the autotest server?

shancock884: Sure. Sounds sensible to use a sub-folder and keep the name instead of renaming to LogMessage_<name>.xml.
In that case should probably use a subfolder for param XMLs too (separate PR).
This would make filtering them out easier, as it can be done but folder, instead of individual file name.

peterbarker:
On Wed, 13 Mar 2024, Simon Hancock wrote:

> In that case should probably use a subfolder for param XMLs too (separate PR).

Indeed.

MAVProxy/tools/MAVExplorer.py
shancock884: Changed to use LogMessages subfolder as proposed.

MAVProxy/tools/MAVExplorer.py
```diff
@@ -332,16 +333,22 @@ def load_graphs():
...
 
+    # list of parameter files to be skipped when loading graph XML
+    paramfiles = ["ArduSub.xml", "ArduPlane.xml", "APMrover2.xml", "ArduCopter.xml", "AntennaTracker.xml", "Blimp.xml", "Rover.xml"]
+    # loop though files
     for file in gfiles:
         if not os.path.exists(file):
             continue
         # skip parameter files.  They specify an encoding, and under
         # Python3 this leads to a warning from etree
-        if os.path.basename(file) in ["ArduSub.xml", "ArduPlane.xml", "APMrover2.xml", "ArduCopter.xml", "AntennaTracker.xml", "Blimp.xml", "Rover.xml"]:
```
peterbarker, line 344: This patch is no longer required.  It could also be a `frozenset` in place so keeping the thing tightly scoped while increasing efficiency - but really, just remove this bit :-)

MAVProxy/tools/MAVExplorer.py
```diff
@@ -1176,6 +1183,38 @@ def cmd_paramchange(args):
         vmap[pname] = pvalue
     mestate.mlog.rewind()
 
+def cmd_logmessage(args):
+    '''show log message information'''
+    mlog = mestate.mlog
+    usage = "Usage: logmessage <help|download>"
+    if len(args) > 0:
```
peterbarker, line 1190:
```suggestion
    if len(args) <= 0:
        print(usage)
        return

```

... save indenting / context complication

.... `logmessage list` to show fields which are understood, perhaps?  Or are in the intersection of what help can provide and what's in the log?  ("future enhancement" :-) )

MAVProxy/tools/MAVExplorer.py
```diff
@@ -1176,6 +1183,38 @@ def cmd_paramchange(args):
...
+        if args[0] == 'help':
+            if len(args) < 2:
+                print(usage)
+                return
+            if hasattr(mlog,'metadata'):
+                mlog.metadata.print_help(args[1])
```
peterbarker, line 1196:
```suggestion
            if hasattr(mlog,'metadata'):
                mlog.metadata.print_help(args[1])
            else:
                print("Incompatible pymavlink; upgrade pymavlink?")
```

MAVProxy/tools/MAVExplorer.py
```diff
@@ -1176,6 +1183,38 @@ def cmd_paramchange(args):
...
+            return
+        if args[0] == 'download':
+            # download XML files for log messages
```
peterbarker, line 1199: Length-wise, this is on the border of making a new separate `cmd_logmessage_download` a good idea.

peterbarker (changes requested):
Very cool.

MAVProxy/tools/MAVExplorer.py
shancock884: Definately good idea!... I think I will need to also check if its a DF or Telem log, and do a different message, e.g. "Log message help is not supported for Telemetry logs".

MAVProxy/tools/MAVExplorer.py
shancock884: removed change.

MAVProxy/tools/MAVExplorer.py
shancock884: updated including additoinal check for `isinstance(mlog, mavutil.mavlogfile)`

MAVProxy/tools/MAVExplorer.py
shancock884: Now that download is now implemented on pymavlink side, this code is shorter, so I maybe length is now OK to keep here?

MAVProxy/tools/MAVExplorer.py
```diff
@@ -332,6 +334,9 @@ def load_graphs():
     mestate.graphs = []
     gfiles = ['mavgraphs.xml']
     for dirname, dirnames, filenames in os.walk(mp_util.dot_mavproxy()):
+        # Skip XML files in the LogMessages subfolder
+        if os.path.basename(dirname) == "LogMessages":
+            continue
```
peterbarker, line 339: Don't need this bit any more.

peterbarker (approved):
LGTM.

Tested it.

Merging, thanks!

PR #1341, opened 2024-03-21
Audio recording enhacement

nishant-sg:
Before:
The user could only record for a fix 5 seconds

After:
As long as the user is talking everything will be recorded. When the user stops talking it will take a three seconds to stop recording. 

Print statements have been removed, for demo purpose please look at the video

https://github.com/ArduPilot/MAVProxy/assets/70840360/66e6ed42-117b-4a0e-882d-eef84e6a5316

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
@@ -85,10 +85,6 @@ class chat_voice_to_text():
             else:
                 isSpeaking = False
             curr_time = time.time()
-            print(time_stop-curr_time)
-            if isSpeaking:
-                print("speaking")
-        print("recording ended")
         # Stop and close the stream
         stream.stop_stream()
         stream.close()
--- a/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
@@ -6,6 +6,8 @@ AP_FLAKE8_CLEAN
 '''
 
 import time
+import math
+import struct
 
 try:
     import pyaudio  # install using, "sudo apt-get install python3-pyaudio"
@@ -15,6 +17,15 @@ except Exception:
     print("chat: failed to import pyaudio, wave or openai.  See https://ardupilot.org/mavproxy/docs/modules/chat.html")
     exit()
 
+def rms( data ):
+    count = len(data)/2
+    format = "%dh"%(count)
+    shorts = struct.unpack( format, data )
+    sum_squares = 0.0
+    for sample in shorts:
+        n = sample * (1.0/32768)
+        sum_squares += n*n
+    return math.sqrt( sum_squares / count )
 
 class chat_voice_to_text():
     def __init__(self):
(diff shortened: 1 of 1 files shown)


adityaomar3:
@nishant-sg I have tried hard coding the  decibel or energy, but it wont work on every system, as system to sytem tuning varies, also while calculating the decibel, the base power has to be included which is again system dependent, I would prefer working on energy values.

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
```diff
@@ -15,6 +17,15 @@
     print("chat: failed to import pyaudio, wave or openai.  See https://ardupilot.org/mavproxy/docs/modules/chat.html")
     exit()
 
+def rms( data ):
```
rmackay9, line 20: Hi, thanks for this.   A few things to fix here:

1. could you add comments above the function?
2. I think we should name the function to be more specific so its purpose is clear.  Perhaps calc_audio_volume() and maybe it should return the decibels directly instead of making the caller do that.
3. our normal style is no spaces after brackets.  So change "def rms( data )" to "def rms(data)"
4. I think it's possible that count could be zero if the microphone is not working.  In any case we should have protection against divide-by-zero which could happen if count is zero.

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
```diff
@@ -55,15 +66,25 @@ def record_audio(self):
...
         curr_time = time.time()
-        time_stop = curr_time + 5
+        time_stop = curr_time + 3
 
         # record until specified time
         frames = []
-        while curr_time < time_stop:
+
+        # logic for recording sound until someone is speaking.
+        isSpeaking = True
```
rmackay9, line 75: Hi, I think our normal style is to use underscores between variables so let's change "isSpeaking" to "is_speaking".

Also maybe change the comment to be "record sound while user is speaking"

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
```diff
@@ -55,15 +66,25 @@ def record_audio(self):
...
+        while curr_time < time_stop or isSpeaking:
             data = stream.read(1024)
             frames.append(data)
+            rms1 = rms(data)
+            if rms1!=0.0:
+                decibel = 20 * math.log10(rms1)
+                isSpeaking = decibel>-80.0  # -80 is the hardcoded threshold. higher number means louder. Set threshold in the range (-100,0)
```
rmackay9, line 82: If possible let's move this -80 to be a definition at the top of the file where it will be easier to find.

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
```diff
@@ -34,7 +45,7 @@ def check_connection(self):
             try:
                 self.client = OpenAI()
             except Exception:
-                print("chat: failed to connect to OpenAI")
+                print("chat: failed to connect to OpenAI - 4")
```
rmackay9, line 48: I think this change to the print statement can be removed.

PR #1344, opened 2024-03-25
Restoring old chats in chat module

adityaomar3:
Added function to restore the old messages in chat module on submitting the correct assistant api key.
Currently it is in testing phase, we have to add thread_id manually, otherwise it will start new thread.

The demonstration is attached below.

https://github.com/ArduPilot/MAVProxy/assets/106031961/2f63c47f-154e-4b4c-88cf-ec720af10a66


adityaomar3:
Next steps is to add a function to automatically store the thread id automatically in the system and retrieve it next time onwards

rmackay9:
This looks pretty good to me.  As you say we need a way to either store the most recent chat in some persistent settings file or we need to be able to display a list to the user of the most recent thread and pick from there... or if we know which thread is the most recent we could just start using that one.  Maybe it would be best to add a new menu item to "Continue from latest thread".. of course things could get interesting if one user has more than a single chat window open but I think that's OK.

I've discussed with @peterbarker recently about whether there is a way to store settings but I don't think MAVProxy has this ability at the moment.

rmackay9:
By the way, once we are closer to merging you might want to squash those two commits into a single commit.

peterbarker:
Ping @adityaomar3 - are you still interested in pursuing this one?

adityaomar3:
Hi @peterbarker, Yes! I will be working on this feature, currently I am working on some other features in chat module itself but this is on my list, I will surely address this but a bit later.

PR #1345, opened 2024-03-25
Chat: added password input field for api_key

adityaomar3:
Before input key field for api key was exposed
key_input_field_before

After changes input key field for api key 

key input field after

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/chat_window.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_window.py
@@ -38,7 +38,8 @@ class chat_window():
 
         # add api key input window
         self.apikey_frame = wx.Frame(None, title="Input OpenAI API Key", size=(560, 50))
-        self.apikey_text_input = wx.TextCtrl(self.apikey_frame, id=-1, pos=(10, 10), size=(450, -1), style=wx.TE_PROCESS_ENTER)
+        self.apikey_text_input = wx.TextCtrl(self.apikey_frame, id=-1, pos=(10, 10), size=(450, -1),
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @adityaomar3,

Thanks very much for this.  It's a small thing but I wonder if you could appeand, "chat: " onto the front of the commit title?  It's not big deal but if you look at our commit list you'll see we normally append the subsystem name or directory name to each commit to make it more clear what it affects.  To be clear, I mean the commit title not the PR title.

This is a habit the AP devs have gotten into especially on the flight code side because it makes backporting to earlier versions easier.  It's not really very important for MAVProxy but we do it anyway.  Txs!

After that change I'll merge.

rmackay9:
I've marked this for the EU dev call, that's just so that we don't forget to merge it.

rmackay9:
Hi @adityaomar3,

By the way, is there any change in behaviour if Enter is pressed instead of pressing the "Set" button?

I forgot to bring this up on the EUDevCall but I'll ping some of the other devs in the morning and we will merge it then.

adityaomar3:
> Hi @adityaomar3,
> 
> By the way, is there any change in behaviour if Enter is pressed instead of pressing the "Set" button?
> 
> I forgot to bring this up on the EUDevCall but I'll ping some of the other devs in the morning and we will merge it then.

Thanks for the review, I checked, it actually broke the enter key feature.

adityaomar3:
A small change was required, now it is working absolutely fine.

rmackay9:
Merged, thanks!

PR #1353, opened 2024-04-09
MAVProxy support retries from cmd line

joshanne:
In some debugging applications, I use more retries than the default 3 that are given.

This exposes retries to the command line, allowing to increase how many attempts before mavproxy gives up.


peterbarker (approved):
Not sure if `--retries` on the command-line is specific enough.

PR #1357, opened 2024-04-12
misseditor: correct display of distance/gradient

peterbarker:
somehow this changed to a string... so just cope with that


MAVProxy/modules/mavproxy_mode.py
```diff
@@ -119,6 +127,61 @@ def cmd_guided(self, args):
             altitude
         )
 
+    def build_pt_ignoremask(self, bits, force_not_accel=False):
```
tridge, line 130: leftover commits?

tridge (changes requested):

tridge (approved):

PR #1367, opened 2024-04-24
wp: allow mission creation without home

tridge:
allows user to create missions before GPS lock and home


tridge:
peter likes it, merging

PR #1392, opened 2024-05-29
CI: Fixed Windows release building

stephendade:
Fixes the issue of stable releases not being generated. Using a different GitHub action appears to fix it.


stephendade:
> Ah, is this related to the missing Windows installer for the recent releases? I suspect it is, thanks!

Assuming it works, then yes :)

stephendade:
Has been tested on my repo (https://github.com/stephendade/MAVProxy/tree/master) and works

PR #1395, opened 2024-05-30
Add option --not_reset_streamrates

python36:
MAV_CMD_SET_MESSAGE_INTERVAL commands are meaningless since the periodic function set_stream_rates constantly resets the intervals.
The messagerate module also does not work because request_data_stream_send resets all intervals.
REQUEST_DATA_STREAM was deprecated in 2015-08 and the code needs to get rid of it.
But as a temporary solution I added option --not_reset_streamrates.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -999,6 +999,9 @@ def open_telemetry_logs(logpath_telem, logpath_telem_raw):
 
 def set_stream_rates():
     '''set mavlink stream rates'''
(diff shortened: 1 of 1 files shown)


peterbarker:
Is this not equivalent to `mavproxy.py --streamrate=-1` ?

peterbarker:
Pretty sure this is equivalent; re-open if you think otherwise.

python36:
@peterbarker many thanks 🙏 ❤️

PR #1400, opened 2024-06-12
Build: Fix CI - pin pyinstaller and update imports

stephendade:
Turns out the culprit was pyinstaller 6.8.0 breaking the build. No idea why.

Also took the opportunity to clean up a bunch of build warnings.


stephendade:
@peterbarker - should be good to merge.

peterbarker:
Thanks @stephendade :-)

PR #1404, opened 2024-06-15
Chat: feature to cancel active runs

adityaomar3:
This PR adds a cancel feature to cancel the active run on Open AI, i.e. it helps cancelling the active prompt given by user.

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/chat_openai.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_openai.py
@@ -48,6 +48,7 @@ class chat_openai():
         self.client = None
         self.assistant = None
         self.assistant_thread = None
+        self.latest_run = None
 
     # check connection to OpenAI assistant and connect if necessary
     # returns True if connection is good, False if not
@@ -99,6 +100,25 @@ class chat_openai():
         self.assistant = None
         self.assistant_thread = None
 
+    # cancel the active run
+    def cancel_run(self):
+        # check the active thread and run
+        if (self.assistant_thread and self.latest_run is not None):
+            run_status = self.latest_run.status
+            if (run_status != "completed" and run_status != "cancelled" and
+                    run_status != "cancelling"):
+
+                # cancel the active run
+                self.client.beta.threads.runs.cancel(
+                    thread_id=self.assistant_thread.id,
+                    run_id=self.run.id
+                )
+            else:
+                if (self.latest_run.status == "completed"):
+                    print("Chat is completed, cannot be cancelled")
+                elif (self.latest_run.status == "cancelled"):
+                    print("Chat is cancelled")
+
     # send text to assistant
     def send_to_assistant(self, text):
         # get lock
@@ -132,7 +152,7 @@ class chat_openai():
                 time.sleep(0.1)
 
                 # retrieve the run
-                latest_run = self.client.beta.threads.runs.retrieve(
+                self.latest_run = self.client.beta.threads.runs.retrieve(
                     thread_id=self.assistant_thread.id,
                     run_id=self.run.id
                 )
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -99,6 +100,23 @@ def set_api_key(self, api_key_str):
         self.assistant = None
         self.assistant_thread = None
 
+    # cancel the active run
+    def cancel_run(self):
+        # check the active thread and run
+        if (self.assistant_thread and self.latest_run is not None):
+            if (self.latest_run.status != "completed" and "cancelled"):
```
rmackay9, line 107: Oops, I think my comment was posted somewhere else.  In any case, I think the 2nd argument looks incorrect in this "if" statement.

MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -99,6 +100,23 @@ def set_api_key(self, api_key_str):
...
+
+                # cancel the active run
+                self.client.beta.threads.runs.cancel(
+                    thread_id=self.assistant_thread.id,
+                    run_id=self.run.id
+                )
+            else:
+                if (self.lastest_run.status == "completed"):
```
rmackay9, line 115: little typo here, "lastest_run" should be "latest_run"

MAVProxy/modules/mavproxy_chat/chat_openai.py
rmackay9, line 107: we also need to handle the "cancelling" state

rmackay9:
Thanks for this, I've tested and found a few little issues but once you fix those up (and re-test yourself) then I'll re-test and get this merged in

MAVProxy/modules/mavproxy_chat/chat_openai.py
adityaomar3: Oops!

MAVProxy/modules/mavproxy_chat/chat_openai.py
adityaomar3: Sure! I got it.

adityaomar3:
HI! @rmackay9 take a look, I have updated the fixes as told, I tested it, You please take a test too, if there is any edge cases missing or changes needed.

MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -165,7 +185,10 @@ def send_to_assistant(self, text):
             reply_messages = self.client.beta.threads.messages.list(self.assistant_thread.id,
                                                                     order="asc",
                                                                     after=input_message.id)
-            if reply_messages is None:
+
+            if (self.latest_run.status == "cancelled"):
+                return "cancelled successfully"
+            elif reply_messages is None:
```
rmackay9, line 191: Not a blocker but the "elif" could be just an "if" because of the return one line higher

rmackay9 (approved):

rmackay9:
Thanks for this.  By the way, for future PRs could you specify what testing has been done?  This is helpful for the reviewer.

I tested this and it works.  I saw one bit of odd behaviour (see below) in which I cancelled a command, "arm and takeoff to 20m" and then successfully cancelled the command but then when I put in an innocuous command a moment later ("helo" <-- was meant to be "hello") it completed the cancelled command.  I'm pretty sure this is an issue on the OpenAI side though and it didn't happen the next time.  Let's just keep an eye on this potential issue.
image

adityaomar3:
Sure @rmackay9 thanks for the review, next time onwards I will surely add testing cases, reports from my side for a better review and tests.

Sometimes it does not cancel, maybe the reason is code latency, for some prompts(mostly in which function calling is not used) responses are generated by OpenAI very fast, while status being "in progress" when the code reaches cancel part it is already completed.

MAVProxy/modules/mavproxy_chat/chat_openai.py
```diff
@@ -99,6 +100,25 @@ def set_api_key(self, api_key_str):
         self.assistant = None
         self.assistant_thread = None
 
+    # cancel the active run
+    def cancel_run(self):
+        # check the active thread and run
+        if (self.assistant_thread and self.latest_run is not None):
+            run_status = self.latest_run.status
+            if (run_status != "completed" and run_status != "cancelled" and
+                    run_status != "cancelling"):
```
peterbarker, line 109:
```suggestion
            if run_status not in ["completed", "cancelled", "cancelling"]:
```

peterbarker:
Merged, thanks!

PR #1412, opened 2024-06-26
chat: enabled chat streaming feature

adityaomar3:
This PR adds the streaming feature by Open-AI assistants API.
Previous: Chats were concatenated together and displayed on screen, the wait time for replies in that approach was too long.
Streaming enables the communication of server sent events, which reduces the latency and wait time.
Below is a sample test-video displaying the working of stream.

https://github.com/ArduPilot/MAVProxy/assets/106031961/aabc933c-73a0-448a-8c85-4c6566babca5

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/chat_openai.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_openai.py
@@ -14,24 +14,54 @@ import time
 import re
 from datetime import datetime
 from threading import Thread, Lock
+from typing_extensions import override
 import json
 import math
 from MAVProxy.modules.lib import param_help
 
 try:
-    from openai import OpenAI
+    from openai import OpenAI, AssistantEventHandler
 except Exception:
     print("chat: failed to import openai. See https://ardupilot.org/mavproxy/docs/modules/chat.html")
     exit()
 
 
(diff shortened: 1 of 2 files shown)


rmackay9 (approved):
Thanks for this.  This is the feature that I've most been looking forward to adding to the chat module!

I've tested this myself and it is working as expected.

In case anyone is interested in the details, the OpenAI documentation shows two different ways that streaming can be done.  We chose to go with the first method shown below.

1. [One method involves overriding the EventHandler class and using a with/as block](https://platform.openai.com/docs/assistants/overview/step-4-create-a-run)

```
with client.beta.threads.runs.stream(
  thread_id=thread.id,
  assistant_id=assistant.id,
  instructions="Please address the user as Jane Doe. The user has a premium account.",
  event_handler=EventHandler(),
) as stream:
  stream.until_done()
```

2. Another method involves creating a run and then iterating over the stream events ([see here and slide the "Streaming" slider to the right](https://platform.openai.com/docs/api-reference/runs/createRun))

```
stream = client.beta.threads.runs.create(
  thread_id="thread_123",
  assistant_id="asst_123",
  stream=True
)

for event in stream:
  print(event)

```

PR #1416, opened 2024-07-04
MAVExplorer: allow arbitrary expressions to be plotted on the map

tridge:
this allows mavextra to construct new mappable objects with Lat and Lng fields, allowing for derived positions from logs

this also generalises and simplifies the handling of instances

for example:
image

this uses the following mavextra function:
```
def pos_offset(GPS,offs_N,offs_E):
    latlon = mp_util.gps_offset(GPS.Lat,GPS.Lng, offs_E, offs_N)
    return PosObj(latlon[0],latlon[1],GPS.Alt)
```


peterbarker (approved):
LGTM

PR #1421, opened 2024-07-10
Flake8 compliance for SIYI module

peterbarker:
Fixes one small issue where tracebacks within an exception would cause another exception


tridge:
looks fine, can be merged if it has been tested

PR #1426, opened 2024-07-24
Add recommended modules

### LINKED ISSUE #1423: SecureCommand module not working properly on windows
**Problem description:**
When trying to remove public keys from a flight controller running ardupilot firmware I ran this command "module load SecureCommand" and got "Please install monocypher with: python3 -m pip install pymonocypher" message.
pymonocypher is already installed via "py -m pip install pymonocypher".
It seems like pymonocypher is not being included by default when building.

**Expected behavior:**
When running "module load SecureCommand" command it should load the module without no problems.

**Current behavior:**
Got "Please install monocypher with: python3 -m pip install pymonocypher" message after running the command.

**Operating system:** 
Windows 10

**Python version:**
3.12.14

**pip version:**
24.0

**Steps to reproduce**
1. Download and install the [latest release](https://firmware.ardupilot.org/Tools/MAVProxy/) of MavProxy for windows
2. Connect a Flight Controller to the Computer by USB
3. Open a command prompt window
4. Run "mavproxy --master=COMx" command (x is replaced by the port assigned to the flight controller)
5. Run "module load SecureCommand"

**Workaround:**
Digging a little I found [this comment](https://github.com/ArduPilot/MAVProxy/issues/181#issuecomment-131489544) which says that mavlink does not use the python environment on windows, thus, external libraries needs to be pasted in the MavProxy/_internal/ directory.
So I copied and pasted pymonocypher library folder and .pyd file into the MavProxy/_internal/ directory and then ran module load SecureCommand but got the same error message.
Then I built mavproxy [v1.8.70](https://github.com/ArduPilot/MAVProxy/tree/v1.8.70) (by running MAVProxyWinBuild.bat file) and then copied and pasted the same library folder and the .pyd file into MavProxy/_internal/ directory and then ran SecureCommand and it worked.
At first I got Mavproxy v1.8.69 from [here](https://firmware.ardupilot.org/Tools/MAVProxy/) (the latest release windows installer).

stephendade:
Add a [recommended] section to setup.py, where we can put all the optional module requirements.

Can use this via ``pip install mavproxy[recommended] --user``

Fixes #1423

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -33,7 +33,7 @@ jobs:
         run: ./installer.exe /verysilent /allusers /dir=inst
       - name: Build MAVProxy
         run: |
-          python -m pip install . --user
+          python -m pip install .[recommended] --user
           python -m pip list
       - name: Prepare installer
         run: |
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,9 @@ on how to use MAVProxy.''',
       extras_require={
         # restserver module
         'server': ['flask'],
+        'recommended': ['flask', 'PyYAML', 'lxml', 'wxpython',
+                        'pymonocypher', 'openai', 'paho-mqtt',
+                        'piexif', 'pynmea2', 'Pygame', 'Pillow']
       },
       scripts=['MAVProxy/mavproxy.py',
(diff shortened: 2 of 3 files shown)


peterbarker:
Merged, thanks!

Issue #1425, opened 2024-07-21
Map Tiles are not loaded (Tiles show up as Unavailable)


Notenlish:
I'm using OpenStreetMap for map tiles but even though it fetches the tiles, they just show up as unavailable for some reason. I tried every other tile provider and it either does the same thing or it just raises this error: `[ WARN:0@191.888] global loadsave.cpp:248 cv::findDecoder imread_('C:\Users\─░hsan\AppData\Local\.tilecache\MicrosoftSat\19\317271\479389.img'): can't open/read file: check file path/integrity`

stephendade:
I've found the Microsoft and Google based maps still work. But some others (like OpenStreetMap) do not - likely because they've changed their server details for downloading tiles.

Notenlish:
I think it was due to my network, now Microsoft's api works but yeah some tile providers need to be updated(or their api has changed)

PR #1427, opened 2024-07-25
Modules: Update tile services

stephendade:
Two things addressed here:

-  A number of tile services no longer worked. In some cases a URL rename was needed. In other cases, the provider no longer existed.
-  Added an url-agent to map tile download requests, as the default agent for python3 urllib was getting blocked.

NB: Was unable to test the ``GoogleChina``, ``StatkartTopo``, ``StatkartBasemap`` and ``Eniro DK,NO,SE,FI,PL`` as they might be geo-blocked (at leat in Australia)?

Fix for #1425

### DIFF
--- a/MAVProxy/modules/mavproxy_map/mp_tile.py
+++ b/MAVProxy/modules/mavproxy_map/mp_tile.py
@@ -52,7 +52,7 @@ class TileException(Exception):
 TILE_SERVICES = {
     # thanks to http://go2log.com/2011/09/26/fetching-tiles-for-offline-map/
     # for the URL mapping info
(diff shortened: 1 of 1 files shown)


stephendade:
@AndKe might be able to confirm some of these services?

peterbarker:
Merged, thanks!

AndKe:
@stephendade I just returned from vacation, I have contacted kartverket.no (the Norwegian mapping service) about "statkart" - the servers are awfully slow, I can barely get any data thru. - but the URL's are still valid. 
Eniro does not seem to work, and I am on the case :)

AndKe:
Hmm, I use the exactly same URL/format from Eniro in MAVproxy as in the QGC, it still works fine in QGC, but not in MAVproxy ...  can it be some User-Agent trickery ?

AndKe:
@peterbarker  : please see https://github.com/ArduPilot/MAVProxy/pull/1430

PR #1431, opened 2024-07-31
MAVProxy: fixed reconnection of serial port on windows after reboot

joshanne:
The line: `master.port.inWaiting()` is causing a SerialException when a reboot is sent to the flight controller from the console.

This change has been tested in windows and linux, it does not appear to have an effect on linux. The exception does not trigger in linux, but does on windows.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -1116,8 +1116,11 @@ def main_loop():
 
         for master in mpstate.mav_master:
             if master.fd is None:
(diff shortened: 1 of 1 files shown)


joshanne:
Also, tested disconnect of USB in windows, and flight controller reconnects as expected.

peterbarker:
Merged, thanks!

tridge:
@joshanne great, thanks!

Cimos:
@joshanne This is probably the best bug fix ever.

PR #1432, opened 2024-08-01
Modules: grapher layout and file export fixed for Windows

stephendade:
Fixed two issues with MAVExplorer graphs not working nicely:
- Changed to "TkAgg" backend on Windows, as the "wxAgg" backend wasn't displaying the chart values (lower right) correctly
- When saving graphs to file ("save" icon on graph window), made the _suggested_ filename safer by removing problematic characters (ie ``Sensors/GPS/GPS Accuracy`` to ``Sensors_GPS_GPS_Accuracy``)

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -50,6 +50,7 @@ jobs:
           $wc.DownloadFile('https://autotest.ardupilot.org/Parameters/ArduPlane/apm.pdef.xml', 'Parameters\ArduPlane.xml')
           $wc.DownloadFile('https://autotest.ardupilot.org/Parameters/ArduSub/apm.pdef.xml', 'Parameters\ArduSub.xml')
           $wc.DownloadFile('https://autotest.ardupilot.org/Parameters/AntennaTracker/apm.pdef.xml', 'Parameters\AntennaTracker.xml')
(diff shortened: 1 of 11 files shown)


peterbarker:
@MichelleRos this seems to say that graphing works for Stephen; could you test this PR to see if graphing actually works for you after this, please?

MichelleRos:
Just tested it. Graphing seems to be working, and the save icon does work on the "canned" graphs I tested, but when I tried just graphing something from the command line it threw this error and didn't bring up the dialog box:

`
Exception in Tkinter callback
  File "tkinter\__init__.py", line 1921, in __call__
  File "matplotlib\backends\_backend_tk.py", line 858, in save_figure
  File "MAVProxy\modules\lib\grapher.py", line 293, in <lambda>
TypeError: 'NoneType' object is not iterable
`

Note that I've been using MAVExplorer via WSL rather than using the Windows one, so I don't know what did or didn't work before. 😅

stephendade:
> Just tested it. Graphing seems to be working, and the save icon does work on the "canned" graphs I tested, but when I tried just graphing something from the command line it threw this error and didn't bring up the dialog box:

That's now been fixed.

I also added a few more fixes based upon https://discuss.ardupilot.org/t/mavproxy-version-in-windows/121556/10?u=stephendade:
- Heli param definitions now included in build
- Fixed up another graphing issue in Windows (magfit in MAVExplorer)
- Modernised the Windows build scripts

PR #1433, opened 2024-08-03
chat: Changed longitude is "lat" to longitude is "lon"

ShashwathD:
Previously said that lat was short for longitude, but I changed it to lon, which may fix some problems with the assistant when it comes to certain prompts from the user.

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/assistant_setup/assistant_instructions.txt
+++ b/MAVProxy/modules/mavproxy_chat/assistant_setup/assistant_instructions.txt
@@ -44,7 +44,7 @@ If the user asks to change the yaw (aka heading) of the vehicle, you should firs
 
 The short form of "altitude" is "alt".
 The short form of "latitude" is "lat".
-The short form of "longitude" is "lat".
+The short form of "longitude" is "lon".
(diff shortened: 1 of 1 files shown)


rmackay9 (commented):
Thanks very much for this!

One small thing is that the commit message should be prefixed with "chat:".  We normally add a prefix of the subsystem (or directory) that has been changed because it makes backporting easier.  If you have a [peek at our recent commit list](https://github.com/ArduPilot/MAVProxy/commits/master/) you'll see they all have a prefix

If you're not familiar with how to fix the commit I could do it for you

Thanks again!

ShashwathD:
I’m not quite familiar with fixing commits so it would be great if you
could do it for me.

Thanks,
Shashwath

ShashwathD:
I added the chat: at the start, so should it work now?

rmackay9:
Hi @ShashwathD,

Thanks, it's not as easy as changing the PR title (which is what you've kindly updated).  It's the commit title that we need to fix

rmackay9:
I've forced push to your branch now so this should be good to merge now, thanks!

peterbarker:
Merged, thanks!

PR #1435, opened 2024-08-13
chat: push to record

adityaomar3:
This PR adds a push to record feature i.e. it will record the audio on mouse left key press down and stop the recording on key release.

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
@@ -5,7 +5,6 @@ Randy Mackay, December 2023
 AP_FLAKE8_CLEAN
 '''
 
-import time
 
 try:
     import pyaudio  # install using, "sudo apt-get install python3-pyaudio"
@@ -15,10 +14,13 @@ except Exception:
     print("chat: failed to import pyaudio, wave or openai.  See https://ardupilot.org/mavproxy/docs/modules/chat.html")
     exit()
 
+# initializing the global list to keep and update the stop_recording state
+stop_recording = [False]
+
 
 class chat_voice_to_text():
     def __init__(self):
-        # initialise OpenAI connection
+        # initialise variables
         self.client = None
         self.assistant = None
 
@@ -53,21 +55,19 @@ class chat_voice_to_text():
             print("chat: failed to connect to microphone")
             return None
 
-        # calculate time recording should stop
-        curr_time = time.time()
-        time_stop = curr_time + 5
-
         # record until specified time
         frames = []
-        while curr_time < time_stop:
+        while not stop_recording[0]:
             data = stream.read(1024)
             frames.append(data)
-            curr_time = time.time()
 
         # Stop and close the stream
         stream.stop_stream()
         stream.close()
         p.terminate()
+
+        # update the recording state back to false globally
+        stop_recording[0] = False
 
         # Save audio file
         wf = wave.open("recording.wav", "wb")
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/mavproxy_chat/chat_window.py
```diff
@@ -17,6 +17,9 @@ def __init__(self, mpstate, wait_for_command_ack_fn):
         # keep reference to mpstate
         self.mpstate = mpstate
 
+        # keep the recording state
+        self.isRecording = False
```
rmackay9, line 21: let's change this to is_recording.  We normally use lower case and underscores

adityaomar3: Sure! In-fact I just took a look, I think its no longer needed, will fix it.

MAVProxy/modules/mavproxy_chat/chat_window.py
```diff
@@ -121,7 +123,7 @@ def apikey_close_button_click(self, event):
 
     # record button clicked
     def record_button_click(self, event):
-        # run record_button_click_execute in a new thread
+        print("record button clicked")
```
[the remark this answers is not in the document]
adityaomar3: I have removed the function record_button_click() and updated the code to record_button_pushed()

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
```diff
@@ -53,22 +55,18 @@ def record_audio(self):
...
             return None
 
-        # calculate time recording should stop
-        curr_time = time.time()
-        time_stop = curr_time + 5
-
         # record until specified time
         frames = []
-        while curr_time < time_stop:
+        # while curr_time < time_stop and not self.stop_recording:
```
rmackay9, line 60: we should remove this line which is commented out

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
```diff
@@ -53,22 +55,18 @@ def record_audio(self):
...
             data = stream.read(1024)
             frames.append(data)
-            curr_time = time.time()
 
         # Stop and close the stream
         stream.stop_stream()
         stream.close()
         p.terminate()
-
+        stop_recording[0] = False
```
rmackay9, line 69: could you add a blank line number this line?

rmackay9: actually I don't think we even need this line, "stop_recording[0] = False" because we check it just a few lines higher so we already know it is false.  It was required when there was still a timeout

rmackay9:
I've tested this and it works great!

One thing that is really impressive is how quick OpenAI can convert the audio to text.  The response is nearly instantaneous

rmackay9 (changes requested):
I've asked for a couple of non-functional changes but after that we can merge it, thanks!

MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
adityaomar3: Hi! We have to update the state of the stop_recording back to false otherwise it would not work for the second time, as globally state would be true always so it will not enter the while loop.

adityaomar3: More over I can give some spaces and add a comment for the same

rmackay9: yes, sorry, you're right

rmackay9 (approved):
LGTM, thanks!

PR #1442, opened 2024-09-01
module: Websocket

jheising:
Added a module to support input and output of MAVLink messages via a websocket connection.

I'm not sure what the process is for submitting a new module, but I figured I would create this PR to start the process. Questions:

1. Do I need to create any tests? Wasn't clear to me if this was being done for other modules because I couldn't find any.
2. Updates to MavProxy documentation? I assume I would submit this as an update the the ArduPilot MavProxy wiki if/when this PR was approved and merged?

Thanks!

### DIFF
--- a/MAVProxy/modules/mavproxy_ws.py
+++ b/MAVProxy/modules/mavproxy_ws.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+'''
+WebSocket Module
+Jim Heising
+August 2024
+'''
+
+from MAVProxy.modules.lib import mp_module, mp_settings
+from websockets.sync.server import serve
+from threading import Thread
+from websockets import ConnectionClosedOK
+from pymavlink import mavutil
+import json
+import traceback
+
+mavlink_map = mavutil.mavlink.mavlink_map
+message_name_2_type_cache = {}
+
+# Some examples of JSON MAVLink messages
+# {"mavpackettype": "COMMAND_LONG", "target_system":0, "target_component": 0, "command": 400, "confirmation": 0, "param1": 1, "param2": 0, "param3": 0, "param4": 0, "param5": 0, "param6": 0, "param7": 0}
+# {"mavpackettype": "PARAM_REQUEST_LIST", "target_system":1, "target_component": 1}
+# {"mavpackettype": "PARAM_REQUEST_READ", "target_system":1, "target_component": 1, "param_id": "", "param_index":1}
+
+# Convert the name of a MAVLink message into its python class type
+def get_mavlink_msg_type_from_name(name):
+    # Check to see if this is cached already
+    msg_type = message_name_2_type_cache.get(name)
+
+    if msg_type:
+        return msg_type
+
+    for msg_id in mavlink_map.keys():
+        msg_type = mavlink_map[msg_id]
+        mav_cmd_name = msg_type.msgname if hasattr(msg_type, "msgname") else msg_type.name
+        if mav_cmd_name == name:
+            # Cache this for future use
+            message_name_2_type_cache[name] = msg_type
+            return msg_type
+
+
+# Convert a JSON message into a proper MAVLink message
+def json_to_mavlink(json_string):
+    # Parse our JSON string
+    mavlink_dict = json.loads(json_string)
+
+    # Get the name of the MAVLink packet type
+    requested_cmd_name = mavlink_dict["mavpackettype"]
+
+    # Check to see if the message type is known
+    msg_type = get_mavlink_msg_type_from_name(requested_cmd_name)
+
+    if msg_type:
+        # If we get here, we've found a mavlink message that matches our mavpackettype name
+        # Now we can parse out our JSON attributes in their proper order create our MAVLink message
+        msg_args = []
+        for field_name in msg_type.fieldnames:
+            field_value = mavlink_dict.get(field_name)
+            # Convert strings to bytes
+            if isinstance(field_value, str):
+                field_value = field_value.encode("ascii")
+            msg_args.append(field_value)
+
+        # Here we actually create the actual MAVLink message from the python class
+        return msg_type(*msg_args)
+
+
+class WebSocket(mp_module.MPModule):
+    def __init__(self, mpstate):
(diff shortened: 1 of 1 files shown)


peterbarker:
Don't we already have websocket support via Pet Hall's work in pymavlink here?  https://github.com/ArduPilot/pymavlink/pull/967

jheising:
Well heck, @peterbarker I didn't notice that!

@IamPete1, I'm curious on your thoughts as to the benefits of this one vs. the one you created. I'm totally fine to close this PR and potentially use yours instead, but there may be some benefits to combining our efforts to gain the maximum amount of functionality.

These were the high-level requirements for my PR:

1. Support for more than one concurrent connection.
2. Support for both JSON and raw message formats.
3. Support for API key validation for increased security in shared network environments.
4. Support for both sending and receiving MAVLink messages.

From what I can tell:

1. This PR supports more than one concurrent connection. Does the other one support more than 1 connection?
6. This PR supports both JSON and raw MAVLink message formats. It looks like the other one only supports raw?
7. This PR supports the ability to specify an API key to make the endpoint more secure. Other PR does not appear to support.
8. This PR supports both inbound and outbound MAVLink messages. The other one says that client->server messages might work, but it doesn't appear to be tested and from a cursory glance at the code I'm not sure it will work (but could totally be wrong).

Anyway, I'm happy to abandon this PR if the other one can provide similar functionality. It's just a question of whether or not it would be easier to adapt this functionality to the other PR and/or to just switch to this. Also I guess the question is: does websocket support belong in a connection_type or a module? I originally though about adding it as a connection_type, but when I saw that the REST interface was created as a module, I went the module direction. Curious what other people think.

Thanks!

IamPete1:
Hi @jheising. My implementation was down at the connection level because I don't want to have to run MAProxy, a simple translation script is enough. I have been using https://github.com/ArduPilot/pymavlink/blob/master/examples/mavtcpsniff.py with the second connection as websocket. What I have done is really a bare minimum any improvements would be great. 

Some thoughts from my end:

- JSON format requires that both ends of the WebSocket understand each message. The binary format requires only that the client end understands the message. 
- We could do both raw and json at once on using the text message and the other using the binary. However that will mean that half the messages we send are wasted.
- If using the raw message multi client is harder as there is no guarantee that each client will send a complete message. 
- I would like to keep it simple enough that we could also add a websocket forwarding option to Mission Planner.

jheising:
@IamPete1 to address some of the concerns:

> JSON format requires that both ends of the WebSocket understand each message

Agreed, although the JSON format I used was simply the `pymavlink.to_json` function, so it should be as close to a "spec" as possible. The nice thing about supporting JSON is that you don't have to import a large client-side library in your code if you don't want to. For example, I can communicate directly to the JSON endpoint by just looking up the MAVLink documentation online and passing it a message with the exact same names  as shown in the documentation:

`{"mavpackettype": "COMMAND_LONG", "target_system":0, "target_component": 0, "command": 400, "confirmation": 0, "param1": 1, "param2": 0, "param3": 0, "param4": 0, "param5": 0, "param6": 0, "param7": 0}`

It also makes it really easy to debug and test with command line tools like `wscat` or even the browser debugger console.

> "We could do both raw and json at once on using the text message and the other using the binary"

There is no need to waste any bandwidth, you just need to indicate the format you want to communicate in when you connect. Remember, websockets start with a standard HTTP request so we have full support for paths, URL parameters, etc. like any other HTTP endpoint. So for example, if you want to send/receive raw/binary format you just connect to `ws://myserver:myport` and if you want to access the JSON format you just connect to `ws://myserver:myport/json`. The websocket server just looks at the URL path of the websocket connection to determine what format it should use.

> "If using the raw message multi client is harder as there is no guarantee that each client will send a complete message."

Not sure how the websocket library you are using works, but the one I'm using treats each inbound connection as it's own instance, so I don't think there should be any issues with incomplete messages. Also the implementation I provided works in complete messages so it will only forward the messages to pymavlink when the message is complete.

Anyway, @IamPete1, totally fine if this version is a bit more overkill than what you needed and doesn't make sense to integrate into pymavlink.

@peterbarker I still think it would make sense to consider this merge into MAVProxy for the following reasons:

1. For those of us wanting to create web-based interfaces into their autopilots, JSON is a logical and easy choice.
2. JSON is partially supported in the REST module, but it appears to only support server->client, not client->server.
3. JSON support via the REST module only works in a polling architecture, whereas this provides a real-time messaging architecture for better performance.

Please let me know what you think, thanks!

jheising:
FYI, if you want to see a quick video of it in action: [https://www.youtube.com/watch?v=5mEAPg7Ye7I](https://www.youtube.com/watch?v=5mEAPg7Ye7I)

IamPete1:
> FYI, if you want to see a quick video of it in action: https://www.youtube.com/watch?v=5mEAPg7Ye7I

Nice!
> There is no need to waste any bandwidth, you just need to indicate the format you want to communicate in when you connect. Since websockets start with a standard HTTP request we have full support for paths, URL parameters, etc., like any other HTTP request. So for example, if you want to send/receive raw/binary format you just connect to `ws://myserver:myport` and if you want to access the JSON format you just connect to `ws://myserver:myport/json`. The websocket server just looks at the URL path of the websocket connection to determine what format it should use.
> 

That is a nice solution, I had not thought of that. It would be great to also support it at the connection level.

MAVProxy/modules/mavproxy_ws.py
```diff
@@ -0,0 +1,179 @@
...
+from websockets import ConnectionClosedOK
+from pymavlink import mavutil
+import json
+import traceback
+
+mavlink_map = mavutil.mavlink.mavlink_map
+message_name_2_type_cache = {}
+
+# Some examples of JSON MAVLink messages
+# {"mavpackettype": "COMMAND_LONG", "target_system":0, "target_component": 0, "command": 400, "confirmation": 0, "param1": 1, "param2": 0, "param3": 0, "param4": 0, "param5": 0, "param6": 0, "param7": 0}
```
IamPete1, line 21: You really want the header in here too. So you can check where the messages are coming from. There is already a "standard" for MAVLink over JSON from the mavlink2rest lib here: https://github.com/mavlink/mavlink2rest?tab=readme-ov-file#api

jheising: That's a great callout. I will update the PR to move to that.

PR #1446, opened 2024-09-04
link: detect Sierra Aerospace USB

tridge:


tridge:
I've found this boards and started debugging the ethernet.
The failure is in waiting for the SWR bit to clear in the ETH_DMAMR register. 
image

PR #1448, opened 2024-09-09
Avoid exceptions when wp module is not loaded

peterbarker:
Exceptions are easy reproduce - just `module unload wp`


peterbarker:
Merged this, thanks @Ryanf55

PR #1450, opened 2024-09-09
mavproxy_link.py: add bytes-available-to-read counter in link stats o…

peterbarker:
…utput

Trying to diagnose 5-to-7-second lag on multiple-link environment.  Potential CPU overload leading to vast backlog of bytes?!


peterbarker:
We could hide this value if it is zero - which it seems to be most of the time

PR #1462, opened 2024-10-01
Use importlib to handle version and package resources, where available

peterbarker:
We fall back to the previous efforts on failure.

Tested back to 3.6

2.7 was a complete write-off due to pre-existing changes.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -24,13 +24,7 @@ import threading
 import time
 import traceback
 
-try:
-    reload
-except NameError:
-    try:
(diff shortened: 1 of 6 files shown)


peterbarker:
Since pkg_resources is going away and importlib has worked since forever, move to just using it now.

peterbarker:
Closes https://github.com/ArduPilot/MAVProxy/issues/1672

stephendade:
Tested on Windows. Works well!

peterbarker:
> Tested on Windows. Works well!

Fantastic, thanks!

Merging this one.

PR #1464, opened 2024-10-04
windows: include requests in package

peterbarker:
so mavproxy_help works


peterbarker:
@stephendade this didn't work to unbreak the help module on Windows.

PyPi got rid of the xmlrpc interace, so I changed us to JSON.  Works on Linux, but missing the "requests" package on Windows.

Any ideas?

stephendade:
I'll take a look sometime this weekend

peterbarker:
> Turns out that the `requests` package isn't included in a Windows Python install. You'll need to `pip install` it, and add to `setup.py`

I've done as you suggested there and it all works - thanks!

PR #1465, opened 2024-10-04
 mavproxy_link: use numeric code for green

peterbarker:
unreadable on Window's concept of "green", so I swiped the RGB values from Linux


peterbarker:
Tested and working on Windows here, Michelle tested an earlier  version which didn't update the status indicators.

PR #1466, opened 2024-10-04
mavproxy_messagerate: do special handling for reset and stop set

magate:
### Before
image

### After
image
image
image


peterbarker (approved):
LGTM

We should probably also add `messagerate setdefault MESSAGEID`

peterbarker:
Merged, thanks!

PR #1467, opened 2024-10-07
wxconsole_ui: reorder setting of terxt and fg/bg colours

peterbarker:
Windows fails to update colours on text in the wx console.  Rearranging the setting ofthe text and colours fixes this.


MichelleRos:
Just tested on Windows. Doesn't fix it, unfortunately.

peterbarker:
> Just tested on Windows. Doesn't fix it, unfortunately.

Thanks for testing!

PR #1472, opened 2024-10-12
MAVProxy: add compatibility code for MAV_TYPE_VTOL_DUOROTOR rename to…

peterbarker:
… MAV_TYPE_VTOL_TAILSITTER_DUOROTOR

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -675,6 +675,19 @@ class LinkModule(mp_module.MPModule):
 
         return True
 
+    mav_type_planes = [
+        mavutil.mavlink.MAV_TYPE_FIXED_WING,
+        mavutil.mavlink.MAV_TYPE_VTOL_QUADROTOR,
+        mavutil.mavlink.MAV_TYPE_VTOL_TILTROTOR,
+    ]
+    # VTOL_DUOROTOR was renamed to VTOL_TAILSITTER_DUOROTOR
+    for possible_plane_type in "VTOL_DUOROTOR", "VTOL_TAILSITTER_DUOROTOR":
+        t = f"MAV_TYPE_{possible_plane_type}"
+        attr = getattr(mavutil.mavlink, t, None)
+        if attr is None:
+            continue
+        mav_type_planes.append(attr)
+
     def master_msg_handling(self, m, master):
         '''link message handling for an upstream link'''
 
@@ -739,11 +752,7 @@ class LinkModule(mp_module.MPModule):
                 self.status.last_mode_announced = master.flightmode
                 self.say("Mode " + self.status.flightmode)
 
-            if m.type in [
-                    mavutil.mavlink.MAV_TYPE_FIXED_WING,
-                    mavutil.mavlink.MAV_TYPE_VTOL_DUOROTOR,
-                    mavutil.mavlink.MAV_TYPE_VTOL_QUADROTOR,
-                    mavutil.mavlink.MAV_TYPE_VTOL_TILTROTOR]:
+            if m.type in self.mav_type_planes:
                 self.mpstate.vehicle_type = 'plane'
                 self.mpstate.vehicle_name = 'ArduPlane'
             elif m.type in [mavutil.mavlink.MAV_TYPE_GROUND_ROVER,


MAVProxy/modules/mavproxy_link.py
```diff
@@ -675,6 +675,19 @@ def heartbeat_is_from_autopilot(self, m):
 
         return True
 
+    mav_type_planes = [
+        mavutil.mavlink.MAV_TYPE_FIXED_WING,
+        mavutil.mavlink.MAV_TYPE_VTOL_QUADROTOR,
+        mavutil.mavlink.MAV_TYPE_VTOL_TILTROTOR,
+    ]
+    # VTOL_DUOROTOR was renamed to VTOL_TAILSITTER_DUOROTOR
+    for possible_plane_type in "VTOL_DUOROTOR", "VTOL_TAILSITTER_DUOROTOR":
```
hamishwillee, line 684: You should never get "VTOL_DUOROTOR" after update because it isn't present in the mavlink XML and what is received will be a number. Perhaps I'm missing something.

peterbarker:
On Sat, 12 Oct 2024, Hamish Willee wrote:

> +    mav_type_planes = [
> +        mavutil.mavlink.MAV_TYPE_FIXED_WING,
> +        mavutil.mavlink.MAV_TYPE_VTOL_QUADROTOR,
> +        mavutil.mavlink.MAV_TYPE_VTOL_TILTROTOR,
> +    ]
> +    # VTOL_DUOROTOR was renamed to VTOL_TAILSITTER_DUOROTOR
> +    for possible_plane_type in "VTOL_DUOROTOR", "VTOL_TAILSITTER_DUOROTOR":
> 
> You should never get "VTOL_DUOROTOR" after update because it isn't present
> in the mavlink XML and what is received will be a number. Perhaps I'm
> missing something.

I'm *really* not sure what you mean by "after update".  MAVProxy *must* 
work with both newer and older mavlink message definitions whereever it is 
reasoanbly possible.

pymavlink and MAVProxy are not the same thing and release on different 
schedules, and we need to maintain backwards compatability for those that 
haven't updated their definitions - for a reasonably period of time, 
anyway.

peterbarker:
.... further, *not* adding backwards compatibility would completely stuff up development against older ArduPilot branches which are using very old bindings, which it would be unwise to fool around with.

hamishwillee:
> .... further, not adding backwards compatibility would completely stuff up development against older ArduPilot branches which are using very old bindings, which it would be unwise to fool around with.

@peterbarker 

How? We didn't change the number sent for enum, so ardupilot will send the same value, and MAVProxy would interpret it as the old value. Upshot, you don't have to change the bindings for those old versions and MAVProxy won't be able to tell.

That's why enum name changes are allowed by MAVLink, because they don't make semantic changes, and just require a recompile swapping the new messages. 

All that should be needed is that MAVProxy replace the old enum value with the new enum value. The only way you'd have a problem is if you tried to match the new MAVProxy with the old message definitions - and that doesn't make sense since the new MAVProxy is written against the new definitions.

From a practical point of view, it is also true that some languages can't defined the same number for different enum names, so creating an alias would potentially break some generated languages. (or so I understand).

Ultimately though, this is rubbish because the change went in more than two years ago and not one single person in MAVProxy land has noticed anything: https://github.com/mavlink/mavlink/pull/1818 :-)

hamishwillee:
PS If you still disagree, please raise this in the MAVLink dev call with @auturgy - we don't do it often, but we've always reserved the right to change the names of values. If that is to change we'd need to formalize it.

peterbarker:
On Tue, 15 Oct 2024, Hamish Willee wrote:

> All that should be needed is that MAVProxy replace the old enum value with the new enum value. The only way you'd have a problem is if you tried to match the new MAVProxy with the old
> message definitions - and that doesn't make sense since the new MAVProxy is written against the new definitions.

That's the issue - the message definitions come from pymavlink, *not* 
MAVProxy.  Or random other message definitions the user wants to use (eg. 
whatever is committed into the modules/mavlink submodule in the ArduPilot 
repo.

We have to work with both old and new bindings in MAVProxy whereever we 
can - check out what we're doing around

> Ultimately though, this is rubbish because the change went in more than two years ago and not one single person in MAVProxy land has noticed anything: mavlink/mavlink#1818 :-)

Right, but we haven't pulled that change into ArduPilot's mavlink message 
set - and we should have by now.  I'd rather not have a situation where we 
*can't* move to be closer to upstream in terms of message set, as we're 
then more likely to end up with incompatible changes.

hamishwillee:
>  Or random other message definitions the user wants to use (eg.
whatever is committed into the modules/mavlink submodule in the ArduPilot repo.

But that's a flaw in MAVProxy - the code assumes existence of an entity that is not guaranteed to exist - since you're allowing any definitions at all. Shouldn't the code only add entities to the list if they exist?
I guess that is arguable since we should be able to assume minimal.xml.

There are two main points here though:
1. I am not sure that all languages support two enum value definitions with the same value. If this is the case then we can consider this because it won't break other languages.
2. You say we don't have to do this forever but how can that be managed. Without the pain point nothing will happen. That is proved by the fact we're here now after two years.

hamishwillee:
@auturgy is going to discuss this with you @peterbarker - we'd prefer not to do this, but if the case is compelling it requires a proper process.

peterbarker:
> > Or random other message definitions the user wants to use (eg.
> > whatever is committed into the modules/mavlink submodule in the ArduPilot repo.
> 
> But that's a flaw in MAVProxy - the code assumes existence of an entity that is not guaranteed to exist - since you're allowing any definitions at all. Shouldn't the code only add entities to the list if they exist? I guess that is arguable since we should be able to assume minimal.xml -but if you're supporting everything forever then it's on your code to be robust.

"Shouldn't the code only add entities to the list if they exist" - yes, that's exactly what this PR does :-)  It checks to see  enumeration entries exist before adding them to a list of known "plane types"

auturgy:
@hamishwillee Whilst this might feel a little “impure”, it’s just compatibility code to solve a problem, and doesn’t break anything external. Beyond learning the lesson on pymavlink enum parsing I don’t think we should lean into it.The “compatability” in the commit name however, causes me angst, and @peterbarker should fix it. “compatibility” please ;)On 10 Nov 2024, at 16:45, Peter Barker ***@***.***> wrote:﻿


Or random other message definitions the user wants to use (eg.
whatever is committed into the modules/mavlink submodule in the ArduPilot repo.

But that's a flaw in MAVProxy - the code assumes existence of an entity that is not guaranteed to exist - since you're allowing any definitions at all. Shouldn't the code only add entities to the list if they exist? I guess that is arguable since we should be able to assume minimal.xml -but if you're supporting everything forever then it's on your code to be robust.

"Shouldn't the code only add entities to the list if they exist" - yes, that's exactly what this PR does :-)  It checks to see  enumeration entries exist before adding them to a list of known "plane types"

hamishwillee:
I'm fine with this code thanks. My problem is the suggestion that we add the old enum value back in. We can't - the toolchain blocks having two enum values with the same number, because that isn't allowed by some languages.

auturgy:
Agreed: not adding it back.Apologies for missing that in the thread.On 12 Nov 2024, at 17:56, Hamish Willee ***@***.***> wrote:﻿
I'm fine with this code thanks. My problem is the suggestion that we add the old enum value back in. We can't - the toolchain blocks having two enum values with the same number, because that isn't allowed by some languages.

peterbarker:
> I'm fine with this code thanks. My problem is the suggestion that we add the old enum value back in. We can't - the toolchain blocks having two enum values with the same number, because that isn't allowed by some languages.

Ah, whoops.  I'd abandoned that thought path as soon as you'd mentioned the language-incompatibility (incompatability?) problem!

hamishwillee:
Excellent. Talking at cross purposes then. Thanks for your patience.

A take home here is that MAVLink feels free to modify enum value names. For this case I messaged ardupilot-gcs@googlegroups.com (the GCS group). IF there is more I should do in this case very happy to discussion.

peterbarker:
> Excellent. Talking at cross purposes then. Thanks for your patience.
> 
> A take home here is that MAVLink feels free to modify enum value names. For this case I messaged [ardupilot-gcs@googlegroups.com](mailto:ardupilot-gcs@googlegroups.com) (the GCS group). IF there is more I should do in this case very happy to discussion.

Odd, I can't find that in the list of discussions on that Google Group.  Check to see if it bounced?  Got an approximate date?

hamishwillee:
March 2022 - https://github.com/mavlink/mavlink/pull/1818 - but I can't find any record there or in my gmail, so it is most likely I did not sent it. Though I recall agreeing I would do so as part of the update.

PR #1480, opened 2024-11-08
MAVCesium: Switch to newer fork

Ryanf55:
# Purpose

Add my MAVCesium fork and expose it a bit better in MAVProxy.

# Expectations

With `mavproxy.py --cesium`, it will start the web server, which you can open at `http://0.0.0.0:5000/mavcesium/#`.

You can also still do `module load cesium`.

Currently, mavproxy does not connect properly when launched in this mode. It works best with the standalone mode, where you launch mavproxy, then launch mavcesium separately in another terminal. 

```bash
python3 cesium_web_server.py
```

I'll chase that down soon - it's likely a difference in default options because the websocket receives no data.
I'd like to propose merging this as-is with the known deficiency because it's still a large improvement.

Here's what you get:

Birds eye view of Kaga (default on open when you start SITL in Kaga)
image

Once you take off, the takeoff orbit is marked in pink.
image
There is a known issue of the vehicle model being yawed 90 off. I'll fix it later.

The forward facing view works too
image

If you switch to FBWB, you can turn on the camera projection, and get a good idea of what it looks like.
image

# Follow up

I'd like to get it fully working (integrated as a mavproxy module) and fix any obvious bugs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "MAVProxy/modules/mavproxy_cesium"]
 	path = MAVProxy/modules/mavproxy_cesium
-	url = https://github.com/SamuelDudley/MAVCesium.git
+	url = https://github.com/ArduPilot/MAVCesium.git


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "MAVProxy/modules/mavproxy_cesium"]
 	path = MAVProxy/modules/mavproxy_cesium
-	url = https://github.com/SamuelDudley/MAVCesium.git
+	url = https://github.com/Ryanf55/MAVCesium.git
```
peterbarker, line 3: We should clone this into the ArduPilot github set of repos instead, I think.

MAVProxy/mavproxy.py
```diff
@@ -1516,6 +1517,9 @@ def quit_handler(signum = None, frame = None):
     if opts.map:
         process_stdin('module load map')
 
+    if opts.mavcesium:
+        process_stdin('module load cesium')
```
peterbarker, line 1521: I'm not a great fan of this.  We have `--load-module` already.

peterbarker (dismissed):

.gitmodules
Ryanf55: Ok. I don't have admin. Is that something you could set up?

MAVProxy/mavproxy.py
Ryanf55: Ah, ok, I was just trying to make it like `mavproxy --map` but for 3D. I'll drop this.

Ryanf55: Dropped.

tridge (approved):

PR #1481, opened 2024-11-08
Add Start Websocket Server to the MAVProxy console

Ryanf55:
# Purpose

Add ability to add a websocket server in the MAVProxy menu. It's only added if wsproto is installed.

# Related

* https://github.com/ArduPilot/ardupilot/pull/28560
* https://github.com/ArduPilot/pymavlink/pull/967

# Demo
[mavproxy_1481_websocket_server.webm](https://github.com/user-attachments/assets/02124a5e-273e-47b8-abe4-f94c087a359e)

# Follow Up

* Give users the ability to change which port number. The current one matches webtools.
   * Any tips on other places that open up a text box when you click?

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -109,12 +109,24 @@ class LinkModule(mp_module.MPModule):
         self.menu_added_console = False
         if mp_util.has_wxpython:
             self.menu_rm = MPMenuSubMenu('Remove', items=[])
-            self.menu = MPMenuSubMenu('Link',
-                                      items=[MPMenuItem('Add...', 'Add...', '# link add ', handler=MPMenulinkAddDialog()),
-                                             self.menu_rm,
-                                             MPMenuItem('Ports', 'Ports', '# link ports'),
-                                             MPMenuItem('List', 'List', '# link list'),
-                                             MPMenuItem('Status', 'Status', '# link')])
+
+            items = [
+                MPMenuItem('Add...', 'Add...', '# link add ', handler=MPMenulinkAddDialog()),
+                self.menu_rm,
+                MPMenuItem('Ports', 'Ports', '# link ports'),
+                MPMenuItem('List', 'List', '# link list'),
+                MPMenuItem('Status', 'Status', '# link')]
+            # wsproto is not installed by default.
+            # Only add the menu if it's available.
+            try:
+                import wsproto  # noqa: F401
+            except ImportError:
+                pass
+            else:
+                items.append(
+                    MPMenuItem('Start Websocket Server', 'Start Websocket Server', '# output add wsserver:0.0.0.0:56781'))
+
+            self.menu = MPMenuSubMenu('Link', items=items)
             self.last_menu_update = 0
 
     def idle_task(self):


IamPete1:
If you start on the same port as mission planner then Telemetry Dashboard should auto connect. That would be `56781`, we could make that the default value too.

Ryanf55:
> If you start on the same port as mission planner then Telemetry Dashboard should auto connect. That would be `56781`, we could make that the default value too.

Where's the 5863 come from? 
image

I'm happy to switch.

IamPete1:
> Where's the 5863 come from?

Its what I was using when I first started, just the SITL port I was using + 100. Completely arbitrary.

MAVProxy/modules/mavproxy_console.py
```diff
@@ -91,11 +91,22 @@ def __init__(self, mpstate):
         # create the main menu
         if mp_util.has_wxpython:
             self.menu = MPMenuTop([])
+            mavproxy_items = [MPMenuItem('Settings', 'Settings', 'menuSettings'),
+                              MPMenuItem('Show Map', 'Load Map', '# module load map'),
+                              MPMenuItem('Show HUD', 'Load HUD', '# module load horizon'),
+                              MPMenuItem('Show Checklist', 'Load Checklist', '# module load checklist')]
```
peterbarker, line 97:
```suggestion
            mavproxy_items = [
                MPMenuItem('Settings', 'Settings', 'menuSettings'),
                MPMenuItem('Show Map', 'Load Map', '# module load map'),
                MPMenuItem('Show HUD', 'Load HUD', '# module load horizon'),
                MPMenuItem('Show Checklist', 'Load Checklist', '# module load checklist'),
            ]
```

MAVProxy/modules/mavproxy_console.py
```diff
@@ -91,11 +91,22 @@ def __init__(self, mpstate):
...
+                pass
+            else:
+                mavproxy_items.append(MPMenuItem('Start Websocket Server', 'Start Websocket Server', '# output add wsserver:0.0.0.0:56781'))
+
             self.add_menu(MPMenuSubMenu('MAVProxy',
-                                        items=[MPMenuItem('Settings', 'Settings', 'menuSettings'),
-                                               MPMenuItem('Show Map', 'Load Map', '# module load map'),
-                                               MPMenuItem('Show HUD', 'Load HUD', '# module load horizon'),
-                                               MPMenuItem('Show Checklist', 'Load Checklist', '# module load checklist')]))
+                                        items=mavproxy_items))
```
peterbarker, line 109: probably don't need to split this onto a separate line now

peterbarker (changes requested):

MAVProxy/modules/mavproxy_console.py
```diff
@@ -91,11 +91,23 @@ def __init__(self, mpstate):
...
+            ]
+            
+            # wsproto is not installed by default.
+            # Only add the menu if it's available.
+            try:
+                import wsproto
+            except ImportError:
+                pass
+            else:
+                mavproxy_items.append(MPMenuItem('Start Websocket Server', 'Start Websocket Server', '# output add wsserver:0.0.0.0:56781'))
```
peterbarker, line 108: Wait.  Why are we adding this if the `ImportError` triggers?

```suggestion
            try:
                import wsproto
                mavproxy_items.append(MPMenuItem('Start Websocket Server', 'Start Websocket Server', '# output add wsserver:0.0.0.0:56781'))
            except ImportError:
                pass
```

... please make sure this code works with/without the Python package present.

peterbarker (changes requested):

MAVProxy/modules/mavproxy_console.py
Ryanf55: The video is proof of it working

peterbarker: Ah, my bad.  Read "else" as "finally" or whatever...

I don't *really* see the point of `else` here - just seems to spread the code out unnecessarily.

*shrug*

peterbarker (changes requested):
Please move to the "Link" menu.

Ryanf55:
> Please move to the "Link" menu.

@tridge Are you ok with Link menu? You originally recommended we not put it there, and instead in the MAVProxy menu.

Ryanf55:
> > Please move to the "Link" menu.
> 
> @tridge Are you ok with Link menu? You originally recommended we not put it there, and instead in the MAVProxy menu.

Tridge is not heavily tied to the MAVProxy menu, and said either option is OK. `Links` doesn't yet do outputs. Another option is to customize the menu through the mavinit file, but that doesn't yet support a user-dialog popup to customize the port, or conditionally load depending if the dependencies are available.

Ryanf55:
[Screencast from 11-14-2024 09:25:41 PM.webm](https://github.com/user-attachments/assets/dc037704-f878-4f27-8dc2-9f1f010e771e)

Done! Here's the new demo.

peterbarker:
Merged, thanks!

PR #1489, opened 2024-11-24
modules: create Cockpit module

Williangalvani:
I'll probably make this either download things dynamically or include it was a submodule, but meanwhile, this works as a proof of concept while I get the other dependencies merged 🙂


peterbarker:
Screenshots?

PR #1494, opened 2024-12-15
modules: add sanity checking to bitmask index when setting bitmask parameter

joshanne:
Sanity check bitmask index when setting bitmask


peterbarker (commented):
LGTM

joshanne:
fixed flake8 failures.

PR #1496, opened 2024-12-16
mavproxy.py: Clean up log_writer exit

### LINKED ISSUE #1414: MAVproxy >= 1.8.66 causes exception in thread log_writer on exit
Platform: RaspberryPi, aarch64, kernel 6.1.21-v8+
Python 3.11.2
numpy 2.0.0 (also tested with 1.26.4 --> same error)

Starting mavproxy script with either
`python3 mavproxy.py --master=/dev/serial0 --baudrate=921600`,
or
`python3 mavproxy.py --master=/dev/serial0 --baudrate=921600 --cmd='set requireexit True'`,
respectively.

When trying to exit the script with either `Ctrl-C` or entering `exit`, respectively, MAVproxy version >= 1.8.66 is throwing the following exception:
```
Exception in thread log_writer:
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
```
and
```
Exception in thread log_writer:
Fatal Python error: _enter_buffered_busy: could not acquire lock for <_io.BufferedWriter name='<stderr>'> at interpreter shutdown, possibly due to daemon threads
Python runtime state: finalizing (tstate=0x0057a4c8)

Current thread 0xf7b57040 (most recent call first):
  <no Python frame>

Extension modules: numpy._core._multiarray_umath, numpy._core._multiarray_tests, numpy.linalg._umath_linalg (total: 3)
```

There is no exception when using MAVproxy version <= 1.8.62
(Versions available via pip install: ..., 1.8.62, 1.8.66, 1.8.67, 1.8.68, 1.8.69, 1.8.70)

Ryanf55:
# Purpose

Closes #1414 

# Details

* The log thread was marked as [daemon=True](https://docs.python.org/3/library/threading.html#threading.Thread.daemon)
* daemon threads are [not recommended anymore by the python maintainers](https://discuss.python.org/t/getting-rid-of-daemon-threads/68836)
* The threading library recommends [using Event for thread synchronization](https://docs.python.org/3/library/threading.html#thread-objects)
* The [Queue.get() call blocks by default](https://docs.python.org/3/library/queue.html#queue.Queue.get), which may was the reason this was put as a daemon thread
* By making [Queue.get()](https://docs.python.org/3/library/queue.html#queue.Queue.get) non-blocking, and checking size before access, because it raises an exception if there is nothing to read, the log writing cleanup can continue as normal. Without this, `Queue..get` can block forever. When daemon=False, this means the thread stays hung and mavproxy is never closed unless you sigterm it with a 2nd ctrl+C (nasty)
* There are no longer exceptions when using ctrl+C to stop MAVProxy
* I use an [Event](https://docs.python.org/3/library/threading.html#threading.Event) as a signaling mechanism to stop the open threads

# Tests Performed

Linux:
- [x] Start mavproxy by itself (no vehicle), and kill it 
- [x] Start mavproxy with a vehicle, ensure it exits cleanly
- [x] Start mavproxy and observe logs are not corrupt
- [x] Test noninteractive mode
- [x] Test `exit` command : `MANUAL> Unknown command 'exit'` => exit doesn't work anymore, even on `master`
- [ ] Test lack of disk space exits mavproxy
- [x] Test permissions error opening logfile has mavproxy exit cleanly -> Mavproxy shows "AP: PreArm: Logging failed" in console, and keeps running. I assume that's ok

MacOS:
- [ ] Queue imports work in [this workaround](https://github.com/ArduPilot/MAVProxy/blob/5127953d3ed929a8643cfebbf7961a82e448e332/MAVProxy/modules/lib/multiproc.py#L70)
- [x] Start mavproxy by itself (no vehicle), and kill it 
- [ ] Start mavproxy with a vehicle, ensure it exits cleanly
- [ ] Start mavproxy and observe logs are not corrupt
- [ ] Test noninteractive mode
- [ ] Test `exit` command
- [ ] Test lack of disk space exits mavproxy
- [ ] Test permissions error opening logfile has mavproxy exit cleanly

Windows
- [ ] Start mavproxy by itself (no vehicle), and kill it 
- [ ] Start mavproxy with a vehicle, ensure it exits cleanly
- [ ] Start mavproxy and observe logs are not corrupt
- [ ] Test noninteractive mode
- [ ] Test `exit` command
- [ ] Test lack of disk space exits mavproxy
- [ ] Test permissions error opening logfile has mavproxy exit cleanly

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -81,7 +81,7 @@ class MPStatus(object):
         self.mav_error = 0
         self.altitude = 0
         self.last_distance_announce = 0.0
-        self.exit = False
+        self.stop_event = threading.Event()
         self.flightmode = 'MAV'
         self.last_mode_announce = 0
         self.last_mode_announced = 'MAV'
@@ -802,7 +802,7 @@ def process_stdin(line):
             print("%-15s : %s" % (cmd, help))
         return
     if cmd == 'exit' and mpstate.settings.requireexit:
-        mpstate.status.exit = True
+        mpstate.status.stop_event.set()
         return
 
     if cmd not in command_map:
@@ -935,8 +935,13 @@ def mkdir_p(dir):
 
 def log_writer():
     '''log writing thread'''
-    while not mpstate.status.exit:
-        mpstate.logfile_raw.write(bytearray(mpstate.logqueue_raw.get()))
+    while not mpstate.status.stop_event.is_set():
+        if not mpstate.logqueue_raw.empty():
+            bytes = mpstate.logqueue_raw.get(block=False)
+            mpstate.logfile_raw.write(bytearray(bytes))
+        time.sleep(0.001)
+
+        # TODO consider wait() the stop event instead
         timeout = time.time() + 10
         while not mpstate.logqueue_raw.empty() and time.time() < timeout:
             mpstate.logfile_raw.write(mpstate.logqueue_raw.get())
@@ -1008,18 +1013,17 @@ def open_telemetry_logs(logpath_telem, logpath_telem_raw):
             stat = os.statvfs(logpath_telem)
             if stat.f_bfree*stat.f_bsize < 209715200:
                 print("ERROR: Not enough free disk space for logfile")
-                mpstate.status.exit = True
+                mpstate.status.stop_event.set()
                 return
 
         # use a separate thread for writing to the logfile to prevent
         # delays during disk writes (important as delays can be long if camera
         # app is running)
         t = threading.Thread(target=log_writer, name='log_writer')
-        t.daemon = True
         t.start()
     except Exception as e:
         print("ERROR: opening log file for writing: %s" % e)
-        mpstate.status.exit = True
+        mpstate.status.stop_event.set()
         return
 
 
@@ -1120,7 +1124,7 @@ def main_loop():
         set_stream_rates()
 
     while True:
-        if mpstate is None or mpstate.status.exit:
+        if mpstate is None or mpstate.status.stop_event.is_set():
             return
 
         # enable or disable screensaver:
@@ -1220,12 +1224,12 @@ def main_loop():
 
 def input_loop():
     '''wait for user input'''
-    while mpstate.status.exit is not True:
+    while not mpstate.status.stop_event.is_set():
         try:
             line = mpstate.rl.input()
             mpstate.input_queue.put(line)
         except (EOFError, IOError):
-            mpstate.status.exit = True
+            mpstate.status.stop_event.set()
 
 
 def run_script(scriptfile):
@@ -1403,7 +1407,6 @@ if __name__ == '__main__':
 
     # global mavproxy state
     mpstate = MPState()
-    mpstate.status.exit = False
     mpstate.command_map = command_map
     mpstate.continue_mode = opts.continue_mode
     # queues for logging
@@ -1443,11 +1446,11 @@ if __name__ == '__main__':
 
     def quit_handler(signum=None, frame=None):
         # print('Signal handler called with signal', signum)
-        if mpstate.status.exit:
+        if mpstate.status.stop_event.is_set():
             print('Clean shutdown impossible, forcing an exit')
             sys.exit(0)
         else:
-            mpstate.status.exit = True
+            mpstate.status.stop_event.set()
 
     # Listen for kill signals to cleanly shutdown modules
     fatalsignals = [signal.SIGTERM]
@@ -1586,7 +1589,7 @@ if __name__ == '__main__':
 
     # use main program for input. This ensures the terminal cleans
     # up on exit
-    while (mpstate.status.exit is not True):
+    while not mpstate.status.stop_event.is_set():
         try:
             if opts.daemon or opts.non_interactive:
                 time.sleep(0.1)
@@ -1608,7 +1611,7 @@ if __name__ == '__main__':
                         m.init(mpstate)
 
             else:
-                mpstate.status.exit = True
+                mpstate.status.stop_event.set()
                 sys.exit(1)
 
     if opts.profile:


Huibean:
queue imports is fine on macos, can you make a to do list like linux with mac

Ryanf55:
> queue imports is fine on macos, can you make a to do list like linux with mac

Added!

MAVProxy/mavproxy.py
```diff
@@ -936,8 +940,16 @@ def mkdir_p(dir):
...
-    while not mpstate.status.exit:
-        mpstate.logfile_raw.write(bytearray(mpstate.logqueue_raw.get()))
+    print("Starting logs thread...")
+    while not mpstate.status.stop_event.is_set():
+        try:
+            bytes = mpstate.logqueue_raw.get(block=False)
+        except queue.Empty:
+            pass
+        else:
+            mpstate.logfile_raw.write(bytearray(bytes))
```
peterbarker, line 950: Why make this change as part of this PR?

I'm not sure it's correct - `multiproc.Queue` doesn't necessarily throw `queue.Empty`.

peterbarker (changes requested):

MAVProxy/mavproxy.py
Ryanf55: Easier approach - check `empty()` before access. Less pythonic, but more portable code.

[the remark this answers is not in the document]
Ryanf55: [pep8](https://peps.python.org/pep-0008/) ` limit the try clause to the absolute minimum amount of code necessary.`

Ryanf55: Actually: https://devguide.python.org/versions/
https://github.com/ArduPilot/MAVProxy/blob/e45946ae2b1a06e662f1cc5c176e2d07eab1e12d/MAVProxy/modules/lib/multiproc.py#L66

python3.7 is EOL, therefore the whole premise of PipeQueue is EOL based on the if statement.

peterbarker: We don't kill user's setups just because the Python is not supported by the Python project.  Look at how long we supported Python2....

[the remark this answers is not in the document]
peterbarker: I'm not a great fan of the `else` construct used here is the point.

Scoping-wise it is awkward having a variable assigned to in the `try` block be valid in the `else` block - and it separates the assignment from the use.

MAVProxy/mavproxy.py
```diff
@@ -1012,11 +1025,10 @@ def open_telemetry_logs(logpath_telem, logpath_telem_raw):
                 mpstate.status.exit = True
                 return
 
-        # use a separate thread for writing to the logfile to prevent
+        # Use a separate thread for writing to the logfile to prevent
```
peterbarker, line 1028: Why would you do this?

Every time I come back in to review this PR I have to look through these all the time.  The patch adds no value to the stated goal of the PR and wastes my time.

MAVProxy/mavproxy.py
```diff
@@ -1584,7 +1598,7 @@ def quit_handler(signum=None, frame=None):
 
     # use main program for input. This ensures the terminal cleans
     # up on exit
-    while (mpstate.status.exit is not True):
+    while (not mpstate.status.stop_event.is_set()):
```
peterbarker, line 1601:
```suggestion
    while not mpstate.status.stop_event.is_set():
```

peterbarker (changes requested):

Ryanf55:
> queue imports is fine on macos, can you make a to do list like linux with mac

Can you please test this PR on your mac?

peterbarker:
... lack of disk space should definitely not cause MAVProxy to exit... that would be a serious bug if it did

peterbarker:
This looks good to me.  I'll test it this weekend.

Huibean:
> > queue imports is fine on macos, can you make a to do list like linux with mac
> 
> Can you please test this PR on your mac?

Yes, I will give feeback later

Ryanf55:
> ... lack of disk space should definitely not cause MAVProxy to exit... that would be a serious bug if it did

It's a feature, not a bug:)
https://github.com/ArduPilot/MAVProxy/blob/b39a31cd38f03dcd49d33aabbe361775367611a3/MAVProxy/mavproxy.py#L1006-L1013

Thank you both, much appreciated.

peterbarker:
> > ... lack of disk space should definitely not cause MAVProxy to exit... that would be a serious bug if it did
> 
> It's a feature, not a bug:)
> 
> https://github.com/ArduPilot/MAVProxy/blob/b39a31cd38f03dcd49d33aabbe361775367611a3/MAVProxy/mavproxy.py#L1006-L1013
> 
> Thank you both, much appreciated.

OK, that's on startup.  That's OK.

Huibean:
```
MANUAL> Got command: exit
mpstate.settings.requireexit False
Unknown command 'exit'
```
exit command still shows unknow in mac

Ryanf55:
> ```
> MANUAL> Got command: exit
> mpstate.settings.requireexit False
> Unknown command 'exit'
> ```
> 
> exit command still shows unknow in mac

Yea, exit is broke on master too, not a regression causes by this PR.

peterbarker:
- [ ] check CPU usage on all three platforms
 - [ ] fix performance regression on Linux with 100% CPU on one core

Ryanf55:
The CPU wait is caused by the now hot loop in the log writer exit because it's no longer blocking.

In [Queue.get()](https://docs.python.org/3/library/queue.html#queue.Queue.get), we're having to deal with `Prior to 3.0 on POSIX systems, and for all versions on Windows, if block is true and timeout is None, this operation goes into an uninterruptible wait on an underlying lock. This means that no exceptions can occur, and in particular a SIGINT will not trigger a KeyboardInterrupt.`. 

The API for the multiproc queue in MAVproxy does not support the same arguments, but if it did, we could set a timeout time, and catch the timeout exception to reduce the loop rate. 

I could also add a small delay in the loop, which will be easier.

Ryanf55:
I fixed the hot loop with the simplest option. a 1ms delay. Here's `htop` results, filtered for mavproxy on my desktop:
image

MAVProxy/mavproxy.py
```diff
@@ -936,8 +936,13 @@ def mkdir_p(dir):
 
 def log_writer():
     '''log writing thread'''
-    while not mpstate.status.exit:
-        mpstate.logfile_raw.write(bytearray(mpstate.logqueue_raw.get()))
+    while not mpstate.status.stop_event.is_set():
+        if not mpstate.logqueue_raw.empty():
```
peterbarker, line 940: We still appear to be looking at the raw queue later in this function - conflict resolution error?

Ryanf55: The event is the exit event. This statement should read

```
while we are not supposed to exit:
   use the queue for logging operations
 ```
 
 Perhaps you missed the `not` when reading the code?  It's the same behavior as before. Run logging if we aren't exiting.

peterbarker: It's the code on line 942/947 I was looking at.

Ryanf55: The reason we look at the raw queue later is to flush it when the program is supposed to stop.

PR #1499, opened 2024-12-17
mavproxy_mode: Allow for mode names to be tab completed

joshanne:
Adds tab completion of mode names when changing mode using the command `mode <MODE>`.

Also prettifies printing of available modes - this previously printed an array of  `dict_keys([..], ...)`


peterbarker (approved):
LGTM

PR #1500, opened 2025-01-09
mavproxy_soar: Add initial soaring display

Ryanf55:
# Purpose

Add an initial soaring module, which displays the location and estimated radius of the thermal from AP's thermal soaring estimator.  
image

Once we have a mavlink packet definition for the thermal estimator, we can move to that from NAMED_VALUE_FLOAT.

# Demo

Enable the NVT for the EKF and run the soaring autotest without speedup.

Terminal 1:
```bash
./Tools/autotest/autotest.py build.Plane test.Plane.Soaring --map --speedup=1 --waf-configure-args "--define HAL_SOARING_NVF_EKF_ENABLED=1"
```

Terminal 2:
```bash
mavproxy.py --console --master tcp:127.0.0.1:5763 --map
module load soar
```
Once the vehicle hits the thermal southwest of home, it will show up on the map. You can see the estimator converge on the thermal.

### DIFF
--- a/MAVProxy/modules/mavproxy_soar.py
+++ b/MAVProxy/modules/mavproxy_soar.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+'''
+Soaring Module
+Ryan Friedman
+
+This module displays the estimated soaring thermals on the map.
+A circle is drawn with the estimated radius from ArduSoar's estimated thermal location.
+
+Note that this module uses NAMED_VALUE_FLOAT messages from the autopilot,
+a temporary measure until a proper mavlink message is decided upon.
+
+AP_FLAKE8_CLEAN
+'''
+
+from MAVProxy.modules.lib import mp_module, mp_util
+from MAVProxy.modules.mavproxy_map import mp_slipmap
+
+import time
+
+
+class soar(mp_module.MPModule):
+    _SOAR_THERMAL_MAP_OBJ_ID = "soar-thermal"
+    _CLEAR_STALE_THERMAL_TIME = 5.0
+
+    def __init__(self, mpstate):
+        """Initialise module"""
+        super(soar, self).__init__(mpstate, "soar", "")
+        self._strength = None
+        self._radius = None
+        self._x = None
+        self._y = None
+        self._last_draw_time = time.time()
+
+    def mavlink_packet(self, m):
+        '''handle mavlink packets'''
+
+        if m.get_type() == 'NAMED_VALUE_FLOAT' and m.name.startswith("SOAR"):
+            if m.name == "SOAREKFX0":
+                self._strength = m.value
+            elif m.name == "SOAREKFX1":
+                self._radius = m.value
+            elif m.name == "SOAREKFX2":
+                self._x = m.value
+            elif m.name == "SOAREKFX3":
+                self._y = m.value
+            else:
+                raise NotImplementedError(m.name)
+
+            self.draw_thermal_estimate()
+
+    def idle_task(self):
+        '''called rapidly by mavproxy'''
+
+        if time.time() - self._last_draw_time > self._CLEAR_STALE_THERMAL_TIME:
+            self.clear_thermal_estimate()
+
+    def draw_thermal_estimate(self):
+
+        if self._radius is None:
+            return
+        if self._x is None:
+            return
+        if self._y is None:
+            return
+
+        wp_module = self.module('wp')
+        if wp_module is None:
+            return
+        home = wp_module.get_home()
+        if home is None:
+            return
+
+        home_lat = home.x
+        home_lng = home.y
+
+        (thermal_lat, thermal_lon) = mp_util.gps_offset(home_lat, home_lng, self._y, self._x)
+
+        slipcircle = mp_slipmap.SlipCircle(
+            self._SOAR_THERMAL_MAP_OBJ_ID, # key
+            "thermals", # layer
+            (thermal_lat, thermal_lon), # latlon
+            self._radius, # radius
+            (0, 255, 255),
+            linewidth=2)
+        for mp in self.module_matching('map*'):
+            self._last_draw_time = time.time()
+            mp.map.add_object(slipcircle)
+
+    def clear_thermal_estimate(self):
+        for mp in self.module_matching('map*'):
+            mp.map.remove_object(self._SOAR_THERMAL_MAP_OBJ_ID)
+
+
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,81 @@
...
+            elif m.name == "SOAREKFX1":
+                self._radius = m.value
+            elif m.name == "SOAREKFX2":
+                self._x = m.value
+            elif m.name == "SOAREKFX3":
+                self._y = m.value
+            elif m.name == "SOARECX":
+                pass
+            elif m.name == "SOARECY":
+                pass
```
Ryanf55, line 37:
```suggestion
```

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+'''
+Soaring Module
+Ryan Friedman
+
+This module displays the estimated soaring thermals on the map.
+A circle is drawn with the estimated radius from ArduSoar's estimated thermal location.
+
```
peterbarker, line 8:
```suggestion

Note that this module uses NAMED_VALUE_FLOAT messages from the uatopilot, a temporary measure until a proper mavlink message is decided upon.

```

... so others taking inspiration from the module can't get too upset when we break these messages...

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,72 @@
...
+                self._y = m.value
+            else:
+                raise NotImplementedError(m.name)
+
+            self.draw_thermal_estimate()
+
+    def draw_thermal_estimate(self):
+
+        if self._radius is None:
+            print("No radius")
```
peterbarker, line 41: Maybe silence these?

Not a blocker...

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,72 @@
...
+            return
+        if self._x is None:
+            print("No x")
+            return
+        if self._y is None:
+            print("No y")
+            return
+
+        home = self.module('wp').get_home()
```
peterbarker, line 50: This module might not be loaded.

peterbarker: .... also, are these thermals really home-relative or are they navigation-origin-relative?

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,72 @@
...
+from MAVProxy.modules.lib import mp_module, mp_util
+from MAVProxy.modules.mavproxy_map import mp_slipmap
+
+
+class soar(mp_module.MPModule):
+    def __init__(self, mpstate):
+        """Initialise module"""
+        super(soar, self).__init__(mpstate, "soar", "")
+
+    def mavlink_packet(self, m):
```
peterbarker, line 21: I sgugest dropping a timeout in the top of here to clear the circle if you haven't got a message in a while.

Ryanf55: Nice, yea good idea. Added!

MAVProxy/modules/mavproxy_soar.py
Ryanf55: > .... also, are these thermals really home-relative or are they navigation-origin-relative?

Per `SoaringController::update_thermalling`:

```c++
    const AP_AHRS &_ahrs = AP::ahrs();
    if (!_ahrs.get_relative_position_NED_home(current_position)) {
        return;
    }
    ...
    
    // update the filter
    _ekf.update(_vario.reading, current_position.x, current_position.y, wind_drift.x, wind_drift.y);
```

Based on that, I take it as home-relative. Not great, but that's how it is today. Seems like moving home would break stuff. Anyways, yea, I currently draw them relative to home to match.

Ryanf55:
Added the clearing stale thermals! Here's a brief recording of it working.
[MavproxySoarThermalClear01-13-2025 10:56:47 PM.webm](https://github.com/user-attachments/assets/bdcc317d-a132-4709-9f29-43e4d0be3b3f)

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,98 @@
...
+    def draw_thermal_estimate(self):
+
+        if self._radius is None:
+            return
+        if self._x is None:
+            return
+        if self._y is None:
+            return
+
+        if not (wp_module := self.module('wp')):
```
peterbarker, line 66: Not a syntax used *anywhere* else in MAVProxy...

... also, need to check for `None`, not truthiness.

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,98 @@
...
+            return
+        if (home := wp_module.get_home()) is None:
```
peterbarker, line 68: .... really, just assign then check...

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,98 @@
...
+            return
+        
+        home_lat = home.x
+        home_lng = home.y
+
+        (thermal_lat, thermal_lon) = mp_util.gps_offset(home_lat, home_lng, self._y, self._x)
+
+
```
peterbarker, line 76:
```suggestion
```

MAVProxy/modules/mavproxy_soar.py
```diff
@@ -0,0 +1,98 @@
...
+        slipcircle = mp_slipmap.SlipCircle(
+            self._SOAR_THERMAL_MAP_OBJ_ID, # key
+            "thermals", # layer
+            (thermal_lat, thermal_lon), # latlon
+            self._radius, # radius
+            (0, 255, 255),
+            linewidth=2)
+        for mp in self.module_matching('map*'):
+            self._last_draw_time = time.time()
```
peterbarker, line 86: I was thinking timeout might be based on message retrieval time rather than draw time.

That way when the vehicle loses the thermal the GCS gets updated to remove the thermal hint

MAVProxy/modules/mavproxy_soar.py
Ryanf55: Not a fan of the walrus operator eh? Fair enough, removed

peterbarker (approved):
LGTM

PR #1502, opened 2025-01-20
RC: Make RC GUI fit according to max received value

stephendade:
Fixes graphical bug in RC and Servo GUI when the PWM value is > 2000


peterbarker:
Merged, thanks

PR #1508, opened 2025-02-19
Add terrain contours to map

srmainwaring:
Use data available in the terrain module to overlay terrain contours on the map.

mavproxy_contour_ardnamurchan

## Motivation

- Be able to overlay terrain contour lines on the existing slip map.
- Use existing terrain data sources to match information available on the FC.
- Initial step in adding support to mavproxy for configuring off-board terrain planning for planes. The idea is to replace the dependency on rviz and simplify the user workflow the example used here https://discuss.ardupilot.org/t/offboard-terrain-navigation-for-plane-with-ros-2-ap-dds/114418. 

## Use

Select the number of contour levels, grid resolution and grid extent using the map settings:

```bash
STABILIZE> map set
STABILIZE>           
          brightness 1
    circle_linewidth 1
 contour_grid_extent 20000.0
contour_grid_spacing 30.0
      contour_levels 20
           font_size 0.5
        loitercircle False
         rallycircle False
     setpos_accuracy 50
        showahrs2pos 0
        showahrs3pos 0
         showahrspos 1
       showclicktime 2
       showdirection False
         showgps2pos 1
          showgpspos 1
          showsimpos 0
           showwpnum True
```

The terrain resolution is determined by the source used. The default is `SRTM3` which may be switched before loading the contours using:

```bash
STABILIZE> terrain set source SRTM1
```

To show the contours use the context menu:

terrain_contours_show

Hide the contours (but do not remove - saves recalculating when showing again):

terrain_contours_hide

To select a different region, remove then show again: 

terrain_contours_move

## Tasks

- [x] add menu item to toggle contour visibility
- [x] add settings to control number of levels (and/or spacing)

## Follow up / optional

- [ ] add settings to control the contour colour map
- [ ] add contour labels
- [ ] replace dependency on [matplotlib.pyplot.contour](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.contour.html) with [ContourPy](https://contourpy.readthedocs.io/en/v1.3.1/index.html)
- [ ] automatically update / calculate contours as the map is moved / zoomed as it's currently fixed
- [ ] shade areas above a certain altitude

### DIFF
--- a/MAVProxy/modules/mavproxy_map/__init__.py
+++ b/MAVProxy/modules/mavproxy_map/__init__.py
@@ -27,6 +27,7 @@ class MapModule(mp_module.MPModule):
         self.wp_change_time = 0
         self.fence_change_time = 0
         self.rally_change_time = 0
+        self.terrain_contour_ids = []
         self.have_simstate = False
         self.have_vehicle = {}
         self.move_wp = -1
@@ -65,8 +66,12 @@ class MapModule(mp_module.MPModule):
               ('showdirection', bool, False),
               ('setpos_accuracy', float, 50),
               ('mission_color', str, "white"),
-              ('font_size', float, 0.5) ])
-
+              ('font_size', float, 0.5),
+              ('contour_levels', int, 20),
+              ('contour_grid_spacing', float, 30.0),
+              ('contour_grid_extent', float, 20000.0),
+            ])
+        
         service='MicrosoftHyb'
         if 'MAP_SERVICE' in os.environ:
             service = os.environ['MAP_SERVICE']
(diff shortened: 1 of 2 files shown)


peterbarker:
This is very cool, and probably useful as-is (ie. mergeable IMO).

Perhaps a better approach might be to modify the things we store in the tilecache so that they can have "overlays" in them.  That will mean not need to recalculate - and essnetially no redraw cost for the layers.

srmainwaring:
@peterbarker - thanks for the feedback.

Could move the contour polygon storage to the `MPTile._tile_cache`, but that would involve a bigger change, and I'd probably want to talk through an approach before taking that on.

The current implementation calculates the contour polygons once for a fixed area at a resolution determined in the settings. The polygons are redrawn as the map is zoomed and scrolled, but the more expensive calculation from the terrain data only occurs if the contours are removed then reshown. I figured this was a reasonable compromise, and shows the contours around a ROI without covering the entire map.

The one change I have wondered about is moving the calculation out of the map module into a separate one (similar to Ryan's soar module). That could be easily done as a follow up if the logic to support the terrain contours started getting more complicated or we wanted to keep it compartmentalised.

peterbarker:
Merged, thanks!

PR #1510, opened 2025-02-26
mavproxy_link.py: fix websocket output sending mavlink pkt during hanshake

Huibean:
``` shell
wsproto.utilities.LocalProtocolError: Event BytesMessage(data=bytearray(b'\xfd\x0f\x00\x00$\x01\x01\x82\x01\x00}\x04\x04\x98\xff\x00\x00\x07\xcb\x92\xb8G\x00\x00\xcc\xd5\xb4'), frame_finished=True, message_finished=True) cannot be sent during the handshake
```
fix raised error when sending mavlink packet during websocket handshake

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -1054,6 +1054,10 @@ class LinkModule(mp_module.MPModule):
             if self.mpstate.settings.mavfwd_rate or mtype != 'REQUEST_DATA_STREAM':
                 if mtype not in self.no_fwd_types:
                     for r in self.mpstate.mav_outputs:
+                        if hasattr(r, 'ws') and r.ws is not None:
+                            from wsproto.connection import ConnectionState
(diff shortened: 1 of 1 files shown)


tpwrules:
Wouldn't this be better to do in the write method (which I assume is in mavproxy)?

Huibean:
> Wouldn't this be better to do in the write method (which I assume is in mavproxy)?

Hi @tpwrules ,thanks for you remind, write seems in pymavlink, I am looking forward if I can locate it

Huibean:
I think better stay current commit, overwrite the write function could blocking mavproxy sending data through handshake

tpwrules:
How about the open function? I think it makes sense to wait until the handshake is complete there. That is how normal TCP sockets work right?

peterbarker:
I've merged this one.  There might be a better way to do this, but the effects of *this* change can clearly be seen

Thanks @Huibean @tpwrules

PR #1515, opened 2025-02-28
mavproxy_joystick: add support for toggle buttons

heeplr:
Since there usually are no persistent switches on gamepads but just momentary switches, I added the "toggle" control to simulate persistent switches.
This allows to e.g. arm/disarm using a button.

`toggle` takes a list of values that are cycled through on every button press/release.

By using a list for `values`, not only on/off switches can be simulated but also multi-position or rotary switches as well.

### DIFF
--- a/MAVProxy/modules/mavproxy_joystick/controls.py
+++ b/MAVProxy/modules/mavproxy_joystick/controls.py
@@ -16,12 +16,14 @@ class Control (object):
     '''Base class for all controls'''
     def __init__(self, joystick,
                  inlow=-1, inhigh=1,
-                 outlow=1000, outhigh=2000):
+                 outlow=1000, outhigh=2000,
+                 values=[]):
         self.joystick = joystick
         self.inlow = inlow
         self.inhigh = inhigh
         self.outlow = outlow
         self.outhigh = outhigh
+        self.values = values
 
 
 class Button (Control):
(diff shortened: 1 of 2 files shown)


heeplr:
I can't make a lot of the Windows build fail. If anyone sees why it's failing, please notify me if it's related to the patch.

peterbarker:
> I can't make a lot of the Windows build fail. If anyone sees why it's failing, please notify me if it's related to the patch.

This isn't your bug :-)

MAVProxy/modules/mavproxy_joystick/controls.py
```diff
@@ -42,6 +44,35 @@ def value(self):
...
+    '''A ToggleButton acts like a toggle or rotary switch.
+    Initially, the corresponding channel value is set to the first value in `values`;
+    when button `id` is pressed consecutively, the values are iterated on each press.
+    Iteration is looped by starting from the beginning when last value is reached.'''
+
+    def __init__(self, joystick, id, **kwargs):
+        super(ToggleButton, self).__init__(joystick, **kwargs)
+        self.id = id
+        # index of value to set next
+        self._current_value = 0
```
peterbarker, line 57:
```suggestion
        self._current_value_index = 0
```

peterbarker (changes requested):
This looks really good.

But I do think we need to re-name this variable, sorry!

MAVProxy/modules/mavproxy_joystick/controls.py
heeplr: Right. Changed it.

peterbarker:
Merged, thanks!

PR #1537, opened 2025-03-26
Added yaw value in console output

Safe-bet:
I have added yaw just like pitch and roll, using the attitude handle function and adding msg.yaw in it.


MAVProxy/modules/mavproxy_console.py
```diff
@@ -66,6 +66,7 @@ def __init__(self, mpstate):
         mpstate.console.set_status('Thr', 'Thr ---', row=2)
         mpstate.console.set_status('Roll', 'Roll ---', row=2)
         mpstate.console.set_status('Pitch', 'Pitch ---', row=2)
+        mpstate.console.set_status('Yaw', 'Yaw ---', row=2)
```
tridge, line 69: we already show 'Hdg' for heading, it should have the same information

Safe-bet: Yes, you are right, Hdg and Yaw are similar, just that as Hdg isn't in degrees, so I added yaw specifically for it.

Issue #1538, opened 2025-04-01
Dataratelogging creates different lostprecent


mjohenneken:
With `link dataratelogging on` we ca write a logfile to debug telementry link issues. However the values saved in the .csv differ from the loss precent shown in `link`. 

Example: link reports 1.7% but the file has 177%

I belive the conversion in this line is wrong:

https://github.com/ArduPilot/MAVProxy/blob/6eaee8f1a123b1cea77c099f7193864976d057f9/MAVProxy/modules/mavproxy_link.py#L157C35-L157C84

It should not be muliplied by 100.

In case you want to fix exsting datarate logs, i wrote this utility script:

```
import csv
import sys
from pathlib import Path

def fix_dataratelog(filepath):
    # Create a backup of original file
    filepath = Path(filepath)
    backup_path = filepath.with_suffix('.csv.bak')
    filepath.rename(backup_path)
    
    # Read from backup, write to new file
    with open(backup_path, 'r') as infile, open(filepath, 'w', newline='') as outfile:
        reader = csv.reader(infile)
        writer = csv.writer(outfile)
        
        # Write header
        header = next(reader)
        writer.writerow(header)
        
        # Process each row
        for row in reader:
            # Convert last column to float, divide by 100, convert back to string
            row[-1] = str(float(row[-1]) / 100)
            writer.writerow(row)
            
    print(f"Fixed {filepath}")
    print(f"Backup saved as {backup_path}")

if __name__ == "__main__":
    if len(sys.argv) != 2:
        print("Usage: python fix_dataratelog.py path/to/dataratelog.csv")
        sys.exit(1)
        
    fix_dataratelog(sys.argv[1]) 
```

PR #1539, opened 2025-04-01
Update mavproxy_link.py

mjohenneken:
Fix conversion see #1538


peterbarker:
@stephendade - your logging format, I think :-)

stephendade:
Thanks for catching that bug. Your solution look good. Could you change your commit message to "Link: Correct dataratelogging format"?

CI failure appears to be unrelated.

stephendade:
Thanks! Merging...

PR #1549, opened 2025-04-12
mavproxy_param: allow setting multiple bitmask bits simultaneously

be-ska:
Sometimes it's very handy to set multiple bits with just one command, ie: setting ARMING_CHECK or LOG_BITMASK.
With this very simple PR user can provide more bit arguments, so
```
param bitmask toggle ARMING_CHECK 0 1 2 3 4 5 6 7 8 10 11 12 13 14 15 18 19
```
change ARMING_CHECK from ALL (=1) to ALL but CAMERA (=982526)

### DIFF
--- a/MAVProxy/modules/mavproxy_param.py
+++ b/MAVProxy/modules/mavproxy_param.py
@@ -544,7 +544,7 @@ class ParamState:
         # Ensure we have at least an action and a parameter
         if len(args) < 2:
             print("Not enough arguments")
-            print(f"param bitmask <{'/'.join(BITMASK_ACTIONS)}> <parameter> [bit-index]")
+            print(f"param bitmask <{'/'.join(BITMASK_ACTIONS)}> <parameter> [bit-index-1 ... bit-index-n]")
             return
 
         action = args[0]
@@ -585,17 +585,19 @@ class ParamState:
             return
 
         # The next argument is the bit_index - if it exists, handle it
-        bit_index = None
-        if len(args) >= 3:
+        bit_indices = []
+        while len(args) >= 3:
             try:
                 # If the bit index is available lets grab it
                 arg_bit_index = args[2]
+                # Remove the bit index from the args list
+                args.pop(2)
                 # Try to convert it to int
-                bit_index = int(arg_bit_index)
+                bit_indices.append(int(arg_bit_index))
             except ValueError:
-                print(f"Invalid bit index: {arg_bit_index}\n")
-
-        if bit_index is None:
+                print(f"Invalid bit index: {arg_bit_index}")
+    
+        if bit_indices == []:
             # No bit index was specified, but the parameter and action was.
             # Print the available bitmask information.
             print("%s: %s" % (uname, phelp.get('humanName')))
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_param.py
```diff
@@ -585,17 +585,19 @@ def param_bitmask_modify(self, master, args):
...
+                args.pop(2)
                 # Try to convert it to int
-                bit_index = int(arg_bit_index)
+                bit_indices.append(int(arg_bit_index))
             except ValueError:
-                print(f"Invalid bit index: {arg_bit_index}\n")
-
-        if bit_index is None:
+                print(f"Invalid bit index: {arg_bit_index}")
+
```
joshanne, line 599: prune the whitespace.

joshanne (commented):
First, I love seeing that people are using this feature already!

I would like to test this and see how it behaves with a vehicle.

joshanne (commented):
@peterbarker only comment I have is above regarding informing of errors.

The functionality appears to work as expected, and doesn't break any workflows with the single bit index.

I feel like there could be a more intuitive way to do the same thing - When i do this, I do each bit at a time, ensuring I don't make a mistake at each step. Not saying it's the perfect way, though.

MAVProxy/modules/mavproxy_param.py
```diff
@@ -607,42 +609,41 @@ def param_bitmask_modify(self, master, args):
...
-        else:
-            # We cannot toggle, set or clear
-            print("Invalid bitmask action")
-            return
+        # Cycle through the bit indices
+        for bit_index in bit_indices:
+            # Sanity check the bit index
+            if bit_index >= NUM_BITS_MAX:
+                print(f"Cannot perform bitmask action '{action}' on bit index {bit_index}.")
+                return
```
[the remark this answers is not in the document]
be-ska: > I personally prefer to see all the failed bits in a single message, rather than have one failure, fix that particular bit, then retry and find i had another invalid bit index.
Agree, I'll add the sanity check loop before performing any action

be-ska:
I've added the suggested changes, I can squash the commits if you approve

joshanne:
@be-ska looks like it's all working fine, i'd be happy if you pruned the whitespace and squash the commits. Really only needs to be the one commit.

@peterbarker leaving you for final approval, but all seems fine to me. Curious if you've made use of the bitmask manipulation yourself?

joshanne (approved):

MAVProxy/modules/mavproxy_param.py
joshanne, line 1: squash commits and prune whitespace.

- [ ]

peterbarker (approved):

peterbarker:
> @peterbarker leaving you for final approval, but all seems fine to me. Curious if you've made use of the bitmask manipulation yourself?

Yep, I'll merge this once CI passes.

And yes, I do use the new command a bit - particularly when combined with the detailed-which-bits-are-set parameter view it's pretty damned handy!

PR #1555, opened 2025-04-16
mavproxy_misc: add support for MAV_CMD_DO_CHANGE_ALTITUDE

peterbarker:
falls back to the mission item way if we get UNSUPPORTED


peterbarker:
@MichelleRos and I flew this.

PR #1558, opened 2025-04-25
mavproxy_arm: try to use mavlink command to set/unset safety switch

peterbarker:
Still works on Copter-4.5.7 which doesn't support the command, falls back appropriately.


peterbarker:
I've been running this since it was authored.

Issue #1361, opened 2024-04-12
Chat: chat module still working even after unloading


adityaomar3:
After unloading the chat module, it still works.

Probable solution: 
define a function "unload" in .__init__ file and collect all the chat threads and stop them and a close function to close chat window.

Issue #1342, opened 2024-03-23
Bug report in chat module on loading, unloading and reloading


adityaomar3:
When we reload or unload and load again the chat module it throws a gtk error and segmentation fault

`(mavproxy.py:13354): Gtk-CRITICAL **: 00:08:18.468: gtk_text_attributes_ref: assertion 'values != NULL' failed

(mavproxy.py:13354): Gtk-CRITICAL **: 00:08:18.468: gtk_text_attributes_ref: assertion 'values != NULL' failed
Segmentation fault (core dumped)`

To reproduce this bug 
Start mavproxy.py
module load chat
module unload chat
module load chat

or 

Start mavproxy.py
module load chat
module reload chat

rmackay9:
Thanks for the report.  I've added this to the [consolidate list of chat module issues and enhancements](https://github.com/ArduPilot/MAVProxy/issues/1281).

PR #1561, opened 2025-04-30
Fix chat module bugs

SriramS-77:
Fixes:
- #1342
- #1361
- A bug under issue 1281:
- > typing "chat show" into the MAVProxy produces an innocuous error. We should either remove that completely or make it re-show the chat GUI in case the user has closed it

Added "chat hide" feature, to interact with existing "chat show" feature. 

### Demo Video:

https://github.com/user-attachments/assets/a1434505-f6a1-4162-87de-1ef789eb4b74

### DIFF
--- a/MAVProxy/modules/mavproxy_chat/__init__.py
+++ b/MAVProxy/modules/mavproxy_chat/__init__.py
@@ -26,7 +26,7 @@ class chat(mp_module.MPModule):
         super(chat, self).__init__(mpstate, "chat", "OpenAI chat support")
 
         # register module and commands
-        self.add_command('chat', self.cmd_chat, "chat module", ["show"])
+        self.add_command('chat', self.cmd_chat, "chat module", ["hide", "show"])
 
         # keep reference to mpstate
         self.mpstate = mpstate
@@ -42,15 +42,18 @@ class chat(mp_module.MPModule):
 
     # create chat window (should be called from a new thread)
     def create_chat_window(self):
+        print("Trying to create chat window")
         if mp_util.has_wxpython:
-            # create chat window
+            # create chat window, and return the chat_window object created
             self.chat_window = chat_window.chat_window(self.mpstate, self.wait_for_command_ack)
+            # Call main loop of chat window
+            self.chat_window.start()
         else:
             print("chat: wx support required")
 
     # show help on command line options
     def usage(self):
-        return "Usage: chat <show>"
+        return "Usage: chat <hide|show>"
 
     # control behaviour of the module
     def cmd_chat(self, args):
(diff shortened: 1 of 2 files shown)


rmackay9:
Hi @SriramS-77,

Thanks very much for this!

Two small administrative things, could you add a "Chat: " prefix to the commit title?  To be clear I mean the commit title not the PR title.  Also if possible could you rebase on master to remove the "Merge branch 'master' into master" commit?

I'll give this a quick test shortly but it looks good, thanks!

MAVProxy/modules/mavproxy_chat/__init__.py
```diff
@@ -42,29 +42,47 @@ def __init__(self, mpstate):
 
     # create chat window (should be called from a new thread)
     def create_chat_window(self):
+        print("Trying to create chat window")
```
rmackay9, line 45: we should remove this debug output

MAVProxy/modules/mavproxy_chat/chat_window.py
```diff
@@ -90,12 +90,13 @@ def __init__(self, mpstate, wait_for_command_ack_fn):
         # show frame
         self.frame.Show()
 
+    def start(self):
```
rmackay9, line 93: Is this change to add the start method necessary?  I think it's only called from one place and it is immediately after the call to __init__ which is where this line was previously.

MAVProxy/modules/mavproxy_chat/__init__.py
```diff
@@ -42,29 +42,47 @@ def __init__(self, mpstate):
...
         self.chat_window.show()
 
+    def hide(self):
+        self.chat_window.hide()
+
+    def close(self):
+        self.chat_window.close()
+
+    # unload function override for module interface
+    def unload(self):
```
rmackay9, line 80: I've tested "module unload chat" and it doesn't seem to work I'm afraid.  Please tell me if I'm doing something wrong.

MAVProxy/modules/mavproxy_chat/chat_window.py
SriramS-77: Yeah, the problem was that this function [app.MainLoop()], does not return till the chat window is closed. So, the chat_window object was never being returned after being created by __init__(). Hence, show, hide and unload were not working before. 
This was the simplest solution I could think of, where you first create and return the object first, then use the object to start the main loop.

MAVProxy/modules/mavproxy_chat/__init__.py
SriramS-77: I have only tested on Windows so far. I will look into it.

SriramS-77: I looked into why it was not working in Linux (I tried in Ubuntu to recreate the error). It seems wxPython is built on top of GTK in Ubuntu, which is much stricter than the Windows version. It is not allowing reinitializing app or rerunning mainLoop in the same process. I tried different approaches to getting it working in the same process, but keep getting seg fault.
I think it would be better to launch a different process for the app using multiprocessing. What are your thoughts?

SriramS-77: Another approach would be to:
- Hide the chat window and clear the data (unloading).
- Show the window along with new assistant thread (reloading).

I think this might be better, since there is no need to rework the implementation for handling mavlink acknowledgement.

rmackay9: Hi @SriramS-77,

Either way sounds good to me.  By the way, do you have an OpenAPI key?  I just checked the list of ArduPilot sponsored API keys and I don't see you there.  If you'd like one, please just PM me on ArduPilot's discord server, I'm "rmackay9" over there too.

PR #1566, opened 2025-05-09
silvus: fixes for streamcaster v5

tridge:
node ID handling has changed

### DIFF
--- a/MAVProxy/modules/mavproxy_silvus.py
+++ b/MAVProxy/modules/mavproxy_silvus.py
@@ -27,6 +27,8 @@ class SilvusModule(mp_module.MPModule):
         # filter_dist is distance in metres
         self.silvus_settings = mp_settings.MPSettings([("gnd_ip", str, ""),
                                                        ("air_ip", str, ""),
+                                                       ("air_node", int, 0),
+                                                       ("gnd_node", int, 0),
                                                        ("nmea_ip", str, ""),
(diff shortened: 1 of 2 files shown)


MAVProxy/modules/mavproxy_silvus.py
```diff
@@ -201,21 +198,23 @@ def get_radio_data(self):
         if len(remoteip.split('.')) != 4:
             return
 
-        localnode = str(self.ip_to_id(localip))
-        remotenode = str(self.ip_to_id(remoteip))
+        localnode = str(self.silvus_settings.gnd_node)
+        remotenode = str(self.silvus_settings.air_node)
```
peterbarker, line 202: So we stop deriving one from the other and get the user to set the node ID manually.

I guess this allows for compatability with both versions of the streamcaster firmware, but might have been nice to have the newer octets-to-nodeid conversion for a node if the node ID was zero in the settings.

peterbarker (approved):

MAVProxy/modules/mavproxy_silvus.py
tridge: the reason for the change is each radio has a 172.x.x.x address and a 192.y.y.y address. The node ID can only be derived from the 172 (native) address, but we commonly only have a route to the 192 ("private") address

PR #1569, opened 2025-05-21
Treewide: Remove python2 conditional logic/imports

Ryanf55:
Remove conditional imports and conditional logic searching for `sys.version_info` that only apply to python2.

Used ChatGTP with a grepped import list to find legacy python2 module logic too.


peterbarker (changes requested):
Has to pass CI.

How was this tested?

MAVProxy/modules/lib/rline.py
```diff
@@ -58,13 +58,13 @@ def __init__(self, prompt, mpstate):
     def set_prompt(self, prompt):
         if prompt != self.prompt:
             self.prompt = prompt
-            if platform.system() != 'Windows' or sys.version_info < (3, 0):
+            if platform.system() != 'Windows' or sys.version_info < (3, 0): #??
```
Ryanf55, line 61: Finish before merge

PR #1575, opened 2025-05-31
tracker: release manual control when setting mode

robertlong13:
When I was testing https://github.com/ArduPilot/ardupilot/pull/30202, it took me a while to figure out how to release manual control. Once you do `tracker position`, you cannot get the tracker to do anything else until you call `tracker position 32767 32767`.

I think it's probably appropriate to release manual control any time the user changes mode, so that's what I've done here.

### DIFF
--- a/MAVProxy/modules/mavproxy_tracker.py
+++ b/MAVProxy/modules/mavproxy_tracker.py
@@ -37,6 +37,7 @@ class TrackerModule(mp_module.MPModule):
               ('debug', int, 0)
               ]
             )
+        self._manual_control = False
         self.add_command('tracker', self.cmd_tracker,
                          "antenna tracker control module",
                          ['<start|arm|disarm|level|mode|position|calpress|mode>',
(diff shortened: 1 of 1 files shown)


tridge:
Full report: https://uav.tridgell.net/DevCallReviews/2026_09_01_AIReview/devcall_pr_reviews.html#prMAVProxy-1575

Reviewed at head `b8dd2d249c`. **COMMENT** — the premise is right and it still applies cleanly to master. Nothing in AntennaTracker clears `manual_control_yaw`/`_pitch` on a mode change, 32767 is exactly the firmware's `0x7FFF` sentinel, and `positions[0]`→yaw / `positions[1]`→pitch matches the firmware's field use.

**One ordering issue.** `cmd_tracker_position` assigns `self._manual_control` at line 103 and only *then* calls `manual_control_send()` at 104. If the send raises on the release path, the flag has already been cleared, the exception propagates out of the nested call so `connection.set_mode()` on the next line never runs, and a retry won't re-attempt the release because the flag now says there's nothing to release — the mode change is lost and the state is wrong. The take path has the mirror problem but is harmless (a spurious release is a no-op). Assigning the flag after a successful send, or a `try`/`finally`, closes both.

For transparency: my two review passes disagreed on this one — one called it a blocking bug, the other a minor note. Both described the same mechanics, so I've landed between them: a real ordering defect with a trivial fix, but it needs the send to actually raise, and this is an antenna tracker rather than a vehicle.

Two smaller notes. The generic `mode` command bypasses the release entirely — with the tracker selected as the current link, `mode AUTO` goes through `mavproxy_mode.py`'s `cmd_mode` (`:41`, `:74`) straight to `master.set_mode()`, leaving the axes latched. Out of scope for a 6-line fix, but worth a follow-up. And `find_connection()` is re-resolved inside the nested call and prefers `self.connection`, so the release can go down a different link than the one that took manual control.

Issue #1582, opened 2025-06-13
Map can't be loaded and then it can't be viewed in the logs


NemanjaDragisic:
There is a logs that appears when I run the simulation and want to have map with me.

/venv-ardupilot/lib/python3.12/site-packages/MAVProxy/modules/mavproxy_map/mp_tile.py:601: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
import pkg_resources
Process Process-3:
Traceback (most recent call last):
File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
self.run()
File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run
self._target(*self._args, **self._kwargs)
File "/venv-ardupilot/lib/python3.12/site-packages/MAVProxy/modules/mavproxy_map/mp_slipmap.py", line 87, in child_task
self.mt = mp_tile.MPTile(download=self.download,
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/venv-ardupilot/lib/python3.12/site-packages/MAVProxy/modules/mavproxy_map/mp_tile.py", line 210, in init
self._loading = mp_icon('loading.jpg')
^^^^^^^^^^^^^^^^^^^^^^
File "/venv-ardupilot/lib/python3.12/site-packages/MAVProxy/modules/mavproxy_map/mp_tile.py", line 615, in mp_icon
img = cv2.imdecode(raw, cv2.IMREAD_COLOR)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cv2.error: OpenCV(4.11.0) 👎 error: (-5:Bad argument) in function 'imdecode'

NemanjaDragisic:
I have just pulled the latest version of github source but again the problem is 

Traceback (most recent call last):
  File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "~/venv-ardupilot/lib/python3.12/site-packages/MAVProxy/modules/mavproxy_map/mp_slipmap.py", line 87, in child_task
[...]

PR #1584, opened 2025-06-17
Map: Use correct file path for icons

stephendade:
For for #1582. Turns out the method of getting the map icon folder was incorrect.


NemanjaDragisic:
The change that I did was done after the one you did. Of course it is improvement, but I added the transformation of raw to numpy array of uint8 after line 634. Otherwise, the raw stays as a stream that can't be displayed, and error happens.

PR #1588, opened 2025-06-18
`wave` is a Python Standard Library module

cclauss:
Installing the third-party ___Whole Architecture Verification___ module https://pypi.org/project/Wave is [breaking our builds](https://github.com/ArduPilot/MAVProxy/actions/workflows/windows_build.yml).

The Python Standard Library module  [`wave`](https://docs.python.org/3/library/wave.html) has an API that corresponds to:
https://github.com/ArduPilot/MAVProxy/blob/a45549f9bcb09603be1507c260de2bafe516212a/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py#L72-L78

Therefore, we should stop doing `pip install wave` and instead use the Standard Library.

The docs at https://ardupilot.org/mavproxy/docs/modules/chat.html#prerequisites must also be modified.
* ArduPilot/ardupilot_wiki#6947

@rmackay9 Your review, please.

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           python3 -m pip install -U wheel setuptools pip
           python3 -m pip install -U pywin32 lxml pymavlink numpy matplotlib pyserial opencv-python PyYAML Pygame Pillow wxpython prompt-toolkit scipy
-          python3 -m pip install -U openai wave pyaudio
+          python3 -m pip install -U openai pyaudio
           python3 -m pip install -U pyinstaller==6.7.0 packaging 
       - name: Download Inno Setup installer
         run: curl -L -o installer.exe http://files.jrsoftware.org/is/6/innosetup-6.3.1.exe
--- a/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
+++ b/MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py
@@ -5,13 +5,12 @@ Randy Mackay, December 2023
 AP_FLAKE8_CLEAN
 '''
 
-
+import wave
 try:
     import pyaudio  # install using, "sudo apt-get install python3-pyaudio"
-    import wave     # install with "pip3 install wave"
     from openai import OpenAI
 except Exception:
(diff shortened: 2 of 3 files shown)


cclauss:
> was it tested as working with the standard module?

Yes, I ran the full test suite that was included in:
* #1278

(Sorry, only joking...)

---

This is a Windows app, and I have no access to Windows, but I did...

On macOS, I created a new `test_record_audio.py` file.
```python
#!/usr/bin/env -S uv run --script

# /// script
# requires-python = ">=3.9"
# dependencies = [
#     "openai",
#     # "playsound",
#     "pyaudio",
# ]
# ///

# Paste `MAVProxy/modules/mavproxy_chat/chat_voice_to_text.py` here.

if __name__ == "__main__":
    from pathlib import Path

    filename = chat_voice_to_text().record_audio()
    assert filename == "recording.wav"
    filepath = Path(__file__).parent / filename
    assert filepath.exists()
    assert filepath.is_file()
    print(f"{filepath.stat() = }")
    print(f"Size of {filepath} is {filepath.stat().st_size:,} bytes.")
```

Make a change to the code:
```diff
-        while not stop_recording[0]:
+        # while not stop_recording[0]:
+        print("chat: recording audio for 5 seconds...")
+        for _ in range(0, int(44100 / 1024 * 5)):  # record for 5 seconds
```
% `uv run test_record_audio.py ` # Say something when the recording starts!
```
chat: recording audio for 5 seconds...
filepath.stat() = os.stat_result(st_mode=33188, st_ino=291196442, st_dev=16777231, st_nlink=1, st_uid=501, st_gid=20, st_size=440364, st_atime=1750304146, st_mtime=1750304146, st_ctime=1750304146)
Size of recording.wav is 440,364 bytes.
```
% `open recording.wav -a GarageBand`
Hit the ___play___ button...

rmackay9:
The author ping'd me about this so I've added it to the EUDevCall.  I'm not really qualified to review the PR

peterbarker:
> @peterbarker Time ran out on today's EUDevCall which only focused on the main repo.

Ah, sorry.  We usually only do focus on the main repo, it's actually very rare that we remember to go and check other repos for DevCallEU!
 
> Are you satisfied with the testing above? What are the todos here? This does fix [broken builds](https://github.com/ArduPilot/MAVProxy/actions/workflows/windows_build.yml).

Our Windows maintainer tested it as good - so I've merged it, thanks!

PR #1589, opened 2025-06-22
sign forwarded messages

tridge:
when forwarding an unsigned message from a downstream link to an upstream link that has signing enabled then re-pack the message to sign it.

This allows for tools like DroneCAN GUI tool to be used more easily


cclauss:
Rebasing should fix the failing Windows build.

tridge:
added mavfwd_signing as an option, default true

PR #1605, opened 2025-07-14
camera_projection.py: Fix undefined names 'json' and 'self'

cclauss:
flake8 rule F821 -- Undefined names have the potential to raise NameError at runtime.

`import json` needed for lines 64 and 106.
~`self` needed for line 87.~
Line 87 mistakenly used `self`.


peterbarker (approved):
I might have moved to using f-string and a `ValueError`, but this is definitely better!

peterbarker:
Merged, thanks!

PR #1609, opened 2025-07-14
rline.py: Fix undefined name `redisplay`

cclauss:
flake8 rule F821 -- Undefined names will be placated by a targeted NameError but not by a too broad Exception.


peterbarker (changes requested):
zx

peterbarker (approved):

PR #1617, opened 2025-08-04
MAVExplorer: allow dump to work with instance ids, also update dump's usage text

MichelleRos:
Allows dump to work with instance ids. 

Not sure if we want tab complete to also show the instance ids?


peterbarker (approved):
LGTM

peterbarker:
Tested, works nicely!

PR #1625, opened 2025-09-09
Add support for home-centered inclusion circle fence

Ryanf55:
Implementation of https://github.com/mavlink/mavlink/pull/2346

Paired with https://github.com/ArduPilot/ardupilot/pull/31064

image

### DIFF
--- a/MAVProxy/modules/mavproxy_fence.py
+++ b/MAVProxy/modules/mavproxy_fence.py
@@ -17,6 +17,8 @@ if mp_util.has_wxpython:
     from MAVProxy.modules.lib.mp_menu import MPMenuCallTextDialog
     from MAVProxy.modules.lib.mp_menu import MPMenuItem
 
+DEFAULT_CIRCLE_SIZE = 500
+
 
 class FenceModule(mission_item_protocol.MissionItemProtocolModule):
     '''uses common MISSION_ITEM protocol base class to provide fence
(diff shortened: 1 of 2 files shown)


srmainwaring:
Would be good if this could be sync'd with the params from the simple cylindrical fence in ArduPilot.

Ryanf55:
Pulled in Peter's move-home support, cleaned commits. Ready for merge assuming the linters pass.

peterbarker (commented):
CI must pass

peterbarker (changes requested):
CI must pass

peterbarker:
@Ryanf55 did you re-test this?

peterbarker:
Throws an error about uploading floating point values when actually used

Ryanf55:
> Throws an error about uploading floating point values when actually used

I fixed in 0e82c6511ca76f2db41c72b4aa7fb375b28d8f71. I'll squash if you are ready for that.
Depending on the order you sent the fence vs list it, you can avoid the bug.
Yes, it seems odd to populate lat/lon there, but you need to do it because `display_polyfences_inclusion_circles` calls  `circles_of_type`. If you don't populate p.x and p.y for the scaled home value, it shows up at 0,0 global coords.

peterbarker:
I've adjusted this branch to handle home change slightly differently so we don't need to munge before the map library sees things.  Also means we can render the circle in a slightly different colour.

I'm going to merge this if you don't have objections, @Ryanf55 :-)

peterbarker (approved):

PR #1626, opened 2025-09-16
misc: add toggleable repeat commands

joshanne:
Allows a repeat command to be toggled rather than remove-then-add.
Updates the list to print the current state of each of the RepeatCommands.

### DIFF
--- a/MAVProxy/modules/mavproxy_misc.py
+++ b/MAVProxy/modules/mavproxy_misc.py
@@ -25,9 +25,10 @@ class RepeatCommand(object):
         self.period = period
         self.cmd = cmd
         self.event = mavutil.periodic_event(1.0/period)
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_misc.py
```diff
@@ -523,6 +524,23 @@ def cmd_repeat(self, args):
                 return
             self.repeats.pop(i)
             return
+        elif args[0] == 'toggle':
+            if len(args) < 2:
+                print("Usage: repeat toggle INDEX..")
+                return
+
+            for i in range(1, len(args)):
```
peterbarker, line 532: Could use `enumerate` here

peterbarker (approved):
LGTM

PR #1628, opened 2025-09-17
MAVExplorer: plot predefined graphs from command line

be-ska:
I find very useful the ability to plot predefined expressions from `mavgraphs.xml`, but it seems there is no convenient way to do that from terminal, without using GUI. 
This PR add the `graphs` command, that's very similar to the `graph` one, with autocompletion support, but it plots predefined graphs.


peterbarker:
Merged, thanks!

PR #1635, opened 2025-09-30
mavproxy_console: add GNSS heading status

mirkix:
This PR add information to the mavproxy_console that show if a GPS / GNSS is providing GNSS-Heading. The information is located after the sat count.

- It will show a 'h' if the GNSS is configured to provide GNSS-Heading but the values are not valid
- It will show a 'H' if the GNSS is configured to provide GNSS-Heading and the data is valid

gnssheading

This show a drone with two GNSS with heading (four antennas in total), booth GNSS are configured and providing valid heading data at the moment.


peterbarker (approved):

mirkix:
@peterbarker will this be merged soon?

PR #1642, opened 2025-11-21
mavproxy_rc: support rc override in SITL for multiple vehicles

srmainwaring:
Ensure `rc` overrides are applied to the vehicle on the current link, or all vehicles if `alllinks` is used.

Current behaviour is that the `rc` override is always applied to the vehicle with the highest link number, regardless of which `sysid` is active.

Single vehicle behaviour is unchanged.

### Example

Start a multi-vehicle session:

```bash
sim_vehicle.py --debug -v Rover -f rover --count 3 --auto-sysid --location CMAC --auto-offset-line 90,5 --console --map              
```

Arm all the vehicles

```bash
allinks arm throttle
```

Ensure vehicle 1 (link 0) is active

```bash
vehicle 1
```

Set the throttle

```bash
rc 3 1600
```

Before this change vehicle 3 will move forward, after this change vehicle 1 will move forward.

Also tested in Gazebo flying two Zephyr's.

### DIFF
--- a/MAVProxy/modules/mavproxy_rc.py
+++ b/MAVProxy/modules/mavproxy_rc.py
@@ -30,7 +30,7 @@ class RCModule(mp_module.MPModule):
         self.add_command('switch', self.cmd_switch, "flight mode switch control", ['<0|1|2|3|4|5|6>'])
         self.rc_settings = mp_settings.MPSettings(
             [('override_hz', float, 10.0)])
-        if self.sitl_output:
+        if self.sitl_outputs:
             self.rc_settings.override_hz = 20.0
         self.add_completion_function('(RCSETTING)',
                                      self.rc_settings.completion)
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_rc.py
```diff
@@ -112,10 +112,10 @@ def mavlink_packet(self, m):
 
     def send_rc(self):
         '''send RC override packet'''
-        if self.sitl_output:
+        if self.sitl_outputs:
             chan16 = self.override[:16]
             buf = struct.pack('<HHHHHHHHHHHHHHHH', *chan16)
-            self.sitl_output.write(buf)
+            self.sitl_outputs[self.master.linknum].write(buf)
```
peterbarker: This looks like it is conflating link numbers and sitl-output-port numbers.

That probably means you could elicit a runtime exception with `link add` - or simply starting with a different number of links to sitl-rc-outputs.

There isn't a 1:1 correspondence between link numbers and SITL outputs - you could have a connection to the same simulation process on 2 of its mavliunk ports, for example.

peterbarker (changes requested):

PR #1645, opened 2025-12-02
Add configurable dgps port

Ryanf55:
# Purpose

Allow changing which port we listen to for DGPS (and allow injecting broadcast RTCM).

Also fix `DGPS: GPS Inject Failed: 'NoneType' object has no attribute 'mav'` when you are receiving RTCM but have not yet connected to the master.

# Test Performed

Using the CLI, we can see the socket is now open on a different port.

```
ryan@friedmanr-ubuntu:~/Dev/MAVProxy$ mavproxy.py --console
/usr/lib/python3/dist-packages/google/protobuf/__init__.py:37: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  __import__('pkg_resources').declare_namespace(__name__)
Auto-detected serial ports are:
/dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
Connecting to /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
Connect /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0 source_system=255
Loaded module console
Log Directory: 
Telemetry log: mav.tlog
Waiting for heartbeat from /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
MAV> set moddebug 3
MAV> module load DGPS
MAV> DGPS: Listening for RTCM packets on UDP://127.0.0.1:13320
Loaded module DGPS

MAV> dgps set portnum 13340
MAV> Settings changed: portnum
DGPS: Listening for RTCM packets on UDP://127.0.0.1:13340
```

```
$ ss -unl4p | grep mavproxy
UNCONN     0           0                    127.0.0.1:13340                0.0.0.0:*         users:(("mavproxy.py",pid=146169,fd=20))
```

I verified this with a vehicle that has a  Ublox F9P and a Sparkfun RTK Base Station.

Note- instead of binding localhost, I now bind 0.0.0.0 which allows receiving packets destined for 192.168.1.255, which is the default behavior for some RTCM corrections such as from the Sparkfun RTK EVK.

### DIFF
--- a/MAVProxy/modules/mavproxy_DGPS.py
+++ b/MAVProxy/modules/mavproxy_DGPS.py
@@ -6,21 +6,70 @@ support for a GCS attached DGPS system
 import socket, errno
 from pymavlink import mavutil
 from MAVProxy.modules.lib import mp_module
+from MAVProxy.modules.lib import mp_settings
+from MAVProxy.modules.lib.mp_settings import MPSetting
 
 class DGPSModule(mp_module.MPModule):
+
+    @staticmethod
+    def default_settings():
+        return mp_settings.MPSettings([
+                MPSetting("portnum", int, 13320),
+                MPSetting("ip", str, "127.0.0.1"),
+            ])
+
     def __init__(self, mpstate):
         super(DGPSModule, self).__init__(mpstate, "DGPS", "DGPS injection support for SBP/RTCP/UBC")
-        self.portnum = 13320
+        self.dgps_settings = DGPSModule.default_settings()
+        self.inject_seq_nr = 0
+        self.cmdname = "dgps"
+        
+        self.create_port()
+        self.add_command(
+            self.cmdname,
+            self.cmd_dgps,
+            f"{self.cmdname} control",
+            [
+                "set (DGPS_SETTING)",
+            ],
+        )
+        self.add_completion_function(
+            "(DGPS_SETTING)", self.dgps_settings.completion
+        )
+        self.dgps_settings.set_callback(self.on_setting_set)
+
+    def create_port(self):
+        print(f"DGPS: Listening for RTCM packets on UDP://{self.dgps_settings.get('ip')}:{self.dgps_settings.get('portnum')}")
         self.port = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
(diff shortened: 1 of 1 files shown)


MAVProxy/modules/mavproxy_DGPS.py
```diff
@@ -6,18 +6,65 @@
...
+            "(DGPS_SETTING)", self.dgps_settings.completion
+        )
+        self.dgps_settings.set_callback(self.on_setting_set)
+
+    def create_port(self):
+        addr = "0.0.0.0"
+        print(f"DGPS: Listening for RTCM packets on UDP://{addr}:{self.dgps_settings.get('portnum')}")
         self.port = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
         self.port.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        self.port.bind(("127.0.0.1", self.portnum))
```
peterbarker: Perhaps make this a parameter?

Listening on every address has security implications which you can easily side-step (AKA "punt to the user") this way

peterbarker (changes requested):

MAVProxy/modules/mavproxy_DGPS.py
Ryanf55: Done, now it's parameterized. Default is 0.0.0.0 if that's ok.

peterbarker: > Done, now it's parameterized. Default is 0.0.0.0 if that's ok.

Well - have to see what tridge says on that.

As I say, listening on that address means having to think about what might connect there and do Bad Things.  Which is why I'd suggest sticking with 127.0.0.1.  Maybe do that so we can get this in and use a second PR to change it to 0.0.0.0 by default.

Ryanf55: Makes sense. Done!

peterbarker (approved):
LGTM

PR #1647, opened 2025-12-06
horizon: fix oversized text on HiDPI displays

rbeall:
**PR Description:**
Recent matplotlib versions (3.5+) report the actual display DPI on HiDPI/Retina screens. The font scaling in calcFontScaling() multiplies by figure.dpi, which now returns ~200 on Retina displays instead of the expected 100, causing all text to render at 2x size.

Fix by using a fixed reference DPI of 100 for the font calculations.

**How Tested:**
- macOS 14.x with Retina display
- matplotlib 3.10.7
- Verified text renders at expected size before/after fix
- Confirmed window resizing still works correctly

### DIFF
--- a/MAVProxy/modules/lib/wxhorizon_ui.py
+++ b/MAVProxy/modules/lib/wxhorizon_ui.py
@@ -166,8 +166,10 @@ class HorizonFrame(wx.Frame):
         if not self.resized:
             oldypx = self.ypx
             oldxpx = self.xpx
-            self.ypx = self.figure.get_size_inches()[1]*self.figure.dpi
-            self.xpx = self.figure.get_size_inches()[0]*self.figure.dpi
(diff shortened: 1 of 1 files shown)


rbeall:
I don't have a non Mac/Retina setup to test this on at the moment to make sure that it doesn't have detrimental affects on other machines, but it does fix it on Mac.

peterbarker:
This causes the display to utterly freeze on Ubuntu when I resize it.

It does do nice things to the font scaling, 'though :-)

peterbarker (approved):
So I don't really like this.  Pretty sure there's got to be a better way to do things than use a fixed value - something else has to be going wrong here somewhere.

But it does fix the display for me, and it's a problem for @rbeall , so it'll do for now.

Merging, thanks!

PR #1668, opened 2026-03-11
Joystick: adding FlySky SM001.

AndKe:
Those devices are available from aliexpress/banggood, and other.

### DIFF
--- a/MAVProxy/modules/mavproxy_joystick/joysticks/SM001.yml
+++ b/MAVProxy/modules/mavproxy_joystick/joysticks/SM001.yml
@@ -1,0 +1,27 @@
+description: >
+  FlySky SM001 - USB "8ch Remote Controller"
+match:
+  - '*XTR+G2+FMS Controller*'
+controls:
+  - channel: 1
+    type: axis
+    id: 3
+  - channel: 2
+    type: axis
+    id: 0
+    invert: true
+  - channel: 3
+    type: axis
+    id: 1
+  - channel: 4
+    type: axis
+    id: 4
+  - channel: 5
+    type: button
+    id: 0
+  - channel: 6
+    type: button
+    id: 2
+  - channel: 7
+    type: button
+    id: 3


tridge:
Full report: https://uav.tridgell.net/DevCallReviews/2026_09_01_AIReview/devcall_pr_reviews.html#prMAVProxy-1668

Reviewed at head `8f11fa6912`. **COMMENT** — no blockers. This is a self-contained data file that can't affect anyone whose joystick doesn't report the matching name.

**What was verified:** the file parses with `yaml.safe_load` and constructs a real `controls.Joystick()` at this head — `chan_max=7`, `read()` returns `[1500, 1500, 1500, 1500, 1000, 1000, 1000]` at rest, no exceptions. Keys match what the module requires (`description` is mandatory), channel numbers are 1-based and axis/button ids 0-based like every sibling file, and the `match` pattern collides with no shipped definition (I ran every pattern against the string). `+` isn't an fnmatch metacharacter, so the pattern is literal and correct. The axis mapping is also identical to the existing `carolbox-usb.yml` for the same XTR/AeroFly/FMS simulator-cable family, which is good corroboration.

**What I can't check, and it's the part that matters:** two things differ from carolbox — `invert: true` on channel 2, and ch5/6/7 as buttons rather than axes. Neither is verifiable from the file, and if either is wrong the user silently gets the wrong control rather than an error. Could you confirm them on the actual device with `python -m MAVProxy.modules.mavproxy_joystick.findjoy`?

Related: the `match` string looks like the USB *simulator cable*'s name rather than the SM001's own, so a different transmitter on the same cable would inherit this mapping. Worth saying whether that name is reported by the SM001 itself.

Minor: the description says "8ch" but only 7 channels are defined, and axis 2 and button 1 are unused (carolbox uses axis 2 for ch7). Channel 8 is just left untouched, so it's harmless — but the gaps are worth a word in the description.

AndKe:
Regarding "Channel 8 is just left untouched, so it's harmless — but the gaps are worth a word in the description."
I have two such  USB devices, and it may be a manufacturing issue or just marketing, but when looking at the joystick data, one control is simply dead/not in use.

PR #1671, opened 2026-04-02
mavproxy_cmdlong: flake8-correctness

peterbarker:


peterbarker:
@stephendade any idea what's up with the windows installer here?  Has github changed their setup somehow?

PR #1675, opened 2026-04-07
Ntrip: Fix SSL connection exception

stephendade:
The NTRIP module currently fails for SSL connections. Fixed by added a ``ssl.create_default_context()``

Tested with ``ntrip.data.gnss.ga.gov.au:443``


peterbarker (approved):
LGTM

PR #1677, opened 2026-04-26
grapher: generalise multi-axis support with auto/dual/multi modes

tridge:
Field suffix :N now selects y-axis N (2-9) instead of just :2.

Adds an axis_mode setting (auto/dual/multi, default auto) wired through the MAVExplorer set command and the grapher.py --axis-mode CLI flag:
- dual preserves the legacy 2-axis layout (legend per axis, no vertical y-labels), capping any :N>=2 to the right axis
- multi stacks each extra axis on the left with a vertical label coloured to the line; when fields share an axis the label parts are coloured individually via an AnchoredOffsetbox/VPacker composite
- auto picks dual for 1-2 axes, multi for 3+

Hover readout (format_coord) generalised to report y for every axis. Double-click rescale now iterates all axes.

examples:
axis-mode dual (or auto):
image
same graph with multi mode:
image
and with 3 axes:
image


peterbarker (approved):
LGTM

Lightly tested locally on a tlog and dataflash log

PR #1687, opened 2026-05-18
mavproxy_signing: allow signing keys with no protocol prefix

tridge:
An entry of the form "host:port" in ~/.mavproxy/signing.keys now matches a device of the form "proto:host:port" for any protocol (udp, tcp, etc).


peterbarker (approved):
Alternatively you could allow globbing instead - less magic.

i.e. `*:192.168.1.56:5760`

tridge:
> Alternatively you could allow globbing instead - less magic.

I considered that, but I think this is simpler from the users point of view

PR #1688, opened 2026-05-19
mp_tile: add China-accessible tile services

hxf0223:
Add GoogleChinaSat (google.cn satellite) and three Tianditu services (TiandituSat/Map/Ter) for faster map loading in China mainland.

Tianditu uses WGS-84 (no coordinate offset), requires a free API key set via TIANDITU_KEY environment variable.
URLs follow the official WMTS standard format from: http://lbs.tianditu.gov.cn/server/MapService.html

Use MAP_SERVICE env variable to select service, e.g.:
  export MAP_SERVICE=TiandituSat
  export TIANDITU_KEY=your_key


peterbarker (approved):
LGTM

PR #1691, opened 2026-05-25
mavproxy_cmdlong: Add componentID to messages

stephendade:
The ``cmdlong`` module has a bunch of commands which are sent to ``target_component=0``, rather than the ``self.settings.target_component``.

I've updated them all to use ``self.settings.target_component``, but I'm conscious that I might be breaking some default behaviors?

This caught me out trying to send a ``cammsg`` to a MAVLink camera, but sine MAVProxy sent it out with a ``target_component=0``, instead of the actual camera's compid and subsequently got filtered out by the camera.


peterbarker (approved):
LGTM

PR #1700, opened 2026-07-03
joystick: match Dual Action names reported by SDL >= 2.0.6

hfujikawa77:
SDL 2.0.6 and later strip the duplicated vendor prefix from the device name, so the Logitech/Logicool F310 in DirectInput mode is reported as "Logicool Dual Action" / "Logitech Dual Action" and no longer matches the shipped definitions.  Add the new names alongside the old ones.


peterbarker (approved):
Can't help but think that first match shouldn't have worked.

PR #1702, opened 2026-07-11
Add support for MAV_CMD_DO_ORBIT

peterbarker:
Companion to https://github.com/ArduPilot/ardupilot/pull/32434

### DIFF
--- a/MAVProxy/modules/lib/mp_menu.py
+++ b/MAVProxy/modules/lib/mp_menu.py
@@ -471,6 +471,124 @@ class MPMenuCallTextDropdownDialog(object):
         # Return tuple with text value and selected dropdown value
         return text_value + " " + dropdown_value
 
+class MPMenuCallMultiTextDropdownDialog(object):
+    '''used to create a dialog callback with multiple values and optional dropdowns.
+
+    Each field is a (label, default) tuple, and each dropdown a
+    (label, options, default) tuple.  An extra trailing element - a key -
+    may be added to either (i.e. (label, default, key) or
+    (label, options, default, key)); when present the corresponding value
+    is emitted as "key=value", otherwise the bare value is emitted.
+    '''
+    def __init__(self, title='Enter Values', fields=None, dropdowns=None):
+        self.title = title
+        self.fields = fields or []
+        self.dropdowns = dropdowns or []
+
+    @classmethod
+    def from_arg_spec(cls, title, spec, ctx=None):
+        '''build a dialog from a command argument spec.
+
+        spec is an ordered dict of key->facets (see
+        MPModule.parse_key_value_spec).  Each entry that carries a 'label'
+        becomes a widget emitting "key=value": a dropdown when it has an
+        'options' list, otherwise a text field.  Entries without a 'label'
+        (e.g. keys entered another way) are skipped.  A 'default' facet may
+        be a plain value or a callable, which is passed ctx and evaluated
+        now (when the dialog is built) to produce the default.
+        '''
+        fields = []
+        dropdowns = []
+        for (key, facets) in spec.items():
+            if 'label' not in facets:
+                continue
+            default = facets.get('default', '')
+            if callable(default):
+                default = default(ctx)
+            if 'options' in facets:
+                dropdowns.append((facets['label'], facets['options'], default, key))
+            else:
+                fields.append((facets['label'], default, key))
+        return cls(title=title, fields=fields, dropdowns=dropdowns)
+
+    def call(self):
+        '''show a dialog with multiple value entries and optional dropdowns'''
+        from MAVProxy.modules.lib.wx_loader import wx
+
+        # Create a custom dialog
+        dlg = wx.Dialog(None, title=self.title, size=(400, 150))
+
+        # Create a vertical box sizer for the dialog
+        main_sizer = wx.BoxSizer(wx.VERTICAL)
+
+        # One row of label plus text entry per field
+        text_ctrls = []
+        for field in self.fields:
+            (label, default) = field[0], field[1]
+            input_sizer = wx.BoxSizer(wx.HORIZONTAL)
+            text_label = wx.StaticText(dlg, label=label + ":")
+            memory_key = self.title + ":" + label
+            default = last_value_selection.get(memory_key, default)
+            text_ctrl = wx.TextCtrl(dlg, value=str(default), size=(200, -1))
+            text_ctrls.append(text_ctrl)
+            input_sizer.Add(text_label, 1, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
+            input_sizer.Add(text_ctrl, 0, wx.ALL | wx.EXPAND, 5)
+            main_sizer.Add(input_sizer, 0, wx.ALL | wx.EXPAND, 5)
+
+        # One row of label plus choice control per dropdown
+        dropdown_ctrls = []
+        for dropdown in self.dropdowns:
+            (label, options, default) = dropdown[0], dropdown[1], dropdown[2]
+            input_sizer = wx.BoxSizer(wx.HORIZONTAL)
+            dropdown_label = wx.StaticText(dlg, label=label)
+            dropdown_ctrl = wx.Choice(dlg, choices=options)
+            dropdown_ctrls.append(dropdown_ctrl)
+
+            default_idx = 0
+            for i in range(len(options)):
+                if options[i] == default:
+                    default_idx = i
+            memory_key = self.title + ":" + label
+            dropdown_ctrl.SetSelection(last_dropdown_selection.get(memory_key, default_idx))
+
+            input_sizer.Add(dropdown_label, 1, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
+            input_sizer.Add(dropdown_ctrl, 0, wx.ALL | wx.EXPAND, 5)
+            main_sizer.Add(input_sizer, 0, wx.ALL | wx.EXPAND, 5)
+
+        # Create button sizer with OK and Cancel buttons
+        button_sizer = dlg.CreateButtonSizer(wx.OK | wx.CANCEL)
+        main_sizer.Add(button_sizer, 0, wx.ALL | wx.ALIGN_CENTER, 10)
+
+        # Set the sizer for the dialog
+        dlg.SetSizer(main_sizer)
+
+        # Fit the dialog to its contents
+        dlg.Fit()
+
+        # Show the dialog and get the result
+        if dlg.ShowModal() != wx.ID_OK:
+            return None
+
+        values = []
+        for (field, text_ctrl) in zip(self.fields, text_ctrls):
+            label = field[0]
+            key = field[2] if len(field) > 2 else None
+            text_value = text_ctrl.GetValue()
+            last_value_selection[self.title + ":" + label] = text_value
+            values.append("%s=%s" % (key, text_value) if key else text_value)
+
+        for (dropdown, dropdown_ctrl) in zip(self.dropdowns, dropdown_ctrls):
+            (label, options) = dropdown[0], dropdown[1]
+            key = dropdown[3] if len(dropdown) > 3 else None
+            dropdown_index = dropdown_ctrl.GetSelection()
+            if dropdown_index != -1:
+                dropdown_value = options[dropdown_index]
+                last_dropdown_selection[self.title + ":" + label] = dropdown_index
+                values.append("%s=%s" % (key, dropdown_value) if key else dropdown_value)
+
+        # Return the entered values, space-separated, dropdown values last
+        return " ".join(values)
+
 class MPMenuConfirmDialog(object):
     '''used to create a confirmation dialog'''
     def __init__(self, title='Confirmation', message='', callback=None, args=None):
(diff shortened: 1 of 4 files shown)


tridge:

<details><summary>Previous review (2026-09-01)</summary>

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_01_AIReview/devcall_pr_reviews.html#prMAVProxy-1702

Reviewed at head `145e62aa3e`. **COMMENT** — no blockers.

**The parameter mapping is correct**, which is the thing most worth getting right here and the easiest to get wrong. Checked field by field against the fetched `common.xml`: p1 radius in metres with *"Positive: orbit clockwise. Negative: orbit counter-clockwise."* passed through signed and documented that way in your help text; p2 velocity m/s with NaN meaning vehicle default; p3 mapped to the `ORBIT_YAW_BEHAVIOUR` enum values; **p4 declared `units="rad"`** and correctly converted from circuits as `orbits * 2 * pi` (verified: 3 circuits → 18.8495559 = 6π); p5/p6 scaled to degE7 in a COMMAND_INT. COMMAND_INT is right given the lat/lon, and the non-`_INT` frame constants are correct — the spec marks the `_INT` variants superseded since 2024-03.

**Two things worth fixing:**

1. `velocity=` and `yaw=` raise an uncaught `ValueError` from a GUI-reachable path. Both are declared `type=str` in `orbit_arg_spec`, so the arg parser's `try/except ValueError` never sees them and the later `float(...)` raises. Reachable by simply clearing the free-text "Velocity (m/s)" box in the Orbit Here dialog; the user gets `ERROR in command [...] : could not convert string to float: ''` with no usage. (`mavproxy_mode.py:248`, `:258`)
2. No mode change is requested and the NACK isn't surfaced. `cmd_guided` sends DO_REPOSITION with `MAV_DO_REPOSITION_FLAGS_CHANGE_MODE`; `cmd_orbit` has no equivalent and nothing reads COMMAND_ACK, so from Loiter it just fails with a generic console line a map-menu user is unlikely to be watching.

That second point matters more than usual because **no released ArduPilot firmware implements this command yet** — a search of a full checkout finds `DO_ORBIT` only in the bundled mavlink XML, zero source files. The only implementation is the still-open ArduPilot/ardupilot#32434 (Copter-only, Guided-only, size-gated), so today every vehicle NACKs this. Worth cross-linking the two PRs in the description.

Smaller: negative `orbits` emits a negative p4, below the spec's `minValue="0"` (ArduPilot's handler `fabsf()`s it, but another stack may not); `yaw=Default` sends NaN for p3, which the spec doesn't annotate as NaN-capable — `ORBIT_YAW_BEHAVIOUR_UNCHANGED` is its actual "leave it alone" value; and `frame=` permanently rewrites the shared `flytoframe` setting, which the Orbit Here popup always emits, so every map orbit silently reassigns the frame Fly To will use next.

</details>

tridge:
@peterbarker fix the ValueError from AI review

tridge:

<details><summary>Previous review (2026-09-07)</summary>

Re-reviewed at head `369df8574b` (previously `145e62aa3e`); my earlier comment above is superseded.

## Everything I asked for last round landed cleanly. **But COMMENT → REQUEST CHANGES** for one new thing: as it stands this makes MAVProxy fail to start for anyone on a released pymavlink.

[Full report](https://uav.tridgell.net/DevCallReviews/followups/2026_09_07_2029/devcall_pr_reviews.html#prMAVProxy-1702)

### Blocking: module-scope enum references kill MAVProxy on any published pymavlink

`MAVProxy/modules/mavproxy_mode.py:20-25` dereferences `mavutil.mavlink.ORBIT_YAW_BEHAVIOUR_*` at **import time**. Those constants exist in `ArduPilot/mavlink` master but in no published pymavlink: the latest on PyPI is **2.4.49 (2025-08-01)**, and its bundled `common.xml` has **zero** occurrences of `ORBIT_YAW_BEHAVIOUR` and zero of `MAV_CMD_DO_ORBIT` — I downloaded the sdist and grepped it.

Confirmed by execution against a real released wheel (2.4.43):

```
hasattr(mavutil.mavlink, 'ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER') -> False
AttributeError: module 'pymavlink.dialects.v10.ardupilotmega' has no attribute
                'ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER'
```

`mode` is in the default module list and `load_module()` catches only `ImportError`, so the `AttributeError` propagates and takes MAVProxy down; `mavproxy_map/__init__.py:17` now imports `mavproxy_mode` too, so the map goes with it. `setup.py` asks for `pymavlink>=2.4.14`, so pip installs a version that cannot work. CI can't catch it — the Python job runs flake8, and the Windows job packages without executing module bodies.

Reached independently three ways (a direct execution test plus both review passes).

Fix: `getattr(mavutil.mavlink, 'ORBIT_YAW_BEHAVIOUR_...', <value>)`, or build `orbit_yaw_behaviours` lazily inside `cmd_orbit` and skip registering the command when the constants are absent — plus raise the `setup.py` floor once a pymavlink release carries DO_ORBIT.

### Previous round

| Previous finding | Status |
|---|---|
| `velocity=`/`yaw=` raise an uncaught `ValueError` from a GUI-reachable path | **RESOLVED**, proven by execution. The new converters are the `type` facet, so `parse_key_value_args()`'s `except ValueError` catches them; `velocity=` and `yaw=` yield NaN, `velocity=abc` gives a clean message plus usage. Zero exceptions across the set |
| Negative `orbits` sent as a negative param4 | **RESOLVED** — `abs()`, with the direction comment |
| `frame=` permanently rewrote the shared `flytoframe` setting | **RESOLVED** — resolved locally now; all three frames tested, setting untouched |
| No mode change requested, and the NACK isn't surfaced | **HALF OF THIS WAS WRONG, sorry.** "Nothing reads COMMAND_ACK" is false — `mavproxy_link.py` prints `Got COMMAND_ACK: DO_ORBIT: FAILED` and `should_show_command_ack()` doesn't filter it. Only the mode-change half stands, below |
| `yaw=Default` sends NaN for param3 | **WITHDRAWN.** Both passes refuted it: the command-level description in `common.xml` explicitly permits NaN/INT32_MAX to mean defaults, and param3 not repeating that wording doesn't make NaN forbidden. Please ignore it from last round |
| Cross-link the two PRs | **RESOLVED** |

### Still open, re-raised

**No mode change is requested**, so from anything but Guided the command fails: the companion firmware PR's `handle_MAV_CMD_DO_ORBIT()` opens with `if (!copter.flightmode->in_guided_mode()) return MAV_RESULT_FAILED;`. With the correction above, this is about the mode prerequisite and console-only feedback rather than an unread ACK — a map-menu user gets a console line they're unlikely to be watching.

### Nits

- `mp_module.py` — `parse_key_value_spec()` folds synonyms in *spec* order, not command-line order: `radius=50 lng=1 longitude=2 lon=3` and the reversed form both yield `lng=3`, and the explicit `lng=1` is discarded silently. Erroring when a canonical key and a synonym are both given would be friendlier.
- `mp_menu.py` — `from_arg_spec`'s docstring says a callable default produces "a live default", but it's called once in `MapModule.__init__`, so `guidedalt`/`flytoframe` are frozen at map-load. Matches the existing 'Fly To' item, so only the wording misleads.

### Checked and correct

The parameter packing is right, which is the thing most worth getting right here — re-checked slot by slot against `common.xml`: p1 signed radius, p2 NaN-default velocity, p3 the yaw enum, p4 `units="rad"` from `orbits * 2π`, and x/y as int32 degE7 inside `command_int_send`, which is the correct call for lat/lon. Also cleared: menu ids unique; Cancel returns `None` and `handle_menu_event` drops it; the dialog result crosses back as a string for `process_stdin`, so no cross-process mavlink access; no import cycle.

One scoping note: your four files are `mp_menu.py`, `mp_module.py`, `mavproxy_map/__init__.py` and `mavproxy_mode.py`. The three-dot compare shows ~17 including a whole `mavproxy_camera` module — that's master moving underneath, ignored.

**Verdict: REQUEST CHANGES** — solely for the import blocker.



</details>

tridge:

Re-reviewed at head `173c9607fc` (previously `369df8574b`); my earlier comment above is superseded.

## The blocker is fixed. **REQUEST CHANGES → COMMENT.**

[Full report](https://uav.tridgell.net/DevCallReviews/followups/2026_09_08_1730/devcall_pr_reviews.html#prMAVProxy-1702)

I installed pymavlink **2.4.49** (the latest release on PyPI) into a venv and imported the module at this head: `IMPORT OK`. All six `getattr` fallbacks match `common.xml` exactly — I checked each against the `ORBIT_YAW_BEHAVIOUR` enum twice, from two independent copies: FaceCentre 0, InitialHeading 1, Uncontrolled 2, Tangent 3, RCControlled 4, Unchanged 5. `mavproxy_map` is unblocked with it.

### Still open: the same treatment was not applied to `MAV_CMD_DO_ORBIT` itself

`MAVProxy/modules/mavproxy_mode.py:308`. A full `mavutil.mavlink.*` symbol audit of all four changed files against 2.4.49 turns up exactly one missing name, and it is that one. Executed at this head:

```
=== orbit radius=50
Orbit (-35.363261, 149.16523) radius 50.0m alt 100.0 frame 3
ERROR in command ['radius=50']: module 'pymavlink.dialects.v10.ardupilotmega'
                                has no attribute 'MAV_CMD_DO_ORBIT'
```

Not fatal — `process_stdin` wraps the dispatch in `except Exception` — so this is much less severe than last round. But the user gets a success-looking `Orbit ...` line followed by an obscure traceback message, and `setup.py:29` still floors at `pymavlink>=2.4.14`, so `pip install MAVProxy` resolves to a version where the command cannot work. One line: `getattr(mavutil.mavlink, 'MAV_CMD_DO_ORBIT', 34)` at `:308` and `:327` (34 confirmed from `common.xml`), or raise the `setup.py` floor once a release carries it.

Worth knowing: the reference at `:327` is currently unreachable on old pymavlink only because `:325`'s `orbit_ack_pending` guard runs first and the flag can never be set. It is protected by ordering, not by design.

### New, on the ack handling

- **`mavproxy_mode.py:336` — the GUIDED hint fires for every non-ACCEPTED result.** On any firmware shipping today the reply is `UNSUPPORTED`, and #32434 returns `DENIED` for a bad radius or centre; in both cases "the vehicle must be in GUIDED mode to orbit" is wrong advice. Gate it on `MAV_RESULT_FAILED`.
- **`:329` — `MAV_RESULT_IN_PROGRESS` is treated as a rejection and clears the pending flag**, so the real final ack is then dropped. Both passes injected it and saw `Orbit rejected (IN_PROGRESS)`. There is also no timeout, so after link loss the flag stays set indefinitely, and a second orbit issued before the first ack means the second rejection goes unreported.
- Minor: the docstring at `:323` says the point is to surface the rejection prominently because the generic console line is easily missed by a map-menu user — but the new message uses bare `print()` (terminal) while `mavproxy_link.py:978` uses `self.mpstate.console.writeln()` (console window). For the GUI user that is arguably less visible than what it improves on.

### Previous round

| Previous finding | Status |
|---|---|
| Module-scope enum refs kill MAVProxy on any published pymavlink | **RESOLVED**, verified by execution against 2.4.49; fallback values all correct |
| No mode change requested, so from anything but Guided the command fails | **STILL OPEN.** Ack reporting was added and it genuinely works — I drove a stub `COMMAND_ACK` through the real `mavlink_packet` dispatch and got the rejection line, and an ack for a different command is correctly ignored. But `cmd_orbit` still sends only the COMMAND_INT; there is no `set_mode`. If that is deliberate, the help text is the place to say "switch to Guided first" |
| `parse_key_value_spec()` folds synonyms in spec order | **STILL OPEN**, nit. Both `lng=1 longitude=2 lon=3` and its reverse still execute to `{'lng': 3.0}` |
| `from_arg_spec` docstring claimed a "live default" | **MOSTLY RESOLVED**, nit. "evaluated now" is the right correction; the parenthetical "(when the dialog is built)" is still off — the only call site is `mavproxy_map/__init__.py:126` at menu construction, while the dialog is built later in `mp_menu.py:514 call()`. The `ctx` values are snapshotted at `:128-130` too, so it is doubly frozen |

### Checked and correct

The lat/lon validation is right, including the part that is easy to get wrong: `math.isfinite` is genuinely load-bearing, because `abs(nan) > 90` is `False` and the range test alone would let NaN through. Bounds are inclusive of the poles and antimeridian, `int(90.0*1e7)` and `int(180.0*1e7)` both fit int32, and `(0,0)` is accepted — correct, since `click_location` is `None` rather than `(0,0)` when unset.

Parameter packing re-verified slot by slot against `common.xml` entry 34: p1 signed radius, p2 NaN-default velocity, p3 the yaw enum, p4 `units="rad"` from `abs(orbits)*2π` — executed, 3 orbits gives 18.84955592153876, exactly 6π — and x/y int32 degE7 via `command_int_send`, whose signature and argument order match. The non-`_INT` frame constants are correct for COMMAND_INT. `name[11:]` and its `KeyError` fallback both behave. `mavproxy_mode.py` is flake8-clean under the repo config; CI green on both jobs.

`OrderedDict` is unnecessary on Python ≥3.7, but ordering *is* load-bearing here (dropdown order and the help string) and it matches the file's existing style — no objection.

**Verdict: COMMENT** — nothing here breaks MAVProxy or misconfigures a vehicle. For transparency: the independent Codex pass returned REQUEST CHANGES; I did not adopt it, because re-deriving from the findings that survived checking, the residual is a dead command with a cryptic message on released pymavlink rather than anything that harms a user who is not invoking `orbit`.

PR #1704, opened 2026-07-15
mavproxy: allow 32 bit MAVLink system IDs

tridge:
Lift the source_system and target_system setting ranges for MAVLink2.1 32 bit system IDs. The wire handling comes from pymavlink; targets in the extended TARGETTED header are overlaid onto decoded messages by pymavlink so vehicle switching and forwarding work unchanged.

Requires https://github.com/ArduPilot/pymavlink/pull/1229

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -279,9 +279,9 @@ class MPState(object):
             MPSetting('rally_breakalt', int, 40, 'Default Rally Break Altitude', range=(0, 10000), increment=1),
             MPSetting('rally_flags', int, 0, 'Default Rally Flags', range=(0, 10000), increment=1),
 
-            MPSetting('source_system', int, 255, 'MAVLink Source system', range=(0, 255), increment=1, tab='MAVLink'),
+            MPSetting('source_system', int, 255, 'MAVLink Source system', range=(0, 0xFFFFFFFF), increment=1, tab='MAVLink'),
             MPSetting('source_component', int, 230, 'MAVLink Source component', range=(0, 255), increment=1),
-            MPSetting('target_system', int, 0, 'MAVLink target system', range=(0, 255), increment=1),
+            MPSetting('target_system', int, 0, 'MAVLink target system', range=(0, 0xFFFFFFFF), increment=1),
             MPSetting('target_component', int, 0, 'MAVLink target component', range=(0, 255), increment=1),
             MPSetting('state_basedir', str, None, 'base directory for logs and aircraft directories'),
             MPSetting('allow_unsigned', bool, True, 'whether unsigned packets will be accepted'),
--- a/MAVProxy/modules/lib/mission_item_protocol.py
+++ b/MAVProxy/modules/lib/mission_item_protocol.py
@@ -1113,7 +1113,12 @@ on'''
             w = self.wp_to_mission_item_int(w)
             tlist = []
             for field in mavmsg.ordered_fieldnames:
-                tlist.append(getattr(w, field))
+                value = getattr(w, field)
+                if field == 'target_system' and value > 255:
+                    # the mission file has an 8 bit target_system field,
+                    # which the autopilot ignores on upload
+                    value = 0
+                tlist.append(value)
             tlist = tuple(tlist)
             buf = mavmsg.unpacker.pack(*tlist)
             fh.write(buf)
--- a/MAVProxy/modules/mavproxy_chat/assistant_setup/send_mavlink_command_int.json
+++ b/MAVProxy/modules/mavproxy_chat/assistant_setup/send_mavlink_command_int.json
@@ -6,7 +6,7 @@
     "parameters": {
       "type": "object",
       "properties": {
-          "target_system": {"type": "integer", "minimum":0, "maximum":255, "description": "vehicle autopilot System ID.  can be omitted"},
+          "target_system": {"type": "integer", "minimum":0, "maximum":4294967295, "description": "vehicle autopilot System ID.  can be omitted"},
           "target_component": {"type": "integer",  "minimum":0, "maximum":255, "description": "vehicle autopilot Component ID.  can be omitted"},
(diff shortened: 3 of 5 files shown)


AP-Review:

Full report: [https://uav.tridgell.net/DevCallReviews/2026_09_24_AIReview/devcall_pr_reviews.html#prMAVProxy-1704](https://uav.tridgell.net/DevCallReviews/2026_09_24_AIReview/devcall_pr_reviews.html#prMAVProxy-1704)

Reviewed at head `50c0c12497`. **REQUEST CHANGES** — one confirmed, reproducible regression with a one-line fix. The core premise is sound and the FTP clamp is correct and necessary.

### BUG — the widened range breaks the console Settings dialog completely

`MAVProxy/mavproxy.py:283` and `:285`

`wxsettings_ui.py:135-138` (`add_intspin`) passes `setting.range` straight into `wx.SpinCtrl(min=minv, max=maxv)`, and wxWidgets takes the bound as a C `int` — so `0xFFFFFFFF` arrives as −1, GTK rejects it (`gtk_spin_button_new_with_range: assertion 'min <= max' failed`), the spin button is never created, and wxPython raises:

```
wxAssertionError: C++ assertion "m_widget" failed at .../src/gtk/window.cpp(4340) in DoSetSize(): invalid window
```

Mutation proof — I built the real `SettingsDlg` with only these two settings, varying nothing but the range:

```
RESULT OLD (0,255):         SettingsDlg constructed OK
RESULT NEW (0,0xFFFFFFFF):  RAISED wxAssertionError
```

and a direct three-way probe of the control agrees: `max=255` OK, `max=2147483647` OK (`GetMax()` reads back correctly), `max=4294967295` raises.

Reachable path: `mavproxy_console.py:118` `MPMenuItem('Settings')` → `:226` `wxsettings.WXSettings(self.settings)` → `child_task` → `SettingsDlg`. The exception propagates in the GUI child process and kills it, so **the Settings dialog never appears at all** — not just those two fields — and the parent's `watch_thread` then blocks on `child_pipe.recv()` indefinitely.

This is a new class of failure rather than something pre-existing: a line-wise sweep of all 31 `MPSetting` lines carrying both `range=` and `increment=` finds exactly two with a max above 2^31−1, the two this PR adds. Both hit `add_intspin` because both kept `increment=1`, which is the branch selector at `wxsettings_ui.py:182`.

**Suggested fix:** drop `increment=1` from the two settings, so `SettingsDlg` falls through to `add_text`, which accepts any value. Clamping `add_intspin` to `2**31 - 1` would stop the crash and fix the whole class, but it would then silently refuse to *enter* a legal ID above that bound — so the text control is the better answer for these two.

Tested on wxPython 4.3.1 / GTK. The MSW port also declares the SpinCtrl bounds as `int`, so it is expected to fail there too, but I could not test Windows.

### Notes

- `setup.py:29` — `pymavlink>=2.4.38` is not bumped, so the widened range ships against a pymavlink that cannot honour it. With a released pymavlink, `set target_system 300` is now accepted by the range check and a *targeted* send then raises `struct.error: 'B' format requires 0 <= number <= 255`, where pre-PR the range check rejected it cleanly. (Not every send — an untargeted HEARTBEAT still packs fine; I overstated that on the first pass.) Your workflow runs a job titled "Run regression tests with released pymavlink", so this is the combination the merged tree ships. Severity is modest, because `MPSettings.__setattr__` bypasses `MPSetting.set()` and therefore the range check entirely, so `--target-system 300` and `vehicle 300` already bypassed the bound before this PR — the regression is confined to the `set` command and the GUI. You do state the #1229 dependency, so this is about the version floor rather than an undisclosed requirement.
- `MAVProxy/mavproxy.py:918` — `getattr(m, 'target_system', -1)` misses a header-only target. #1229's `decode()` overlays the header target onto the decoded message only when the message type has a payload target field, and #1229 explicitly supports messages carrying a header target with no payload target field — `set_target()` works on any message. For those, `getattr` returns −1 and the slave-forwarding link selection falls back to the default link. `m.get_target_system()` handles both. Nothing currently emits a header-only target, so this is latent — but it is the one place the "targets are overlaid onto decoded messages" claim is narrower than it reads.
- The new clamp is untested: `grep -rln "ftp_upload|wp_ftp|mission_item_protocol|oldwp" tests/` returns nothing. A small unit test over the `ordered_fieldnames` pack loop would be cheap.

### Checked and clear

**The released-pymavlink import test is clean**, which is the headline check for this repo — a module-scope MAVLink constant raises `AttributeError` on import, `load_module` does not catch it, and CI cannot see it. Against pymavlink 2.4.49, all of `mavproxy`, `mission_item_protocol`, `oldwp`, `wp`, `fence`, `rally`, `chat`, `mp_settings`, `wxsettings_ui`, `link` and `console` import cleanly; the PR adds no reference to any pymavlink symbol at any scope.

**The 8-bit narrowing sweep is clean, with no head-capped greps**: no `& 0xFF` on any system ID, no `chr()`/`bytes([...])`/`%c` applied to a sysid, and no `struct.pack` of a system ID anywhere except the two mission-file writers you fixed. Vehicle state is keyed on the full Python int everywhere — `mav_param_by_sysid[(target_system, target_component)]`, `mav.sysid_state[sysid]`, the `(srcSystem, srcComponent)` tuples, `lat_lon_heading[get_srcSystem()]` — so two vehicles cannot collide on a key.

The clamp itself is correct and necessary, mutation-proved in both copies: with `target_system=300` and no clamp, `struct.error: 'B' format requires 0 <= number <= 255`; with it, packs cleanly. Your comment's justification checks out on the ArduPilot side too — `finish_upload_mission` and the fence/rally converters never read `target_system`. Forwarding and re-signing preserve the 32-bit header target (`decode()` sets `_target_is_received` and `_pack()` re-asserts the flag). Range bounds match #1229's own guard exactly with no off-by-one, and 0 remains special. All 18 chat schema files parse, and exactly two carry `target_system`, both with the new max — no stale 255. Test suite: **286 passed, 5 skipped, 72 subtests** under Xvfb against released pymavlink; flake8 exits 0. CI all green.

PR #1712, opened 2026-07-25
output: add per-output component ID discard

tridge:
"output discardcompid N X" stops messages from component X being sent to output N, and "output discardcompid N clear" removes the restriction.

This came from remote bin logging (LOG_BACKEND_TYPE=3), where the flight controller streams log data from MAV_COMP_ID_LOG (155). MAVProxy fans every master message out to all outputs, so on an aircraft with both a high-bandwidth link and a low-bandwidth backup link that log traffic swamps the slow one. Filtering has to be per output rather than global, because the messages must still reach the links that can carry them, and it cannot be done on the master side for the same reason.

The list is applied in the master->outputs fan-out, which runs before module dispatch, so modules still receive the messages and local logging is unaffected - only the link is relieved.

It is attached to the connection object rather than kept in a table keyed by output index, because "output remove" shifts the indices of later outputs; an index-keyed table would silently re-associate a discard list with the wrong link. Outputs created by --out never get MAVProxy attributes, so the fan-out reads it with getattr().

An unknown output number reports the valid range and changes nothing, rather than silently doing nothing - a stale index in a startup command would otherwise leave the flood running and look like it had worked. "output list" shows any non-empty discard list.

### DIFF
--- a/MAVProxy/modules/mavproxy_link.py
+++ b/MAVProxy/modules/mavproxy_link.py
@@ -1104,6 +1104,11 @@ class LinkModule(mp_module.MPModule):
             if self.mpstate.settings.mavfwd_rate or mtype != 'REQUEST_DATA_STREAM':
                 if mtype not in self.no_fwd_types:
                     for r in self.mpstate.mav_outputs:
+                        # per-output component discard ("output discardcompid N X"). The
+                        # attribute is absent on --out connections, so read it defensively.
+                        discard = getattr(r, 'discard_comps', None)
+                        if discard and m.get_srcComponent() in discard:
+                            continue
                         if hasattr(r, 'ws') and r.ws is not None:
                             from wsproto.connection import ConnectionState
                             if r.ws.state != ConnectionState.OPEN:  # Ensure Websocket handshake is done
--- a/MAVProxy/modules/mavproxy_output.py
+++ b/MAVProxy/modules/mavproxy_output.py
@@ -4,6 +4,8 @@
     output add 10.11.12.13:14550
     output list
     output remove 3      # to remove 3rd output
+    output discardcompid 1 155   # don't send messages from component 155 to output 1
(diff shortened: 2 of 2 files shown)


peterbarker:
On Fri, 24 Jul 2026, Andrew Tridgell wrote:

> "output discardcompid N X" stops messages from component X being sent to output N, and "output discardcompid N clear" removes the restriction.
> 
> This came from remote bin logging (LOG_BACKEND_TYPE=3), where the flight controller streams log data from MAV_COMP_ID_LOG (155). MAVProxy fans every master message out to all outputs, so
> on an aircraft with both a high-bandwidth link and a low-bandwidth backup link that log traffic swamps the slow one. Filtering has to be per output rather than global, because the
> messages must still reach the links that can carry them, and it cannot be done on the master side for the same reason.
> 
> The list is applied in the master->outputs fan-out, which runs before module dispatch, so modules still receive the messages and local logging is unaffected - only the link is relieved.
> 
> It is attached to the connection object rather than kept in a table keyed by output index, because "output remove" shifts the indices of later outputs; an index-keyed table would
> silently re-associate a discard list with the wrong link. Outputs created by --out never get MAVProxy attributes, so the fan-out reads it with getattr().
> 
> An unknown output number reports the valid range and changes nothing, rather than silently doing nothing - a stale index in a startup command would otherwise leave the flood running and
> look like it had worked. "output list" shows any non-empty discard list.

I'd *really* rather see a per-message-type filter than a by-component-id 
filter.

Sending from 155 is something I want to *stop* ArduPilot doing!

Shape change on this PR might be terming things in terms of adding a 
filter rather than an explicit "discard fom this component ID".  As in 
"output filter X discard fromcompid=N" sort of thing

tridge:
> I'd *really* rather see a per-message-type filter than a by-component-id filter.

if you'd like to add that as well then great, do it, I still want this capability

PR #1722, opened 2026-08-06
mavproxy.py: correct shutdown when logging queue is full

peterbarker:
MAVProxy could fail to exit on SIGTERM for two independent reasons:

1. **Log queue feeder threads blocking exit.** The telemetry log queues are
   `multiprocessing` queues, each with a feeder thread that can block writing
   into a full pipe. `multiprocessing`'s atexit handler joins those threads,
   so a blocked one hung MAVProxy forever at exit, after every module had
   already been unloaded. The log writer now waits briefly for `main_loop` to
   stop, drains whatever it can from the queues within a bounded time, and
   unconditionally cancels the feeder thread joins so exit can never block on
   them. Write or flush failures during the drain (e.g. disk full) are
   reported rather than skipping the cancellation, and a warning is printed
   if the drain times out or `main_loop` is still running, since the end of
   the log may then be missing.

2. **Signal ignored while waiting for input.** When the main thread was
   blocked in `input()` waiting for a command, the signal handler set
   `stop_event` but the read was simply restarted (PEP 475), so the signal
   went unnoticed until the user pressed enter. The handler now raises
   `ShutdownRequested` to abort the read.

### DIFF
--- a/MAVProxy/mavproxy.py
+++ b/MAVProxy/mavproxy.py
@@ -25,6 +25,7 @@ import time
 import traceback
 
 from importlib import reload
+from queue import Empty
 
 from pymavlink import mavutil
 
@@ -68,6 +69,12 @@ if __name__ == '__main__':
 # The MAVLink version being used (None, "1.0", "2.0")
 mavversion = None
 mpstate = None
+
+
+class ShutdownRequested(Exception):
+    '''raised in the main thread by the fatal-signal handler to break it
+    out of the blocking read for user input'''
+    pass
 
 
 class MPStatus(object):
@@ -962,6 +969,73 @@ def log_writer():
             mpstate.logfile.flush()
             mpstate.logfile_raw.flush()
 
+    # main_loop only notices stop_event at the top of each iteration,
+    # so it can still be logging packets; give it a bounded time to
+    # finish before we decide the queues are empty.  It is created
+    # after this thread, so may not exist yet on an early exit:
+    main_thread = getattr(mpstate.status, 'thread', None)
+    if main_thread is not None and main_thread is not threading.current_thread():
+        main_thread.join(timeout=1)
+        if main_thread.is_alive():
+            print("main_loop still running at exit; the end of the log may be missing")
+
+    drain_log_queues()
+
+
+def drain_log_queues(timeout=2):
+    '''write out whatever is left in the log queues as we shut down.
+
+    These are multiprocessing queues, so each has a feeder thread which
+    will still be blocked writing into a full pipe if we stop reading
+    with data still queued.  multiprocessing's atexit handler joins
+    those feeder threads, so leaving one blocked hangs MAVProxy forever
+    at exit - after every module has already been unloaded.
+    '''
+    tend = time.time() + timeout
+    try:
+        # main_loop should have stopped putting to the queues by now
+        # (module threads may still be sending; we do not wait for
+        # those), but a feeder thread can be part-way through pushing
+        # what it has into the pipe, so a single empty queue does not
+        # mean we are done; wait for two idle passes:
+        idle_passes = 0
+        while idle_passes < 2 and time.time() < tend:
+            wrote = False
+            for (queue, logfile) in [(mpstate.logqueue_raw, mpstate.logfile_raw),
+                                     (mpstate.logqueue, mpstate.logfile)]:
+                # empty this queue before moving on to the other one;
+                # the timed get only costs us its full timeout when the
+                # queue is idle, so this never throttles the drain:
+                while time.time() < tend:
+                    try:
+                        data = queue.get(block=True, timeout=0.01)
+                    except Empty:
+                        break
+                    logfile.write(bytearray(data))
+                    wrote = True
+            if wrote:
+                idle_passes = 0
+            else:
+                idle_passes += 1
+        if time.time() >= tend:
+            print("Timed out draining log queues; the end of the log may be missing")
+        mpstate.logfile.flush()
+        mpstate.logfile_raw.flush()
+    except Exception as ex:
+        # e.g. disk full or file closed - the very conditions under
+        # which a log queue backs up in the first place
+        print("Failed to drain log queues: %s" % ex)
+    finally:
+        # a feeder thread can still be part-way through writing
+        # something the drain above did not see, so make sure exiting
+        # can never block on one.  This must run regardless of how the
+        # drain went, or the hang we are here to fix comes back:
+        for queue in [mpstate.logqueue, mpstate.logqueue_raw]:
+            try:
+                queue.cancel_join_thread()
+            except Exception:
+                pass
+
 
 # If state_basedir is NOT set then paths for logs and aircraft
 # directories are relative to mavproxy's cwd
@@ -1484,6 +1558,16 @@ if __name__ == '__main__':
             sys.exit(0)
         else:
             mpstate.status.stop_event.set()
+            if quit_handler.in_input_loop:
+                # the main thread is blocked in readline waiting for a
+                # command.  That read is simply restarted when we return
+                # from here (PEP 475), so the flag we just set would go
+                # unnoticed until the user happened to press enter -
+                # meaning we would never act on the signal at all.
+                # Raising instead aborts the read:
+                raise ShutdownRequested()
+
+    quit_handler.in_input_loop = False
 
     # Listen for kill signals to cleanly shutdown modules
     fatalsignals = [signal.SIGTERM]
@@ -1640,7 +1724,16 @@ if __name__ == '__main__':
             if opts.daemon or opts.non_interactive:
                 time.sleep(0.1)
             else:
-                input_loop()
+                quit_handler.in_input_loop = True
+                try:
+                    input_loop()
+                finally:
+                    quit_handler.in_input_loop = False
+        except ShutdownRequested:
+            # a fatal signal arrived while we were waiting for input;
+            # stop_event is already set, so fall through to the cleanup
+            # below
+            pass
         except KeyboardInterrupt:
             if mpstate.settings.requireexit:
                 print("Interrupt caught.  Use 'exit' to quit MAVProxy.")


tridge:

<details><summary>Previous review</summary>

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_01_AIReview/devcall_pr_reviews.html#prMAVProxy-1722

Reviewed at head `ce294439d0`. **COMMENT** — no blockers. Both diagnoses are right and the mechanism is sound; two things would make the guarantee actually hold.

**1. A flush failure skips the cancellation and the original hang comes back.** The two `flush()` calls at the end of `drain_log_queues()` aren't inside a `try`, and they precede the `cancel_join_thread()` loop. If either raises — disk full, closed file, i.e. the very conditions under which a log queue backs up in the first place — cancellation never runs and MAVProxy hangs at exit exactly as before. The `logfile.write()` inside the drain loop propagates the same way. Putting the cancel loop in a `finally` makes it unconditional.

**2. The alternating 10 ms timeouts throttle the drain to roughly 200 records/second.** Each pass does a `get(timeout=0.01)` on *both* queues, so whenever one is idle it contributes a full 10 ms to every pass. Measured with your loop's exact shape: one queue empty, 500 records on the other, 2-second budget → **198 of 500 drained**; the other 302 are then silently discarded by `cancel_join_thread()`. Draining each queue until empty before switching, or a much shorter timeout, recovers most of it. Related: when the budget expires the log tail is dropped with no message — reporting that the drain timed out would tell the user their log is short rather than leaving them to find out later.

**Checked and correct, each of which could plausibly have been a bug:**

- `from queue import Empty` is the right exception for a *multiprocessing* queue — confirmed by running it, since importing the wrong `Empty` would make the drain silently never catch anything.
- `log_writer` only starts after `logfile`/`logfile_raw` are opened, so the drain's preconditions hold.
- `ShutdownRequested` genuinely reaches its handler: `input_loop` catches only `(EOFError, IOError)`, `rline.input()` on POSIX is a bare `return input(...)`, and the file's only broad `except Exception` is off the input path.
- `stop_event` is set *before* the raise, so the outer `while` exits rather than re-entering the prompt.
- The window between setting `in_input_loop` and the inner `try` is covered by the outer `try`.
- `SIGINT` is only added to `fatalsignals` in daemon/non-interactive mode — which is exactly the branch that never calls `input_loop()` — so the new raise path can't fire for it. That's a nice property and worth keeping in mind if the signal set ever changes.

One gap I did not close: the Windows path is different machinery. There `rline.input()` goes through `prompt_toolkit`'s `session.prompt()` inside `patch_stdout()` rather than plain `input()`, and raising from a signal handler through that may behave differently or leave the terminal dirty (`SIGQUIT` also doesn't exist there). Stating it as unverified rather than as a defect — the POSIX path is what I checked.



</details>

tridge:
@peterbarker can you fix the flush error in AI review?

tridge:

<details><summary>Previous review (2026-09-04)</summary>

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_04_1358/devcall_pr_reviews.html#prMAVProxy-1722

Re-reviewed at head `9afe8926a6`; my earlier comment above is superseded. **Both findings from the previous round are resolved**, and one of them is measurably fixed. The verdict still moves COMMENT → **REQUEST CHANGES**, and I want to be clear it is not on a review argument: the cold cross-check *reproduced the exit hang this PR exists to prevent*, through a path the new `finally` sits below.

### The throttle is gone, by about 33×

I re-ran my measurement from last round against both loop shapes with identical inputs:

```
queues(0,500) budget 2s: old drained 198 in 2.01s | new drained 500 in 0.06s
queues(500,0) budget 2s: old drained 197 in 2.01s | new drained 500 in 0.06s
queues(250,250) budget 2s: old drained 500 in 0.04s | new drained 500 in 0.06s
```

The 198 matches what I reported last time exactly, and the cross-check independently measured 198/302 and 500/0.062 s for the same pair. Note the symmetric case was never throttled — the fix targets precisely the asymmetric case, which is the realistic one. The timeout message is there too.

And the cancellation is unconditional within the drain, which is what I asked for. The `main_thread.join(timeout=1)` you added on top wasn't something I raised, and it's right: I checked that `mpstate.status.thread` is assigned the `main_loop` thread at line 1716 and nowhere else, and that both queue producers — `process_master()` and the forwarding path — run on that thread, so a successful join really does stop the main producer.

### The hang is still reachable

**An ordinary log-write failure kills `log_writer` before the `finally` is ever entered.** The main loop of `log_writer()` — lines 956-970 — does `logfile_raw.write()`, `logfile.write()` and two `flush()` calls with **no `try` anywhere**. The new protection lives inside `drain_log_queues()`, which is only reached if that loop exits normally. So a disk-full write during ordinary operation propagates out of the thread, and neither the join, nor the drain, nor the `cancel_join_thread()` loop ever runs.

Both passes found this independently, and the cold one reproduced it rather than arguing it: real multiprocessing queues, a simulated disk-full `write()`, a backed-up 20 MiB queue — the logging thread died with `_joincancelled == False` and interpreter shutdown hung until a four-second timeout killed it. That's this PR's own failure mode, in the disk-full case the description names, and it falsifies "exit can never block on them". The fix is the shape you already applied one function down.

### One queue can still starve the other — and I had this wrong

I reasoned that starvation needed a producer to keep feeding the first queue, and concluded it was unreachable once the join succeeds. The cold pass found the mechanism I missed: **the writes themselves being slow is enough.** The inner `while time.time() < tend` only exits on `Empty` or the deadline, so with slow storage the raw queue can consume the whole two-second budget and the cooked queue is never visited at all — measured at 14 raw entries written, zero cooked, with the cooked entry still queued when `cancel_join_thread()` discarded it. `idle_passes` gives no fairness, since it's only updated after both inner loops. The single outer `try` compounds it: a raw-log failure jumps to the `except` and skips the cooked drain and both flushes.

### Minor

- **The two-second budget isn't a hard bound.** The deadline guards loop *entry* only; `write()` and the flushes can block indefinitely, so stuck storage can still delay shutdown past it. Worth a word in the description, which promises a bounded drain.
- **A narrow race worth one line.** `mpstate.status.thread` is assigned, configured and started in three separate statements (1716-1718), and `log_writer` starts before that. If it observed the attribute between the assignment and `.start()`, `join()` would raise `RuntimeError: cannot join thread before it is started` — and since the join precedes the drain, that skips the cancellation too. It needs shutdown requested in the microseconds between two adjacent startup statements, so it's very unlikely; adding `main_thread.is_alive()` to the existing guard closes it, since that returns False for an unstarted thread rather than raising.

### Checked and clear

The shutdown paths themselves hold up: the cold pass walked SIGTERM, SIGINT, stdin EOF and module exceptions in both interactive and daemon modes with full-process tests on POSIX; all terminate as intended, exit codes unchanged, and an orderly shutdown costs about 40 ms of extra idle polling while preserving queued data. `from queue import Empty` re-confirmed by running it. 28/28 unit tests and flake8 pass, and the Windows build passes.

My previous comment left the Windows `prompt_toolkit` path explicitly unverified. The cold pass inspected it and ran an equivalent `prompt_toolkit` test showing `ShutdownRequested` propagates through `rline.py:84` — but did not execute on Windows, so it stays unconfirmed rather than closed. Separately: Windows `os.kill(pid, SIGTERM)` uses `TerminateProcess`, which bypasses Python's handler and all cleanup regardless of this PR.


</details>

AP-Review:

Re-reviewed at head `39bba66feb` (you were last told `9afe8926a6`); my earlier comment above is superseded. Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_16_1759/devcall_pr_reviews.html#prMAVProxy-1722

(My previous comment above is superseded by this one. It's left un-collapsed because these reviews now post from the `AP-Review` account, which can't edit the earlier comments posted under `tridge`.)


**All four findings from my previous comment are resolved, and every one of them is proven by reverting the fix and watching the failure come back. Verdict REQUEST CHANGES → COMMENT.** Two new things are worth a look, both one-liners; neither is a regression against master.

### Previous round — all resolved

- **RESOLVED: an ordinary log-write failure killed `log_writer` before the cleanup ran.** The `try` at `mavproxy.py:956` with its handler at `:972-977` closes it. Reproduced with real multiprocessing queues, a 20 MiB backed-up raw queue and an `ENOSPC` `write()`, `stop_event` unset:

  ```
  fixed (PR head):  _joincancelled=True  on both queues, exit rc=0  in 1.61s
  try/except reverted: _joincancelled=False on both, hung until SIGKILL at 30s (rc=124)
  ```

- **RESOLVED: one queue could starve the other.** `share = max(0, tend - time.time()) / len(queues)` at `:1050`, applied at `:1052-1054`, and the per-queue `try` moved down into `drain_log_queue()` at `:1005-1019` with the failing queue dropped at `:1059`. Same inputs to old and new code, 0.14 s per write:

  ```
  OLD  in(raw=100,cooked=1)  -> raw=15 cooked=0    (reproduces the 14/0 I measured last round)
  NEW  in(raw=100,cooked=1)  -> raw=14 cooked=1
  OLD  in(raw=500,cooked=500)-> raw=199 cooked=0
  NEW  in(raw=500,cooked=500)-> raw=99  cooked=100
  ```

  Reverting just `time.time() + share` back to `tend` restores `raw=15 cooked=0`, so that line is load-bearing. The second half is fixed too — with the raw `write()` failing immediately, the old code wrote nothing and flushed nothing, the new one still drains the cooked log and does both flushes.

- **RESOLVED: the two-second budget is not a hard bound.** The docstring at `:1032-1033` now says so, and says it accurately.

- **RESOLVED: the unstarted-thread race.** The guard at `:986-988` adds `main_thread.is_alive()`, which returns `False` for an unstarted thread where `join()` raises `RuntimeError` (checked directly on 3.14.4). The window is real — `log_writer` starts at `:1154` and `mpstate.status.thread` is assigned at `:1765`, made daemon at `:1766` and started at `:1767` — and short-circuiting on `is_alive()` closes it.

The core fix measures well. SIGTERM at an interactive prompt, 20 consecutive runs: **0.151 s to exit with all 23 modules unloaded**, against **30 s and SIGKILL** on the parent commit. Terminal `ECHO`/`ICANON` are correctly restored.

### New — worth fixing in this pass

- **BUG: the signal is still lost if it lands while the readline completer is running.** `rline.py:333` installs a Python completer, and CPython calls it from `on_completion()` in `Modules/readline.c`, which does `PyErr_Clear()` on any exception — so the `ShutdownRequested` raised at `mavproxy.py:1617` is silently discarded. Both my passes found this independently and both reproduced it end to end. Real MAVProxy at this head, TAB over a 100 000-entry directory (MAVProxy's own `complete_filename` takes 249 ms there), SIGTERM delivered during the completion:

  ```
  SIGTERM  61 ms after TAB -> still alive 12 s later, rc=-9, 0 modules unloaded
  SIGTERM 101 ms after TAB -> still alive 12 s later, rc=-9, 0 modules unloaded
  SIGTERM 162 ms / 201 ms after TAB -> same
  control A (same text, no TAB)            -> 23 unloads, rc=1, exit 0.122s
  control B (TAB over a small directory)   -> 23 unloads, rc=1, exit 0.121s
  ```

  The window is only open while someone is holding TAB, so a `systemctl stop` of an unattended MAVProxy will almost never hit it — but when it does, the process hangs forever with nobody at the keyboard, which is the failure this PR exists to remove. Worse, the escape hatch makes it unclean: a second SIGTERM takes the `sys.exit(0)` at `:1607` and discards the module unload entirely.

  A fix that keeps your design — don't let the second signal be worse than the first:

  ```python
  if mpstate.status.stop_event.is_set():
      if quit_handler.in_input_loop:
          # the first raise was lost (e.g. inside the readline completer,
          # where CPython clears exceptions); try again
          raise ShutdownRequested()
      print('Clean shutdown impossible, forcing an exit')
      sys.exit(0)
  ```

  If you take that, also move `quit_handler.in_input_loop = True` (`:1776`) inside the `try`, or reset it in the `except ShutdownRequested` handler — a signal landing between the assignment and the `try` leaves it stale `True`, which is harmless today but would matter once the flag gates a re-raise.

- **ISSUE: `mavproxy.py:977` — the error-reporting `print()` sits outside any `finally`, so if it raises, the cleanup it is announcing never runs.** `drain_log_queues()` at `:993` is the only caller of `cancel_join_thread()`, and nothing between `:977` and `:993` is guarded — the `print()` at `:991` and the `join()` at `:989` are equally exposed. Confirmed that `print()` to a pipe whose reader has closed raises `BrokenPipeError`, which is an `Exception`; my cross-check reproduced the full hang that way, with `_joincancelled == [False, False]` and the process killed after 4 s. Narrow trigger, but it is the exact hang this PR fixes, reappearing one level up from where you fixed it. A `try:` around `:979-993` with `drain_log_queues()` in a `finally:` closes it — the same shape you already used inside `drain_log_queues()`.

### Notes (non-blocking)

- **A second fatal signal during the 150 ms cleanup truncates the module unload.** Once `stop_event` is set, any further fatal signal takes `sys.exit(0)` at `:1607`; `SystemExit` is a `BaseException`, so `except ShutdownRequested` at `:1781` does not catch it and the unload loop at `:1810-1813` stops part-way:

  ```
  2nd SIGTERM 10ms after the first: "... Unloading module fenceClean shutdown impossible, forcing an exit"
                                    (ftp/param/relay/... never unloaded)
  2nd SIGTERM 50ms after the first: "Exception ignored while joining a thread in _thread._shutdown():
                                     ... SystemExit: 0"
  ```

  Not a regression — before this PR the first SIGTERM did nothing and the second gave the same unclean exit with zero unloads — but the PR creates a real window in which an impatient `killall` can now cost you the cleanup it just added. Worth deciding deliberately.

- **After a write failure with `stop_event` unset, logging is silently dead and both queues grow without bound.** The `except` at `:972` joins, drains, cancels and returns while `main_loop` keeps putting. Measured at ~245 records/s of 1 KiB on each queue: feeder threads stay alive and blocked in the pipe, `put()` never blocks (`SEM_VALUE_MAX`), so it is the queues' internal `_buffer` deques that grow — ~260 KB/s of permanently leaked RSS, all discarded at exit. **Not a regression:** the old code died with a traceback and leaked at the same slope, *and* hung at exit (rc=124 after 120 s vs rc=0 in 31 s here). Purely a possible follow-up — e.g. clearing `mpstate.logqueue`/`logqueue_raw` after the cancellations so `:863` and `:930` stop enqueueing.

- **The drain's messages are worded for shutdown but are now reachable mid-flight.** With a failing raw log, a *healthy* cooked log and `stop_event` unset, the user gets "Timed out draining log queues; the end of the log may be missing" in the middle of a flight — and the healthy tlog stops too (measured: cooked records written froze at 806). Again the old code did the same and then hung, so it is the wording that is newly misleading, not the behaviour.

- **Interactive Ctrl-C still bypasses all of this.** `fatalsignals` (`:1622-1629`) only adds SIGINT under `--daemon`/`--non-interactive`, so an interactive Ctrl-C goes to `except KeyboardInterrupt` at `:1786` and exits with **0 modules unloaded** (measured; identical on the parent, so not a regression). Since Ctrl-C is how most interactive users quit, they don't get the clean shutdown this PR builds.

- **The PR description overstates the guarantee.** It says the writer "unconditionally cancels the feeder thread joins so exit can never block on them". `log_writer`'s thread is non-daemon (`:1153-1154`), so `threading._shutdown()` joins it regardless, and an indefinitely-blocking `write()`/`flush()` hangs exit before `multiprocessing`'s atexit is reached — reproduced, rc=124 after 40 s. The docstring already gets this right; only the description needs the softer wording.

- **No regression test.** The failure mode is cheaply testable with a pty, `send_signal` and a bounded wait — the parent commit fails such a test in 30 s and this head passes it in 0.15 s. Given this hang shipped once already, it seems worth locking in.

### What was checked

- Shutdown matrix at this head, all clean unless noted: interactive SIGTERM 0.151 s / 23 unloads, SIGQUIT 0.151 s / 23, SIGINT 0.150 s / 0 (`KeyboardInterrupt` path, same as parent), `requireexit=True` + SIGTERM 0.151 s / 23, `--daemon` 0.187 s / 23, `--non-interactive` 0.180 s / 23, non-tty stdin 0.109 s / 23.
- **`quit_handler` is never called directly** — one `signal.signal` registration at `:1631-1632`, no call site anywhere in `MAVProxy/` or `tests/`, and it is a closure-local inside `if __name__ == '__main__':`, so it cannot be reached by import. The "raised in the wrong thread" hazard does not exist.
- **`except ShutdownRequested: pass` does reach the cleanup**, confirmed at bytecode level (`POP_EXCEPT; JUMP_BACKWARD` to the `while` test, which then sees `stop_event` set) and by the 20/20 runs above. No path re-enters `input_loop()`.
- `queues.remove((queue, logfile))` at `:1059` is identity comparison (neither type defines `__eq__`), so it removes the right entry; `share` cannot divide by zero; `open_telemetry_logs()` has one caller and `log_writer` one start site, so no second writer can race a cancelled queue. All queue producers enqueue bytearrays — no integer path that `bytearray(data)` would silently turn into zero bytes.
- flake8 clean; `pytest MAVProxy/tools/ MAVProxy/modules/` 56 passed, 14 subtests passed. Under released pymavlink 2.4.49 the suite shows 8 failures that are present with this PR reverted too (the `MAV_CMD_DO_ORBIT` constant in unchanged code) — not yours.
- CI snapshot: both `build` and `build (3.10)` passing.
- **Windows remains UNCONFIRMED.** `signal.SIGTERM` is accepted by `signal.signal()` there so the registration is harmless, but `os.kill(pid, SIGTERM)` is `TerminateProcess` and bypasses Python handlers entirely, which would make this fix inert on Windows; and the input path there is `prompt_toolkit`, not readline, so the PEP 475 justification in the comment at `:1611-1616` doesn't describe it. I confirmed on Linux that `ShutdownRequested` does escape `prompt_toolkit`'s `session.prompt()` (3.0.53), but that is not a Windows measurement.

**COMMENT** — nothing here blocks merge, and this is a clear improvement on master in every scenario I measured. The completer hang and the unguarded `print()` are the two I'd fix before it lands.



</details>

AP-Review:

Re-reviewed at head `c48e1d9f74` (previously `39bba66feb`); my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_17_0037/devcall_pr_reviews.html#prMAVProxy-1722

---

**Verdict: REQUEST CHANGES** (was REQUEST CHANGES) — every previous finding is fixed, and the new test file is genuinely load-bearing: run against the previous head it gives **14 failed, 5 passed**; here, 19 passed. The problem is the second commit's rewrite of the *normal* logging path.

### Previous round — 8 of 9 resolved

The signal-in-completer bug, the `finally` gap, the second-fatal-signal truncation, the unbounded queue growth after a write failure, the shutdown-worded drain messages, and the interactive Ctrl-C bypass are all fixed, each covered by a test that fails at `39bba66feb`. The stale `in_input_loop` flag is gone entirely, replaced by a stack walk (`:1697-1705`) — better than what I suggested, since it cannot go stale.

Still open: **the PR description**. It is unchanged, still says exit "can never block on them", and does not describe the second commit at all — the SIGURG forwarder, the grace period, interactive Ctrl-C and `DiscardQueue` are undocumented outside the commit message.

**Also: a bug this commit fixes that I missed last round.** At `39bba66feb`, with the default module set loaded, a plain SIGTERM at the prompt hung **4 times out of 8** — the kernel picks the thread, and readline's `select()` runs no bytecode, so the handler never ran. My "20/20 clean" last round was measured with fewer threads and got lucky. At this head the same test passes 8/8.

### BUG — `MAVProxy/mavproxy.py:1029` — a busy raw log starves the tlog indefinitely

`write_logs()` sets `tend = time.time() + 10` at `:1026` and hands the *same* `tend` to `drain_log_queue()` for each log in turn at `:1029`. `drain_log_queue` (`:988-999`) only returns early on `Empty`, and its `queue.get(block=True, timeout=0.01)` never times out while data arrives more often than every 10 ms. So the raw drain consumes the whole window, and when it returns `tend` has passed — the cooked log's `while time.time() < tend` never executes its body. Next iteration, same thing.

Master (`31ac7477:949-963`) shares the 10 s deadline too, but uses `while not queue.empty() and ...`: the drain outruns the producer, the queue empties in microseconds, and the second loop runs. At `39bba66feb` `drain_log_queue` existed but was used *only* by the shutdown drain; this commit is what wires it into the normal path.

Driving the extracted functions directly (one cooked record at t=0, raw fed at a steady rate):

| raw puts/s | this head | `39bba66feb` | master |
|---|---|---|---|
| 80 | 0.01 s | 0.00 s | 0.00 s |
| 100 | 0.97 s | 0.00 s | 0.00 s |
| 120 | **never** (written at 12.01 s, once the feed stopped) | 0.00 s | 0.00 s |
| 300 | **never** | 0.00 s | 0.00 s |
| 1000 | **never** | 0.00 s | 0.00 s |

Real MAVProxy process, `--cmd="set flushlogs 1"`, ~1000 datagrams/s: the tlog froze at 326,762 bytes for **34 consecutive seconds** while the raw log grew normally. RSS climbed ~85 KB/s throughout (123,420 → 126,236 KB) and is flat at both comparison revisions, so the backlog is unbounded.

**Data is lost, not just delayed.** A clean SIGTERM recovers everything via the shutdown drain. An unclean exit does not — `kill -9` after 30 s of load: this head keeps 326,437 bytes of tlog, the previous head 978,527, master 978,462. **67% of the flight record gone** that master keeps, and power loss / OOM kill / GCS crash are exactly what the tlog is for. `flushlogs` stops working entirely, because it flushes a file nothing has been written to.

The threshold is one put per <10 ms on the queue being drained, i.e. ~100/s — over UDP/TCP one datagram is one `recv()` in `process_master` (`:873-874`) and so one raw put, so a param download (every startup), a log or FTP download, raised `SRx` rates, or several links all clear it. It is not raw-specific either: in the first 10 s window the starvation ran the other way, `mav.tlog.raw` at 0 bytes for 10 s while the tlog grew. Whichever queue is non-empty when the iteration begins takes the whole window.

Equal-share budgeting (as `drain_log_queues` does at `:1100`) is **not** enough — it only halves the stall, to 5.00 s. Restoring master's escape hatch works:

```python
logfile.write(bytearray(data))
wrote = True
if queue.empty():
    return (True, wrote)
```

With that, the first cooked write is back to 0.00 s at 300/s and 1000/s, the real process's tlog grows every second (identical to master at t=20 s), and RSS is flat.

### Notes

- `:1793` — `interrupt_handler` raises unguarded where `quit_handler` guards on `in_input_loop(frame)` (`:1811-1815`). Only reachable on Windows, where `POKE_SIGNAL` is `None`; a Ctrl-C at `:2010-2011`, outside the `try` at `:2012`, would escape the loop and skip the module unload. I cannot test Windows, but the guard you already have would close it for free.
- `:2024` — the `except SystemError` net is never exercised (instrumented, 0 hits across all 11 end-to-end tests), which follows from `SIGURG` (23) sorting above `SIGINT`/`SIGQUIT`/`SIGTERM` in `PyErr_CheckSignals`. If it ever *is* taken with a `__cause__` that is neither `KeyboardInterrupt` nor `ShutdownRequested`, `raise` at `:2029` skips exactly the cleanup this commit adds.
- CI never runs the new test. `.github/workflows/` is flake8 (`MAVProxy/` only, `AP_FLAKE8_CLEAN` files only) plus a Windows build; `grep -rn "pytest\|unittest\|tests/" .github/workflows/` is empty. Pre-existing gap, but this hang has now shipped twice, so a step running `tests/` would be worth the same commit.

Also checked: `pytest MAVProxy/tools/ MAVProxy/modules/` 56 passed; flake8 clean on both changed files; no module-scope MAVLink constants added; all of the above tested against released pymavlink 2.4.49, not master.

AP-Review:
**Automated review note — AI-generated (Claude), cross-checked by an independent Codex pass against the live diff.** Please sanity-check before acting.

**Re-reviewed at head `982d0ae79f`** (previously `c48e1d9f74`) — my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_24_1756/devcall_pr_reviews.html#prMAVProxy-1722

**Verdict: COMMENT** (was REQUEST CHANGES). **The blocking bug is fixed, and the fix is at measured parity with master on every axis I could test. Nothing blocking remains.** What is left is the PR description and two CI nits.

### Resolved — the log-starvation bug

Fixed at `MAVProxy/mavproxy.py:1005-1011` — the exact escape hatch I suggested — plus `:1046`, replacing the shared 10 s deadline with a per-log 0.1 s slice.

Time to first cooked write (extracted-function harness, two independent passes):

| raw puts/s | this head | `c48e1d9f` | master |
|---|---|---|---|
| 100 | 0.00 s | 0.09 s | 0.00 s |
| 120 | 0.00 s | 0.46 / **12.00 s** | 0.00 s |
| 300 | 0.00 s | 2.95 / **12.01 s** | 0.00 s |
| 1000 | 0.00 s | **12.01 s** | 0.00 s |

Real process, `--cmd="set flushlogs 1"`, 1000 msg/s for 25 s: this head wrote **1,102,270 B** with a **0 s** longest freeze and flat RSS; `c48e1d9f` wrote 440,633 B with a 15 s freeze and +78.3 KB/s RSS; master wrote 1,102,402 B. **`kill -9` after 30 s: this head keeps 1,322,990 B — byte-identical to master's 1,322,990 B; `c48e1d9f` keeps 0 B.** The 67% loss I measured last round is gone.

Which change does the work, by mutation on the head: removing the escape hatch but keeping the slice bounds the stall at 0.10 s; keeping the hatch and restoring the shared 10 s deadline gives 0.00 s; reverting both brings the 12 s stall back. So the hatch restores master's behaviour and the slice is belt-and-braces. No throughput cost — paced 5,000 and 20,000 records/s, head and master both drain with zero backlog — and idle CPU is *better* than master (1.45% of one core vs 3.25%), since the 1 kHz poll loop is gone.

### Also resolved

- **The unguarded `interrupt_handler` raise** is guarded at `:1809-1815`. Honest caveat: I could not demonstrate the hazard — removing the guard left all tests passing, and 60 SIGINTs via `tgkill` with the forwarder disabled gave 20/20 clean unloads either way. The window is real but too narrow to hit from a test; the guard closes it by construction and costs nothing.
- **The `except SystemError` bare `raise`** is gone: at `:2046-2055` an unexpected `__cause__` reports and sets `stop_event`, falling through to the module unload at `:2070-2078`.
- **CI now runs the new tests.** `.github/workflows/tests.yml:57` runs `pytest tests/ -ra` on push/PR, matrixed over pymavlink master and release; both legs passed at this head (306 passed / 10 skipped each). Mutation confirms it would have caught the original bug — reverting both fixes fails three of the new tests.

### Still open — re-raised, not new

**The PR description.** Re-read at this head immediately before writing this: unchanged. Still undocumented outside the commit messages are the SIGURG forwarder, `SHUTDOWN_GRACE` (`:82`), interactive Ctrl-C unloading modules, `DiscardQueue` (`:972`), the log-starvation fix, the new test file and the new CI workflow. Point 2 is also stale — it says "The handler now raises `ShutdownRequested` to abort the read", which no longer describes the forwarder-driven design.

Two corrections to my own framing here, both from the cross-check and both verified against the source before I accepted them. It is **not** true that `poke_handler` is the only place that raises: with no forwarder (`request_fd is None`, i.e. the Windows path) `interrupt_handler:1814` and `quit_handler:1837` call `raise_to_end_input()` directly — so the description should describe the forwarder *and* its fallback rather than being called simply wrong. And `SHUTDOWN_GRACE` is **not** an automatic timeout: `quit_handler` only reaches `os._exit(0)` on a *repeat* signal arriving after the grace period.

I also **withdraw** the "can never block on them" half of my old complaint: `cancel_join_thread()` is unconditional in the `finally` at `:1083-1101`, so the feeder joins cannot block exit. (The non-daemon `log_writer` at `:1215` can still hold exit on stuck storage, but your docstring at `:1107-1108` already says so — not a finding.)

### New — nits, none blocking

- **`.github/workflows/tests.yml:21` — `continue-on-error` on the release leg would hide a real regression.** The stated reason, that released pymavlink lacks `MAV_CMD_DO_ORBIT`, no longer holds: the constant is indeed absent from 2.4.49, but `tests/test_mission_geometry.py:137,165,168,195` and `MAVProxy/modules/lib/mp_util.py:275` use `getattr(..., 34)` with a fallback, and `tests/test_map_arc_compat.py:14` installs it with `monkeypatch.setattr(..., raising=False)` while the fixture also tests its absence. Locally `pytest tests/` against 2.4.49 gives 311 passed / 5 skipped / 0 failed, and the release leg passed 306/306. As written, a release-only regression reports green.
- **The new workflow largely duplicates a `tests` job that landed in master while this was in flight.** `.github/workflows/python-cleanliness.yml` gained one running `xvfb-run -a … pytest -q tests/` with `pymavlink==2.4.49`, in master commit `6b118e13` — present in your merge base but not in the previous round's, so the commit message's "Nothing ran tests/ in CI" is now stale. It ran on this PR and passed. The genuinely new coverage in `tests.yml` is the **pymavlink-master** leg; the release leg is redundant (and unpinned, where the existing job pins 2.4.49). Worth dropping the release leg or folding the master leg into the existing job.
- **Coverage observation:** deleting the escape hatch at `:1007` while keeping the 0.1 s slice leaves the logging tests passing, because they assert `< 0.5 s` (`tests/test_mavproxy_shutdown.py:484`) and the slice alone bounds the stall at ~0.10 s. Not a bug — the slice independently prevents starvation — but the line most responsible for restoring master's behaviour can be removed silently. Tightening that bound, or asserting against a backlog rather than a deadline, would pin it.

<sub>Checked clean: `flake8` rc=0 on both changed files; `pytest MAVProxy/tools/ MAVProxy/modules/` 56 passed; `pytest tests/` 311 passed / 5 skipped against released pymavlink 2.4.49 (verified by `__file__`/`__version__` in the environment actually used, since `~/.local` shadows the venv here). **The 14 end-to-end signal tests were run 25 times — 25/25, 0 hangs**, i.e. 350 signal-at-the-prompt trials, with no sign of the 4-in-8 readline hang seen at `39bba66feb`. No module-scope MAVLink constants added, so no `AttributeError`-at-import risk against released pymavlink. Not verified: Windows (the no-forwarder path is only simulated by deleting `signal.pthread_kill`); that the `interrupt_handler` guard is load-bearing; the ~100 puts/s threshold is soft and machine-dependent, though the 300 and 1000 results were stable across both passes.</sub>

AP-Review:

Re-reviewed at head `5d7ec185a6`. My earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_24_2221/devcall_pr_reviews.html#prMAVProxy-1722

**All four items from my previous comment are resolved, and the one I was least sure about — whether the tests actually pin the log-starvation fix — is now proven by mutation.** Two non-blocking things below, one of which is a pre-existing hazard rather than anything this PR introduces.

### Resolved

- **The PR description** — rewritten. The SIGURG forwarder and its no-`pthread_kill` fallback, `SHUTDOWN_GRACE`, interactive Ctrl-C unloading modules, `DiscardQueue`, the log-starvation fix, the new test file and the CI change are all described now, and the stale point about raising `ShutdownRequested` is gone.
- **`continue-on-error` on the release leg** — resolved better than I asked: `tests.yml` is deleted outright. I checked nothing from master was lost — `git ls-tree` on the merge base lists only `python-cleanliness.yml`, `python-publish.yml` and `windows_build.yml`, so `tests.yml` was introduced by this PR and has now been withdrawn within it. The release leg is the pre-existing unconditional `tests` job (pinned `pymavlink==2.4.49`, no `continue-on-error`), so a release-only regression now fails the run.
- **The duplicated workflow** — the pymavlink-master leg is now `python-cleanliness.yml:43-61`, beside the existing job, exactly as suggested.
- **"The escape hatch can be deleted silently"** — fixed, and this is the one worth showing. The tests now assert *how many* records went to the busy log in the meantime (`tests/test_mavproxy_shutdown.py:519-526`, `BUSY_RECORDS_BEFORE_ROTATING = 10`), not just a wall-clock bound:

| mutation | at the old head | at this head |
|---|---|---|
| delete `mavproxy.py:1007-1011` (`if queue.empty(): return`) | PASSED | **FAILS** — `27 not less than 10 : 27 records written to raw first` (5/5 on the raw-log test) |
| `mavproxy.py:1046` `time.time() + 0.1` back to a 10 s deadline | — | **FAILS** — `0 is not true : cooked never written` |

So both halves of the fix are independently pinned. 12 consecutive runs of the file: 25 passed every time, 0 failures, 0 skips — 300 test executions, including 168 end-to-end signal-at-the-prompt runs against a real MAVProxy on a pty. I also went looking for a flake under load, on the theory that a contended runner could let the busy queue build a backlog that the escape hatch drains in one go: 64 spinners on a 32-core box, and again pinned to 2 CPUs shared with 6 spinners. No failures. The feeder thread is descheduled along with the writer, so the backlog does not grow.

### Worth fixing, not blocking

**`tests/test_mavproxy_shutdown.py:253-257` — the reworked `wait_until_blocked()` turns an environment problem into a silent skip, and the Windows-fallback tests are the ones that vanish.** `/proc/<pid>/wchan` reads back `'0'` both when the task is running and when wchan is unreadable (hidepid, a hardened or container kernel), and the new code `skipTest()`s at the deadline where the old code slept 1 s and ran the test anyway. Reproduced by pointing the path at an unreadable file:

```
3 skipped, 22 deselected in 31.17s
SKIPPED [1] tests/test_mavproxy_shutdown.py:270: main thread never seen blocked (0)
SKIPPED [1] tests/test_mavproxy_shutdown.py:276: main thread never seen blocked (0)
SKIPPED [1] tests/test_mavproxy_shutdown.py:282: main thread never seen blocked (0)
```

The suite reports green. What is silently lost is `test_sigterm_without_forwarder`, `test_sigint_without_forwarder` and `test_sigint_with_requireexit_without_forwarder` — the only coverage of the no-`pthread_kill` path, which is the Windows path the description itself flags as never having been run on Windows. It does not bite here (wchan is readable on this box, 0 skips in 12 runs), so this is robustness rather than a live failure. Cheap fix: skip only when the read actually raised `OSError`, and fail or fall back to the old bounded sleep when wchan was readable but never showed poll/select. Independently, `-rs` on both pytest invocations in `python-cleanliness.yml` (lines 41 and 61) would put any skip and its reason in the CI log — the deleted `tests.yml` had `-ra` for exactly that.

### A pre-existing hazard, offered as information

A cold second-opinion pass flagged `mavproxy.py:1832` as a deadlock: `quit_handler` calls `mpstate.status.stop_event.set()` unconditionally, and if a repeat fatal signal is delivered while the main thread is already inside `Event.set()`, the nested `set()` blocks forever on the Event's own non-reentrant lock. The mechanism is real — I reproduced it deterministically by forcing the interleaving, and hit it naturally under a maximal signal storm (~62k `set()` calls, 523 handler runs, hung in 3.1 s) — and the `SHUTDOWN_GRACE` force-exit did **not** rescue it in my harness.

**But it is not something this PR introduces.** Master's `quit_handler` also calls `stop_event.set()` from a signal handler, and under the same forced interleaving master deadlocks identically; I ran both. What the PR changes is the width of the window: master returns early once `_flag` is set, so it is exposed only between `_cond.acquire()` and `self._flag = True`, whereas the unconditional `set()` here is exposed for the whole body including `notify_all()`. A `set()` call is about 0.3 µs and happens once per shutdown, so at anything like human Ctrl-C rates this is not reachable — which matches the storm test I ran last round, 40 SIGINTs at 2 ms spacing, 40/40 handled cleanly on both heads.

So: not a blocker, and I am explicitly not asking you to fix a master bug in this PR. Mentioning it because the PR now owns this code and because restoring something like master's `if not stop_event.is_set():` guard, or setting the event outside the handler, would narrow it back for free.

Two smaller ones in the same spirit: `signal_state.handler_runs += 1` at `:1799`/`:1820` is the exact read-modify-write shape the new comment at `:1691-1694` says can lose an update in a handler — harmless here, since a lost update can only under-count and extra pokes are no-ops, but someone will ask. And `python-cleanliness.yml:43` has no `timeout-minutes`; the deleted `tests.yml` had 30, and the default is 360, for a suite whose subject is processes that fail to exit.

### Dropped

I could not show the counter→list refactor fixing anything reachable: 40 SIGINTs at 2 ms under `requireexit` gave 40/40 handled and a clean 21-module unload on **both** this head and the old counter version, and a direct nesting probe lost nothing in ~1400 handler entries. The reasoning is sound — `list.append` completes inside one C call where `x += 1` spans bytecodes — so I read it as free hardening and am not claiming a bug was fixed. Separately, the `exit` command not going through the new machinery (it sets `stop_event` from `main_loop` with nothing to wake the forwarder, so MAVProxy sits at the prompt until a keypress) behaves **identically on master**, so it is not yours and I am not asking for it here.

### Verdict: COMMENT

No blocking defect. The only thing I would actually like changed is the silent skip, because it can hide the Windows-path coverage this PR specifically cannot test any other way.

### What I checked and cleared

Termination ordering in `log_writer()` (`:1073-1101`): the `finally` that cancels both queues is unconditional, so the drain cannot be skipped by a write failure, a flush failure, or a `print()` to a closed pipe — and `test_an_unexpected_error_still_cancels` pins exactly that. Nothing joins the writer before the drain. No bounded-queue deadlock: both log queues are `multiprocessing.Queue` so `put()` never blocks, and a failed log's queue is swapped for a falsy `DiscardQueue` that the `if mpstate.logqueue:` guards at `:881` and `:948` then skip. `input_loop` catches only `(EOFError, IOError)`, so a `KeyboardInterrupt` or `ShutdownRequested` from `poke_handler` always reaches the handlers. `sys.excepthook` is replaced so a dying main thread still sets `stop_event`. No MAVLink constants are added at module scope, so there is no import-time `AttributeError` against released pymavlink — and the full suite (311 passed, 5 unrelated skips) was run under pymavlink 2.4.49, which is the leg CI pins. `flake8` clean via the repo's own `scripts/run_flake8.py`.

Not verified: Windows. The no-forwarder path is still only simulated on Linux by deleting `signal.pthread_kill`, and those are precisely the three tests the skip issue above can hide.

AP-Review:

Re-reviewed at head `a080c07163`. My earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_24_2306/devcall_pr_reviews.html#prMAVProxy-1722

**The one thing I asked to be changed is fixed, and I proved it by mutation rather than by reading it.** The other half of that suggestion — adding `-rs` — has a consequence I did not anticipate when I suggested it, and it is worth one more small change. That is the only open item, and it is mine, not yours.

### Resolved — the silent skip

`wait_until_blocked()` now skips only when the read actually raises, and falls back to the old bounded sleep when wchan is readable but never reports a wait state (`tests/test_mavproxy_shutdown.py:253-258`). Making `wchan_path` a class attribute (`:94`) is also what made this testable from the outside.

Proven by pointing the path at a readable file containing `0` — the "readable but never says poll/select" case — and running the three fallback tests on both heads:

| wchan is readable but never shows poll/select | `5d7ec185a6` (old) | `a080c07163` (this head) |
|---|---|---|
| `test_sigterm_without_forwarder`, `test_sigint_without_forwarder`, `test_sigint_with_requireexit_without_forwarder` | **3 skipped** — `main thread never seen blocked (0)` | **3 passed** (45.6 s: the 10 s deadline plus the 1 s grace, per test) |

And the genuinely-unreadable case still skips, now naming the errno: `cannot read <path>: [Errno 2] No such file or directory`. That is exactly the split I asked for, and the Windows-path coverage is no longer losable to an environment quirk.

Full suite at this head under pymavlink 2.4.49 (the version CI pins): **310 passed, 1 failed, 5 skipped**. The failure is `tests/test_raw_thermal.py::ThermalTests::test_display_and_save_preserve_native_samples`, and it is **not yours** — it fails identically at your merge base `2eb83952`, and it is an artefact of the numpy/opencv versions I happened to install locally. The 5 skips are the pre-existing GUI/vendor-video ones.

### Worth one small change — `-rs` hides which test failed

This one is my fault: I suggested `-rs` so skips and their reasons show up in CI. It does that, but `-r` **replaces** pytest's default report selection (`fE`) rather than adding to it, so on a failing run the short summary no longer names the failing test. Reproduced at this head on the failure above:

```
-rs      ->  (no "short test summary info" section at all)
             1 failed, 2 passed, 8 subtests passed in 0.14s

(default) ->  =========== short test summary info ===========
              FAILED tests/test_raw_thermal.py::ThermalTests::test_display_and_save_preserve_native_samples

-rfsE    ->  =========== short test summary info ===========
              FAILED tests/test_raw_thermal.py::ThermalTests::test_display_and_save_preserve_native_samples
              (plus the SKIPPED lines when there are skips)
```

So as it stands, a red run of a 300-test suite prints a traceback and a count but never the test id. `-rfsE` on both invocations (`python-cleanliness.yml:41` and `:61`) gives you the skip reasons I wanted *and* keeps the failure line. The `tests.yml` you deleted used `-ra`, which also works and is shorter.

### A note on the fallback, not a request

A cold second-opinion pass pointed out that the new grace path returns after `time.sleep(1)` for **any** readable state, not only an uninformative one — so on a box where the main thread parks in, say, `futex_wait_queue`, the helper returns without the synchronisation guarantee its docstring promises, and the caller signals into the race described at `:242-245`. The mechanism is real and follows directly from the code, but this is the option I offered you ("fall back to the old bounded sleep"), it is what the code did before this PR, and the alternative — failing — turns an environment quirk into a red CI. I measured no flakiness: the three fallback tests passed on every run here. If you ever want it tighter, the distinction to draw is "wchan told us nothing" versus "wchan told us it is waiting somewhere else".

### Still open from earlier rounds, all non-blocking and unchanged

- `python-cleanliness.yml:43` still has no `timeout-minutes`; the deleted `tests.yml` had 30 and the default is 360, for a suite about processes that fail to exit.
- `quit_handler`'s unconditional `stop_event.set()` (`mavproxy.py:1832`) and the `handler_runs += 1` read-modify-write (`:1799`, `:1820`) are unchanged. As I said last round, the first is a pre-existing master bug I am explicitly not asking you to fix here, and the second can only under-count harmlessly.

### Verdict: COMMENT

No blocking defect, and the substantive fix from last round is complete and now pinned by a test I could break on the old head and not on this one. The only thing I would still change is `-rs` → `-rfsE`, which is my error to correct.

PR #1724, opened 2026-08-07
Cope with new ListDirectoryWithTime FTP command, fix bugs

peterbarker:
Adds support for new MAVFTP ListDirectoryWithTime command.

Aborts listings when other operations are started.

Stop being clever about round-tripping listing state, have our own state (cross-contamination issues)


tridge:
already included in #1740

PR #1733, opened 2026-08-18
mavproxy_ftp: support concurrent, faster and resilient transfers

tridge:
## Summary

This reworks the MAVProxy FTP client to support multiple simultaneous operations and substantially improve transfer performance and resilience.

This replaces #1724 

- allow independent FTP workers so parameter FTP, directory listing, downloads, and uploads can run concurrently
- queue operations when all MAVFTP session IDs are in use, and quarantine recently retired IDs to avoid delayed-packet collisions
- pipeline reads and writes, negotiate accelerated upload capabilities, and batch outgoing MAVLink packets
- keep network writes below a safe datagram size while preserving efficient serial batching and reliable stream writes
- recover missing blocks efficiently under receive or transmit loss, with configurable loss and variable-lag simulation
- cancel delayed simulated traffic when a worker terminates
- support directory modification timestamps
- preserve idle FTP status output
- retain compatibility with older MAVFTP servers and reject accidental capability matches

## Why

The previous FTP module effectively serialized operations and used conservative request/response behavior. This prevented commands such as parameter FTP, `ftp list`, and `ftp get` from progressing together, limited throughput on fast links, and amplified stalls caused by loss or variable latency.

The new worker manager isolates transfer state per operation, bounds active sessions, and uses pipelining plus selective retry to keep the link busy without sacrificing compatibility.

## Compatibility

Accelerated uploads are enabled only after a server returns the `MFQ1` capability marker. Older servers continue to use the legacy single-request upload path. Cumulative acknowledgements remain compatible across old and new client/server combinations.

## Validation

- `pytest -q tests/test_mavproxy_ftp.py`: 22 passed
- ArduPilot SITL integration coverage for concurrent list/get/put operations
- SITL upload batching, duplicate-request, filename, timestamp, and poor-link regression tests
- transfer testing against Pixhawk6X over USB and STLink-V3
- final adversarial review of cancellation, session reuse, capability negotiation, batching, retry caching, and write-window behavior

### DIFF
--- a/MAVProxy/modules/mavproxy_ftp.py
+++ b/MAVProxy/modules/mavproxy_ftp.py
@@ -37,6 +37,9 @@ OP_TruncateFile = 12
 OP_Rename = 13
 OP_CalcFileCRC32 = 14
 OP_BurstReadFile = 15
+# Like OP_ListDirectory, with an mtime field appended to file entries.  Older
+# servers either NACK this request or ignore it, so callers must fall back.
+OP_ListDirectoryWithTime = 16
 OP_Ack = 128
 OP_Nack = 129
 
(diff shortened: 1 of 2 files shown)


tridge:
closing for #1740

PR #1736, opened 2026-08-22
decode_devid starts to understand encoding of serial ports into the MAVn_DEVID parameter

peterbarker:


tpwrules (changes requested):
How did you decide the port ranges? For instance we have default three `SCR_SDEV` ports and more can be defined. I wonder if this would be better implemented as ranges in case the user compiles in more ports. It seems we allow 1-9.

Also do we use the name USB0 in the firmware at all?

peterbarker:
> How did you decide the port ranges? For instance we have default three `SCR_SDEV` ports and more can be defined. I wonder if this would be better implemented as ranges in case the user compiles in more ports. It seems we allow 1-9.

I've changed this to be range-based.
 
> Also do we use the name USB0 in the firmware at all?

That was dumb, I've changed it to SERIAL0

peterbarker:
```
MAV1_DEVID: bus_type:SERIAL(6)  bus:0 address:0(0x0) devtype:0(0x0) SERIAL0 (6)
MAV2_DEVID: bus_type:SERIAL(6)  bus:0 address:0(0x0) devtype:1(0x1) SERIAL1 (65542)
MAV3_DEVID: bus_type:SERIAL(6)  bus:0 address:0(0x0) devtype:2(0x2) SERIAL2 (131078)
MAV4_DEVID: bus_type:SERIAL(6)  bus:0 address:0(0x0) devtype:5(0x5) SERIAL5 (327686)
MAV5_DEVID: bus_type:SERIAL(6)  bus:0 address:0(0x0) devtype:21(0x15) NET_P1 (1376262)
```

PR #1744, opened 2026-08-31
Plot arcs on maps

peterbarker:
Stop using straight-line segments where we should be using arcs.

Works for the new arc waypints, loiter commands, ....

### DIFF
--- a/MAVProxy/modules/lib/mp_module.py
+++ b/MAVProxy/modules/lib/mp_module.py
@@ -140,6 +140,16 @@ class MPModule(object):
     def get_mav_param(self, param_name, default=None):
         return self.mpstate.functions.get_mav_param(param_name, default)
 
+    def default_circle_radius(self):
+        '''the radius a circling mission item uses when it does not carry one
+        of its own.
+
+        This is the forward-flight loiter radius: vehicles without one (a
+        multicopter, say) sit at the point rather than circling it, so None
+        is the right answer for them and nothing should be drawn
+        '''
+        return self.get_mav_param('WP_LOITER_RAD')
+
     def param_set(self, name, value, retries=3):
         self.mpstate.functions.param_set(name, value, retries)
 
--- a/MAVProxy/modules/lib/mp_util.py
+++ b/MAVProxy/modules/lib/mp_util.py
@@ -339,6 +339,26 @@ def vehicle_cruise_speed(params):
     return _first_param(params, CRUISE_SPEED_PARAMS)
 
 
+def vehicle_track_convergence(params):
+    """the distance over which the vehicle pulls itself back onto the track
+    it is supposed to be following, or None if we cannot tell.
+
+    This is the L1 controller's own distance, which is what decides how
+    sharply it converges: 1/pi times the damping, the period and the speed
+    """
+    period = param_value(params, 'NAVL1_PERIOD')
+    damping = param_value(params, 'NAVL1_DAMPING')
+    speed = vehicle_cruise_speed(params)
+    if period is None or speed is None:
+        return None
+    if damping is None:
+        damping = 0.75
+    distance = 0.3183099 * float(damping) * float(period) * speed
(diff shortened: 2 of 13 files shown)


peterbarker:
image

image

tridge:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_04/devcall_pr_reviews.html#prMAVProxy-1744

Reviewed at head `9d55953fac`. **COMMENT.** The geometry is correct — I checked it independently and it holds — but the mapping from mission commands to what a vehicle actually flies has three gaps, one of which makes a new helper dead code. Worth saying that I had reached APPROVE on my own read and the cold cross-check moved it; the first item below is the one that changed my mind.

### `mission_circle_turns()` has no production caller

The helper reads `LOITER_TURNS` param1 and converts `DO_ORBIT` param4 from radians, and both of those are right — I checked them against `common.xml`, where DO_ORBIT param4 is indeed `units="rad"`. But its only references anywhere in the diff are four assertions in `tests/test_mission_geometry.py`; every production call site uses `mission_circle_radius()`. So a three-quarter orbit is drawn as a full circle, and the tests pass because they exercise the helper directly. Either wire it up or drop it.

### A zero radius is not "use the default" on Copter

`mission_circle_radius()` does `if radius == 0: radius = default_radius`, so a zero-radius `LOITER_TURNS` is plotted as a `WP_LOITER_RAD` circle — 42 m by default. ArduCopter treats that as a panorama, and is explicit about it in `ModeAuto::circle_movetoedge_start()`:

```cpp
// ... a zero (panorama) radius is stored explicitly because get_radius_m() falls
// back to the CIRCLE_RADIUS parameter when the commanded radius is zero
circle.radius_m = is_positive(radius_m) ? copter.circle_nav->get_radius_m() : 0.0f;
```

So the map draws a 42 m circle where the vehicle sits still and rotates. Note also that `DO_ORBIT` specifies *NaN*, not zero, as the request for a default radius.

### The tangent entry/exit model does not match Copter

The renderer joins the circle on a tangent chosen from the adjacent mission items and adjusts the sweep to leave on another. The same ArduCopter function says otherwise, in its own comment: *"Always fly to the exact edge of the circle first... that destination must land exactly on the circle at zero speed."* Copter flies to the closest point on the circle, stops, then orbits. The new test pins the renderer's model rather than the vehicle's, so it will not catch this.

### Lower confidence

The `LOITER_TO_ALT` turn estimate subtracts altitudes that may be in different frames (terrain altitudes are converted to AMSL while relative ones stay relative), and the hovering-vehicle classification omits Rover — which also holds position for `LOITER_UNLIM`/`_TIME` — and `MAV_TYPE_DECAROTOR`. I did not verify these as carefully as the three above. The cross-check also notes that wrapping `gps_distance`/`gps_bearing` affects three callers that measure raw `minlon..maxlon` bounding boxes, so a log genuinely spanning more than 180° of longitude would get a 35.6 km camera distance instead of 40,052 km — real, but it needs a single log spanning half the globe.

### Checked and clear

The arc geometry is right. I worked the centre-bearing formula through by hand in both directions: a 180° clockwise arc gives the chord midpoint at `chord/2`, and 90° gives `chord_bearing + 45` at `chord/√2`, both correct, and the counter-clockwise branch mirrors properly. `radius = chord / (2·sin(θ/2))` is the standard chord relation, the degenerate cases are handled and documented, and the step count is sane. Every parameter index in `mission_circle_radius()` checks out against `common.xml` — radius is param3 for `LOITER_UNLIM`/`_TURNS`/`_TIME`, param2 for `_TO_ALT`, param1 for `DO_ORBIT` — as does the sign convention for direction. The antimeridian fix is algebraically equivalent to the old form whenever `|Δlon| ≤ 180` and corrects it otherwise.

The cold pass returned REQUEST CHANGES; I did not follow it there. These are rendering-fidelity gaps in a visualisation tool rather than correctness failures, and the geometry underneath them is sound.

PR #1747, opened 2026-09-10
ftp: show a directory's modification time in a listing

peterbarker:
A listing asked for with times gives a directory the same trailing fields as a file, "D<name>\t<size>\t<mtime>", but we printed everything after the D as the name, so a directory came out as "subdir\t0\t1788928014".

Take the name from the front and the time from the end, as the file case already does. Servers differ on whether a directory carries the fields at all, so an entry without them still prints as the bare name it is.


peterbarker:
Tested on ArduPilot supporting with-time and without

tridge (approved):

PR #1749, opened 2026-09-12
paramedit: fix forkserver startup and descriptor leaks

tridge:
Starting the GUI with a bound method serializes all of MPState, including closed console pipes. On Python 3.14 this fails repeatedly from idle_task until file descriptors are exhausted.

Pass only GUI data to the child, unload after startup failure, and clean up partially initialized resources and worker threads. Drain pending GUI output before joining the queue feeder on shutdown.


tridge:

Reviewed at head `baba7767bd`.

Full report (with the evidence behind each point): https://uav.tridgell.net/DevCallReviews/2026_09_12_AIReview/devcall_pr_reviews.html#prMAVProxy-1749

**Verdict: REQUEST CHANGES** — the diagnosis and the core of the fix are right, but the rewritten `close()` adds two paths that can block MAVProxy's exit indefinitely, and one of them is a regression against master on the Python this project's CI builds with.

**CI at this head:** 1 passing / 0 failing / 1 pending.

## Actionable

- **BUG** | VERIFIED | `MAVProxy/modules/mavproxy_paramedit/param_editor.py:250-254` | **`terminate()` followed by an unconditional `join()` hangs forever under the `fork` start method** — the Linux default on every Python before 3.14, including the 3.10 this project's CI builds against. `mavproxy.py:1480-1486` installs `quit_handler` for SIGTERM, and on its *first* delivery that handler only sets `mpstate.status.stop_event`; it does not exit. A forked GUI child inherits that Python-level handler (forkserver and spawn children do not, which is why the new tests pass on 3.14), and the wx GUI never consults `stop_event` — so `self.child.terminate()` at :253 is absorbed and `self.child.join()` at :254 never returns. Reproduced directly: fork child + MAVProxy's actual `quit_handler`, `terminate()` left it alive and the join was still blocked after 5 s; only `kill()` ended it. Master's `close()` called `terminate()` and returned with no join, so this is new here. Fallout: an interactive `module unload paramedit` leaks a permanently stuck thread (`mavproxy.py:415-425` joins with `timeout=5`, prints "unload on module paramedit did not complete" and carries on), and the exit path at `mavproxy.py:1668-1671` calls `m.unload()` directly with no timeout, so MAVProxy hangs on exit. Suggested fix: bound the post-terminate join and fall back to `self.child.kill()`.

- **BUG** | VERIFIED | `param_editor.py:261-262` | **The shutdown drain does an unbounded blocking `get()` on a queue whose other reader may have just been SIGTERMed mid-read**, which deadlocks the same two call sites. `get()` with no timeout takes CPython's `with self._rlock: res = self._recv_bytes()` path (`multiprocessing/queues.py:97-101`), and `_rlock` is shared with the child. The child reads the same queue from a wx timer (`param_editor_frame.py:250-253`, `gui_event_queue.get(block=False)`), which also holds `_rlock` across `_recv_bytes()`; if the `terminate()` at :253 lands in that window the child dies holding the lock and the parent blocks forever. `wx.CallAfter(ExitMainLoop)` schedules work on the GUI thread, it does not interrupt a timer callback already inside a queue read, so the window is reachable. Both Codex passes reproduced this independently, under spawn and under forkserver, by suspending the child inside `_recv_bytes()`. Worth noting neither obvious one-liner is sufficient alone: `get(timeout=...)` bounds the lock acquire and the poll but not the `_recv_bytes()` after it, and `cancel_join_thread()` avoids the blocking join but leaves a full-pipe feeder thread alive (also reproduced), which matters for unload/reload inside one session. `cancel_join_thread()` on both queues **plus** a bounded drain covers it.

- **ISSUE** | VERIFIED (mechanism) / UNCONFIRMED (on native Windows) | `param_editor.py:240-269` | **Only `gui_event_queue` is drained, but on Windows `event_queue` needs it too.** There the "child" is a `threading.Thread` in the same process, so both queues' feeders belong to the parent — and `close()` stops `event_queue`'s only reader first (`event_thread`, joined at :240-242) and then joins the GUI thread at :246-248. Anything the GUI queues in that window (e.g. a `PEE_WRITE_PARAM` carrying a large modified-parameter dict) sits in the feeder with nobody reading, so `event_queue.join_thread()` at :269 blocks forever. Reproduced with real threads and real queues, stopping the reader before the puts as `close()` does: `join_thread()` still blocked after 10 s. Note my *first* reproduction attempt did **not** hang — with the reader still draining concurrently, `close()` breaks the pipe and the feeder exits — so it is the ordering that makes it bite.

- **ISSUE** | VERIFIED | `tests/test_paramedit.py:131-141` | `test_process_start_methods_and_shutdown_with_full_gui_queue` installs its wx stand-ins through `process_factory`, which `start()` only calls on the non-Windows branch — on Windows `child_class` is `threading.Thread` and the factory is never used, so the test would drive the real wx entry point. The thread backend is the branch with the `event_queue` problem above and has no coverage at all. Separately, the stand-in child never reads `gui_event_queue`, which is why the suite passes despite both deadlocks.

<details><summary>Lower-severity notes (2)</summary>

- **NOTE** | VERIFIED | `__init__.py:40-44` | The comment says "Report the original failure once, then unload", but nothing reports it at default settings: `mavproxy.py:1110-1116` catches the re-raised exception and prints only when `moddebug >= 1`, and moddebug defaults to 0 (`mavproxy.py:264`). The user does see "Unloading module paramedit", so it is not silent — but the reason never reaches them, which is the one thing worth knowing when a GUI refuses to open. Not a regression; an explicit `print()` would deliver what the comment promises.
- **NOTE** | VERIFIED | `param_editor.py:195-196` | `wx_util.safe = True` by hand duplicates `MAVProxy/modules/lib/wx_processguard.py`, whose whole job is that assignment and which ~10 other GUI entry points import for exactly this (`mp_menu.py`, `wxconsole.py`, `mission_editor.py`, `live_graph.py`, `mavvidplay.py`, `wxsaildash.py`).

</details>

## What was checked and is clean

I mutation-tested the fix rather than reading it — reverting one piece at a time and re-running `tests/test_paramedit.py` on 3.14.4. Restoring the bound target fails the spawn and forkserver subtests with `OSError` from `connection.py` (the reported bug); removing the `idle_task` guard fails `test_failed_startup_is_not_retried`; removing the `close()`-on-failed-start fails `test_partial_queue_allocation_is_cleaned_up`; removing the drain makes the suite **hang** until killed at 300 s. Every part of the change is load-bearing and every test is a real regression test.

Also cleared: `dict(mav_param)` is sufficient (traced every use through `set_param_init` → `redraw_grid` and `param_received` at `param_editor_frame.py:273-290`, `:500`, `:512`, `:523`, `:545` — only `.items()`/`.keys()`/`.clear()`/`__setitem__`, no MAVParmDict-specific method, and the full dict still arrives in the `PEGE_READ_PARAM` event); clearing `mpstate.param_editor` to `None` is not a new `AttributeError` source (nothing outside the package reads it); `ParamEditorMain.idle_task()` cannot run after `close()` sets `self.child = None`, because of the new `needs_unloading` guard; `self.child.close()` is correctly confined to the non-Windows branch and is legal on a never-started `Process`; and moving the parent-only `mavlink_message_queue` to `queue.Queue` is right and saves two descriptors.

_Both an independent Codex pass and this review were run on this head and reconciled; findings that one pass refuted are not listed above._

PR #1752, opened 2026-09-12
camera: add live settings from MAVLink camera definitions

tridge:
Load camera definition XML from CAMERA_INFORMATION and build tabbed controls with binary PARAM_EXT reads, acknowledged writes, and dependency rules. Support HTTP, MAVFTP and XZ definitions while keeping downloads and dialog updates off the MAVLink processing path.

Prefer matching camera components over autopilot proxies and report settings absent from a complete camera parameter list. Add parser, protocol, targeting and FTP regression tests.

### DIFF
--- a/MAVProxy/modules/mavproxy_camera/README.md
+++ b/MAVProxy/modules/mavproxy_camera/README.md
@@ -40,6 +40,43 @@ keeps narrower streams nested inside wider streams with the same boresight.
 `fov_max_range` rejects unreliable terrain intersections beyond 10 km by
 default; set it to zero to disable this range limit.
 
+## Custom camera settings
+
+Choose **Camera → Custom Settings** in the console or map menu, or run
+`camera custom`. The dialog uses the selected camera; use
+`camera select SYSID:COMPID` when more than one camera is discovered.
+Automatic selection prefers a matching camera component over ArduPilot's
+duplicate camera information from component 1. The autopilot proxy does not
+serve the camera's extended parameters. Explicit selections are preserved.
+
+`CAMERA_INFORMATION.cam_definition_uri` supplies the camera's XML definition.
+HTTP, HTTPS and MAVFTP downloads and XZ-compressed definitions are supported.
+Load `module load ftp` for MAVFTP URLs. Downloads run without blocking camera
+telemetry. A changed URI or definition version reloads the metadata and values.
+Use `camera definition` to retry a download, or `camera definition FILE|URL` to
+load an explicit definition, such as a vendor-supplied local XML file.
+
+The dialog preserves XML order across tabs, with checkboxes for booleans,
+dropdowns for named options, sliders for bounded stepped values, and numeric
+fields for other settings. Checkboxes, dropdowns and released sliders apply
+immediately; numeric fields apply on Enter or their Apply button. Values are
+validated against the type, bounds, steps and currently allowed options.
+Read-only settings cannot be changed, and settings that have not replied to
+parameter requests remain disabled. XML defaults are metadata, not assumed
+current camera values. Vendor-specific `custom` binary types have no editor.
+
+Changes use binary `PARAM_EXT_SET` values and wait for `PARAM_EXT_ACK`.
+Pending writes, rejections and timeouts are shown in the dialog. Exclusion
+rules, conditional option ranges and dependent parameter refreshes follow the
+camera definition, including self-refreshing actions such as Workswell's NUC.
+Incoming camera values update open dialogs, and periodic reads pick up changes
+made by another controller. **Refresh** requests the full parameter list with
+`PARAM_EXT_REQUEST_LIST`, then retries missing parameters individually. Closing a
+dialog does not undo changes already applied to the camera.
+
+`camera params` lists the received values, and `camera param NAME VALUE`
+provides the same validated, acknowledged writes from the command line.
+
 ## ArduPilot and MT11 configuration
 
 For the MT11, use one `CAM1_TYPE=6` MAVLink Camera v2 backend and one
--- a/MAVProxy/modules/mavproxy_camera/__init__.py
+++ b/MAVProxy/modules/mavproxy_camera/__init__.py
@@ -9,6 +9,7 @@ from MAVProxy.modules.lib import mp_module
 from MAVProxy.modules.lib import mp_settings
 from MAVProxy.modules.lib import mp_util
 from MAVProxy.modules.lib import camera_projection
+from MAVProxy.modules.mavproxy_camera.parameters import CameraParameters
 from pymavlink import mavutil
 from pymavlink.quaternion import Quaternion
 
@@ -81,6 +82,7 @@ class CameraDevice:
         self.storage = {}
         self.streams = {}
         self.definition = None
+        self.parameters = None
 
     def label(self):
         if self.information is None:
@@ -135,7 +137,7 @@ class CameraModule(mp_module.MPModule):
         ])
         self.add_command(
             "camera", self.cmd_camera, "MAVLink camera control",
-            ["<status|discover|select|info|streams|view|projection|photo|stopphotos|record|zoom|focus|mode|source|stream|mount|set>",
+            ["<status|discover|select|info|custom|definition|params|param|streams|view|projection|photo|stopphotos|record|zoom|focus|mode|source|stream|mount|set>",
              "set (CAMERASETTING)"])
         self.add_completion_function("(CAMERASETTING)",
                                      self.camera_settings.completion)
@@ -143,6 +145,7 @@ class CameraModule(mp_module.MPModule):
         self.gimbals = {}
         self.manager_attitudes = {}
         self.selected_camera = None
+        self.camera_selection_explicit = False
         self.selected_gimbal = None
         self.views = {}
         self.last_status_request = 0.0
@@ -159,6 +162,7 @@ class CameraModule(mp_module.MPModule):
             self.menu = MPMenuSubMenu("Camera", items=[
                 MPMenuItem("Status", returnkey="# camera status"),
                 MPMenuItem("Discover", returnkey="# camera discover"),
+                MPMenuItem("Custom Settings", returnkey="# camera custom"),
                 MPMenuItem("Take photo", returnkey="# camera photo"),
                 MPMenuItem("Toggle recording",
                            returnkey="# camera record toggle"),
@@ -176,6 +180,8 @@ class CameraModule(mp_module.MPModule):
         for view in self.views.values():
             view.close()
         self.views.clear()
+        for camera in self.cameras.values():
+            camera.parameters.close()
         if self.menu is not None:
             for name in self.menu_modules:
                 module = self.module(name)
@@ -189,6 +195,10 @@ class CameraModule(mp_module.MPModule):
   camera discover                     request fresh discovery information
   camera select [SYSID:]COMPID         select a camera
   camera info                          show selected camera information
+  camera custom                       open live Custom Settings dialog
+  camera definition [FILE|URL]         reload or override camera definition
+  camera params                       show custom parameter values
+  camera param NAME VALUE             set a custom camera parameter
   camera streams                       show discovered video streams
   camera view <ID|rgb|thermal|all>     open RTSP viewer(s)
   camera projection                    show/refresh map projection status
@@ -219,6 +229,7 @@ class CameraModule(mp_module.MPModule):
         camera = self.cameras.get(key)
         if camera is None:
             camera = CameraDevice(*key)
+            camera.parameters = CameraParameters(self, camera)
             self.cameras[key] = camera
         camera.last_seen = time.time()
         if self.selected_camera is None:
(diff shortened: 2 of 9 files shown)


tridge:

<details><summary>Previous review (2026-09-12)</summary>

Reviewed at head `fe4f3c46d5`.

Full report (with the evidence behind each point): https://uav.tridgell.net/DevCallReviews/2026_09_12_AIReview/devcall_pr_reviews.html#prMAVProxy-1752

**Verdict: REQUEST CHANGES** — on the first item below. Nothing here is architectural; the design underneath is sound and all four bugs have cheap fixes.

The architecture is right and the risky parts are mostly handled well: the dialog is a real child process following the established MAVProxy pattern, all network and XML work is off the packet loop, async results cross back through a generation-guarded queue, and the wire-type mapping is correct. The findings are concentrated in validation of camera-supplied data.

### BUG — a camera can permanently wedge MAVFTP with five malformed advertisements

`MAVProxy/modules/mavproxy_camera/parameters.py:116`. Verified end-to-end:

1. A camera advertises `mavftp:///caf%C3%A9.xml`.
2. `parameters.py:116` calls `ftp.cmd_get([unquote(path)], ...)`; `unquote()` yields the non-ASCII `/café.xml`.
3. `_launch()` (`mavproxy_ftp.py:1623-1637`) does `self.workers[session] = worker` and *then* calls the method.
4. `FTPWorker.cmd_get` runs `enc_fname = bytearray(fname, 'ascii')` (`mavproxy_ftp.py:481`) → `UnicodeEncodeError`.
5. The exception propagates out of `_launch()`, so the `if worker.last_op is None: self.worker_done(worker)` cleanup on the very next line never runs. The slot is never reclaimed.

`max_sessions` defaults to 5 (`mavproxy_ftp.py:1420`), so five such advertisements exhaust every slot and **all subsequent MAVFTP work — parameter download included — queues forever with nothing on the wire**. I reproduced the encode failure and then the leak semantics against the exact `_launch()` control flow: five iterations leave 5 of 5 slots occupied.

The missing `try`/`except` in `_launch()` is pre-existing, but this PR adds the first caller whose filename is chosen by a remote device, and auto-triggers it from `CAMERA_INFORMATION` with no user action. Worth fixing at both ends — validate/encode the path before submitting, and make `_launch()` reclaim the slot on exception.

### BUG — the step check is wrong in both directions

`definition.py:125-128`. One root cause: `steps = (value - self.minimum) / self.step` compared with `math.isclose(..., rel_tol=1e-5, abs_tol=1e-5)`, so the tolerance scales with the step *count*.

- **Accepts invalid values.** With `min=0, step=10`: `validate(999999)` and `validate(100009)` both pass despite `value % 10 == 9`, while 10009, 1009, 109 and 19 are correctly rejected. The threshold is exactly `value >= 100000`, where `rel_tol * steps >= 0.1`.
- **Rejects valid ones.** `convert()` quantises to float32 before the check, so for `<parameter type="float" min="-100" max="100" step="0.01"/>`, `validate('-99.99')` raises — 20 of 20,001 legal slider positions refused, and 8 of 12,732 for a very plausible `min="-273.15" max="1000" step="0.1"` thermal range.

Exact integer arithmetic for integer types, and a ULP-derived tolerance computed against the parsed (double) min/step for floats.

### BUG — `equal_value()` conflates distinct float32 options

`definition.py:41-44`. `rel_tol=1e-6` makes the **8 nearest distinct float32 values above 1.0** compare equal to 1.0, starting with the immediately adjacent `1.0000001192092896` (measured). Options are already canonicalised to their exact wire value by `convert()` (`definition.py:113`), so their identity should be exact equality — the tolerance buys nothing and costs correctness. The dropdown highlights the wrong option when the camera reports a near neighbour, and `validate()` accepts an option the definition excludes. Same comparison backs exclusions and range conditions.

### BUG — one stray attribute on one `custom` parameter takes down the whole definition

`definition.py:79-81` and `:86-87`. `TYPES['custom'] = (11, None)`, and `default` is correctly guarded with `if self.fmt else None` at `:78` — but `min`, `max`, `step` and the option values are not, so they call `convert()`, which does `'<' + self.fmt` with `fmt is None`. Reproduced: `<parameter name="D" type="custom" min="1"/>` raises `ValueError: invalid camera parameter D: can only concatenate str (not "NoneType") to str`, `CameraDefinition.__init__` re-raises, and **every setting on that camera becomes unavailable**, with a raw Python `TypeError` string shown to the user. A well-formed parameter alongside it does not survive either. Same `if self.fmt` guard fixes it. I found no real vendor file that does this, so likelihood is unproven — but the blast radius is total.

### ISSUE — the deselected proxy component duplicates the download and all PARAM_EXT traffic

`mavproxy_camera/__init__.py:977-978`. `information()` runs for every camera *before* `_prefer_camera_component()`, and `parameters.py:317` runs `request_all()` for every camera in `idle()` regardless of selection. Measured with the repo's own test fakes, feeding CAMERA_INFORMATION from comp 1 and comp 100 sharing one `mftp://[;comp=100]/camera.xml`: 2 `cmd_get` calls (same file fetched twice) and `param_ext_request_read_send` issued 12 times to *each* of (1,1) and (1,100). For a 60-parameter definition that is one wasted full MAVFTP transfer plus ~120 wasted PARAM_EXT messages, aimed at a component the README says does not serve extended parameters.

### ISSUE — declared pymavlink floor is 24 releases too low

`setup.py:29` says `pymavlink>=2.4.14`, but `decode_value()` (`definition.py:24`) reads `message._param_value_raw`, which the generator only started emitting in pymavlink `667cfcd4` (2022-11-06) — first released in **2.4.38** (`git tag --contains`). On 2.4.14–2.4.37 the `getattr` falls back to the decoded `str` and every read fails with "extended parameter has no raw binary value": degraded rather than crashing, but every setting is unavailable and the message gives no hint why. Released pymavlink is 2.4.49 so most users are fine; the floor is still wrong.

### ISSUE — XML expansion is not bounded by `MAX_DEFINITION_SIZE`

`definition.py:9`, `:138`. Stdlib `ElementTree`. On this box (expat 2.7.4) the runtime's amplification guard does most of the work — I confirmed nested billion-laughs is rejected and an external-entity XXE is rejected as an undefined entity. But single-level internal entity expansion is allowed to roughly 100x: I expanded 50,467 bytes into a 5,000,000-character string, and 1,049,013 bytes into 94.4 MB of text at 288 MB RSS. Scaled to the 4 MB input cap that is order 360 MB of text / ~1 GB RSS, forced by a camera on the link. **Unconfirmed:** exposure on older expat — MAVProxy CI runs ubuntu-22.04, whose expat is older than this box's, and I had no old expat to test against. `defusedxml`, or an `XMLParser` with entity handling disabled, closes both.

### ISSUE — the MAVFTP size cap runs after the file is already buffered

`parameters.py:110-114`. `ftp_done()` reads `MAX_DEFINITION_SIZE + 1` from the handle, but the FTP module has already accumulated the complete transfer in an in-memory `SIO()` before the callback fires, so an advertised multi-megabyte file is bounded only by the transfer. Largely a pre-existing MAVFTP property — but this PR is what lets a camera choose the file.

### Notes

- `settings_dialog.py:177-182` — tooltips and the step error print the float32-quantised bounds, so `min="0.1" step="0.1"` shows `Minimum: 0.10000000149011612`. Keeping the parsed double alongside fixes this and the float half of the step bug together.
- SSRF surface is narrow and mostly handled (`http`/`https` only, TLS verification on, `timeout=15`, bounded read, off the packet loop). Residual: the fetch is automatic on `CAMERA_INFORMATION` with no user action, nothing stops a camera pointing at `http://127.0.0.1:...`, and urllib's redirect handler permits `ftp` as a redirect target. Response is never reflected, so practical risk is low — flagging so the trust boundary is a deliberate choice.
- None of the four bugs above is covered by the 340 new test lines. Also untested: `download_definition` and the whole HTTP path, the scheme rejection at `parameters.py:119-121`, `definition_bytes`' size/XZ limits, the `local` file path, `decode_value`'s non-finite rejection, the `custom` decode branch, and all 280 lines of `settings_dialog.py` — including `_DialogChannel` and `CameraSettingsDialog.send/events/close`, which are pure logic and testable without a display.

### What was checked and came back clean

- **Module-scope MAVLink constants — clean.** Every `mavutil.mavlink.*` reference is inside a method (`parameters.py:231`, `:246`, `:251` only); `definition.py` and `settings_dialog.py` reference pymavlink not at all. All constants used verified present in released pymavlink 2.4.49 by both passes.
- **The shared MAVFTP change is backward-compatible.** `_submit()` gains keyword-only `target_system`/`target_component` defaulting to `None` with an `is None` fallback, so an explicit `0` is honoured. All three existing `cmd_get` callers — `mavproxy_param.py:332`, `mavproxy_oldwp.py:914`, `mission_item_protocol.py:1022` — pass neither kwarg, so parameter and mission download are byte-identical; the log module uses `LOG_REQUEST_DATA`.
- **PARAM_EXT type handling is correct.** Both passes independently packed and MAVLink-round-tripped all ten numeric types at their extremes — all exact. `TYPES` wire numbers 1..11 match `MAV_PARAM_EXT_TYPE_*`, `encode()` gives exactly 128 zero-padded LE bytes, and the `_param_value_raw` approach is genuinely necessary.
- **wx threading and headless.** Child process via `multiproc.Process` + `Pipe` + `Event`, `wx_processguard`, `child_close_fds()`, all wx classes constructed in the child. No wx call on the MAVLink thread. Module scope imports only `queue` and `multiproc`, and `parameters.py:285-288` checks `has_wxpython` and prints a CLI fallback before importing the dialog.
- **The packet path cannot raise** for any input I could construct.
- Tests as they stand: 63 passed for the three new files, 83 passed / 9 skipped for the whole tree, and `scripts/run_flake8.py MAVProxy` exits 0.

**CI at this head:** 1 passing / 0 failing / 1 pending.

_Both an independent Codex pass and this review were run on this head and reconciled; every finding above was re-reproduced here before inclusion._


</details>

tridge:
**Automated review note — AI-generated (Claude), validated against the live diff (Claude + Codex cross-checked).** Please sanity-check before acting.

Re-reviewed at head `5769cf8de0` (previously `fe4f3c46d5`); my earlier comment above is superseded. Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_14_1047/devcall_pr_reviews.html#prMAVProxy-1752

**All 8 findings from the last round (4 BUGs, 4 ISSUEs) are fixed, and each fix was proven by reverting it and watching a test fail.** The verdict moves REQUEST CHANGES → **COMMENT**. One new issue: the proxy de-duplication fix stops some explicit reloads from working.

CI at this head: **2 passing**, none failing.

### Previous findings

| Previous finding | Status |
|---|---|
| **BUG** — malformed MAVFTP names use up session slots | **RESOLVED** — `MAVProxy/modules/mavproxy_camera/parameters.py:98-100` checks the name with the new `encode_filename()` (`MAVProxy/modules/mavproxy_ftp.py:79`), and `_launch()` at `mavproxy_ftp.py:1660-1673` always frees the slot. Mutations fail 4 tests. |
| **BUG** — step check wrong both ways | **RESOLVED** — `MAVProxy/modules/mavproxy_camera/definition.py:118`, `:158-179`: exact integer arithmetic, ULP-based float tolerance, double-precision bounds. 1,774,731 valid float positions over 16 ranges: 0 wrongly rejected, and no off-grid value accepted. The old head's slider rejected 20 and 77 positions on two sample ranges; this head rejects none. |
| **BUG** — `equal_value()` merges adjacent float32 options | **RESOLVED** — `definition.py:45-48` is exact equality; the mutation fails the test. |
| **BUG** — stray attribute on a `custom` parameter kills the definition | **RESOLVED** — early return at `definition.py:109-110`; the mutation fails the test. |
| **ISSUE** — deselected autopilot proxy duplicates the download and PARAM_EXT traffic | **RESOLVED** — gate at `MAVProxy/modules/mavproxy_camera/__init__.py:985-990`; the mutation fails 3 tests. It introduced the regression below. |
| **ISSUE** — pymavlink floor too low | **RESOLVED** — `setup.py:29` requires `pymavlink>=2.4.38`. |
| **ISSUE** — XML entity expansion | **RESOLVED** — defusedxml with `forbid_dtd=True` at `definition.py:189`. Billion-laughs is rejected; swapping back to stdlib ElementTree fails 6 test cases. |
| **ISSUE** — MAVFTP size cap applied after buffering | **RESOLVED** — checked on the advertised size (`mavproxy_ftp.py:512`), each burst packet (`:608`) and gap reads (`:697`). HTTP reads stop at the limit + 1 byte, chunked included. The gap-read check itself is untested (NOTE below). |
| NOTE — tooltip precision | **RESOLVED** |
| NOTE — SSRF / redirects | **RESOLVED** — non-http(s) redirects are rejected (`definition.py:71-75`), and the README now documents the trust boundary. |
| NOTE — test gaps | **MOSTLY RESOLVED** — HTTP, local file, XZ limits, `decode_value` and the dialog channel are now tested. wx frame interaction is still untested. |

### New finding

**ISSUE — the proxy gate blocks explicit reloads when the autopilot proxy was auto-selected.** `MAVProxy/modules/mavproxy_camera/__init__.py:985`, with `:424-439`. The gate only lets component-1 CAMERA_INFORMATION through when component 1 is explicitly selected or is the `camera_component` override. Three cases, reproduced with the repo's test fakes at both heads by Claude and Codex independently:

| Scenario (component 1 auto-selected) | fe4f3c46 | 5769cf8de0 |
|---|---|---|
| `camera definition` retry, then CAMERA_INFORMATION re-sent | download starts | 0 downloads, status stays "Waiting for camera information" |
| after `camera custom`, the camera advertises a new URI/version | reloads | keeps the old definition, contradicting `MAVProxy/modules/mavproxy_camera/README.md:62` |
| CAMERA_SETTINGS, then `camera custom`, then CAMERA_INFORMATION | download starts | 0 downloads (`open_when_ready` is ignored) |

Explicit `camera select 1:1` still works, so there is a workaround. Suggested fix: let the gate also pass when `parameters.identity is not None or parameters.open_when_ready`, and handle `camera definition` with no arguments like the other subcommands. `identity is not None` alone does not cover the third case.

### Notes (non-blocking)

- **NOTE — the defusedxml import can take out the whole camera module.** `MAVProxy/modules/mavproxy_camera/definition.py:12-13` imports it at module level, and the camera module imports it on load. Without defusedxml, photo, video, zoom and gimbal control are lost too, not just the settings feature (simulated). `setup.py:30` declares the dependency, so pip installs are fine; git checkouts and existing environments are not. A lazy import inside `CameraDefinition.__init__` with an "install defusedxml" error would avoid it.
- **NOTE — double parameters can reject the top slider position.** For `min=0.1 max=25.5 step=0.1` the slider's last position is `25.500000000000004` and is rejected as above maximum (`MAVProxy/modules/mavproxy_camera/settings_dialog.py:152`, `definition.py:160`). Float types are unaffected, and this predates the latest commit. Snapping the top position to `maximum` would fix it.
- **NOTE — `timeout=15` doesn't bound total HTTP time** (`definition.py:82`). It is per socket read: a 1 byte/s server kept a 25-byte download going for 24 s. This runs on a worker thread and predates the latest commit.
- **NOTE — `forbid_dtd=True` also rejects a harmless `<!DOCTYPE mavlinkcamera>`** (`definition.py:189`). The README says this is intended. Whether any vendor definition file carries a DOCTYPE is unconfirmed.
- **NOTE — the gap-read size check at `MAVProxy/modules/mavproxy_ftp.py:697` has no test.** Removing it leaves all tests passing. It is defensive, since gap offsets come from already-checked burst offsets.

**Also checked, no finding:**
- No MAVLink constants at module scope (checked against released pymavlink 2.4.49).
- vermin reports Python 3.5+ for the module code.
- Malformed definitions can't raise inside the packet handler.
- The new FTP `max_size` argument defaults to off, so other callers are unchanged.

PR #1753, opened 2026-09-14
Tests and fixes for plotting of arcs on maps

peterbarker:
# pr/plot-arcs-better

Follow-up fixes to the arc and loiter mission drawing already on
`master` ("map: plot new arc waypoints better" through "map3d: show which
way the mission is flown"). They were found by review and by flying
test missions in SITL, then comparing the maps against the dataflash logs.

## Fixes

| Commit | What was wrong |
|---|---|
| map3d: resolve altitude frames before measuring a loiter climb | The live 3D map subtracted raw altitudes across frames, so an AMSL item before a home-relative `LOITER_TO_ALT` read as a huge dive and drew the wrong number of turns. |
| mp_util: only read param4 as a crosstrack choice where it is one | The pull-back onto the track out of a loiter centre was drawn for any circling item with `param4 <= 0`. ArduPilot only reads param4 that way for `LOITER_TURNS`, `LOITER_TIME` and `LOITER_TO_ALT`. |
| map3d: draw the turns LOITER_TURNS and DO_ORBIT ask for | `mission_circle_turns()` was never called, so a three-turn loiter that climbs while circling was drawn as one turn. |
| mp_util: take a LOITER_TIME radius of +-1 as a direction | ArduPilot can't store a `LOITER_TIME` radius and returns ±1 for the direction, which drew a 1 m circle. It now draws `WP_LOITER_RAD`, which is what the vehicle flies. |
| MAVExplorer: draw only the last mission a log holds | A log dumps the mission at boot and on every upload, and the 3D map drew all of them joined end to end. Items are now keyed by sequence, as mavflightview does. |
| MAVExplorer: draw a takeoff which has no position of its own | Positionless takeoffs were dropped from logs. They now climb from home, or from where the flight started if home was empty when the mission was uploaded. |

## Behaviour change

**map: draw loiter circles and direction arrows by default.** The 2D map's
`loitercircle` and `showdirection` are now on, as is `showdirection` in the
map3d module and MAVExplorer, so every view of a mission looks the same.

## Test missions

`tests/missions/plane-mission-geometry.txt` and
`copter-mission-geometry.txt` are SITL-flyable missions that exercise the
drawing: climbing and descending spirals, crossing and non-crossing
tangents between loiters, both param4 settings, mixed altitude frames,
arc waypoints (copter only, since ArduPlane has no `NAV_ARC_WAYPOINT`),
and the loiters a multirotor holds position for. `tests/missions/README.md`
describes each one.

## Verified against SITL

- **Copter** — drawn line against the flown track: median 0.2 m on the
  ground and 0.0 m in altitude. The −270° arc bulges 302 m off its chord,
  and the flight stays within 0.9 m of the drawn curve.
- **Plane** — median 0 m in altitude, mean 19 m on the ground. The larger
  horizontal figure is mostly ArduPlane flying loiters 15–20% wider than
  commanded.

`tests/test_mission_geometry.py` goes from 67 to 78 tests. flake8 and the
tests pass at every commit.

## Not addressed

- A level circle is drawn as one lap however many it asks for, since extra
  laps would sit on top of the first.
- `LOITER_TIME`'s dwell isn't turned into turns.
- Spiral turn counts come from `TECS_CLMB_MAX` / `TECS_SINK_MIN`. In SITL
  the plane climbed a little slower than that and descended faster.


tridge:

<details><summary>Previous review (2026-09-14)</summary>

Reviewed at head `282b7cddd9`. Full report: https://uav.tridgell.net/DevCallReviews/2026_09_14_AIReview/devcall_pr_reviews.html#prMAVProxy-1753

**Verdict: COMMENT — no blockers, two issues worth fixing.** The arc and loiter fixes match ArduPilot's AP_Mission, ArduPlane and logger behaviour, and reverting each fix makes tests fail.

### Issues
- **`MAVProxy/tools/mavflightview.py:523`: with the latest released pymavlink, drawing a tlog mission now crashes.**
  - pymavlink 2.4.49, the newest release, has no `MAV_CMD_NAV_ARC_WAYPOINT` or `MAV_CMD_DO_ORBIT`.
  - Those references are already on master (e.g. `mp_slipmap_util.py:45`) and `setup.py` still asks only for `pymavlink>=2.4.38`, but master never read tlog missions.
  - Tested with a tlog holding a 3-item mission: the base reads 0 items and draws; this head reads 3 and raises AttributeError. 32 of the new tests fail the same way under 2.4.49.
  - Fix: raise the requirement once a release has the constants, or look them up with `getattr`.
- **`MAVProxy/modules/lib/mp_util.py:256`: LOITER_TIME is still drawn at an explicit param3 radius that ArduPlane never flies.**
  - `Plane::verify_loiter_time()` calls `update_loiter(0)`, which uses `WP_LOITER_RAD` (or `LOITER_RADIUS_DEFAULT` if `|WP_LOITER_RAD| <= 1`). AP_Mission keeps only the sign of param3.
  - So a mission loaded from a file, or an upload read from a tlog, with param3=80 and `WP_LOITER_RAD=60` is drawn at 80 while the plane flies 60. With `|WP_LOITER_RAD| <= 1` the map draws a 1 m circle or nothing.
  - `tests/test_mission_geometry.py:200-202` asserts the current behaviour.

### Notes
- **Test gap:** reverting the callers at `mavproxy_map3d/__init__.py:344` or `MAVExplorer.py:858` to reading param4 directly (the bug 930075b7 fixes) still passes every test.
- **Test order:** `pytest tests/test_mission_geometry.py tests/test_mavflightview.py` fails `test_a_mission_given_to_draw_is_left_alone` with a TypeError. Loading MAVExplorer.py (`tests/test_mission_geometry.py:279`) sets `MAVLINK20=1` after the MAVLink1 dialect is already loaded. Restoring the variable in the test, or setting it in a conftest, fixes it. Also, the mavflightview tests skip only when cv2 is missing, but the tool imports wx.
- **DO_JUMP past the last item:** pymavlink's `view_indexes()` doesn't check that the target exists, so `polygon_list()` raises AttributeError. That was already reachable from a dataflash log or `--mission` file; it now also comes from a tlog, including a download that stopped early and is drawn by `LogMission.fill()` (`mavflightview.py:479`). Catching it around `wp.polygon_list()` in `mission_objects()` (`mavflightview.py:643`) would keep the plot going. Found by the Codex pass; it predates this PR.
- mavflightview still doesn't clear the mission on a dataflash "New mission" followed by zero items (`MISSION_TYPES` excludes `MSG`, `mavflightview.py:391`). MAVExplorer now does. Pre-existing.

Checked and correct: clockwise-positive arcs, endpoints and radius at high latitude and across ±180°; param4 crosstrack semantics, including NaN; the LOITER_TIME ±1 direction; "New mission"/"Mission: %u" log handling; the FTP parameter merge; 9 LogMission edge cases; no new module-scope MAVLink constants.

CI: 2 passing (flake8 and the Windows build; no workflow runs pytest).


</details>

tridge:

<details><summary>Previous review (2026-09-15)</summary>

**Automated review note — AI-generated (Claude), validated against the live diff (Claude + Codex cross-checked).** Please sanity-check before acting.

Re-reviewed at head `163018e2ed` (previously `282b7cddd9`); my earlier comment above is superseded. Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_15_1123/devcall_pr_reviews.html#prMAVProxy-1753

**Verdict: COMMENT, no blockers.** All six findings from my previous comment are fixed. Both reviewers showed each fix is tested: undoing it makes a test fail. The new flown-path code (`plane_track.py`) matches ArduPlane closely, but it has a main-thread cost in the live 3D map, one hang, and draws some altitudes wrong.

### Previous round
- **RESOLVED — tlog missions crashed with released pymavlink 2.4.49.** `mp_util.py:18-19` now names `MAV_CMD_DO_ORBIT` and `MAV_CMD_NAV_ARC_WAYPOINT` itself. All 35 MAVLink names the PR uses exist in 2.4.49. Going back to the dialect lookups gives 33 and 2 test failures under 2.4.49.
- **RESOLVED — LOITER_TIME drawn at its param3 radius.** `mp_util.py:269` uses `WP_LOITER_RAD`, falling back to 60 m (`:283`), as `update_loiter()` does. Undoing either part fails a test.
- **RESOLVED — no test on the param4 callers.** Reverting `mavproxy_map3d/__init__.py:375` or `MAVExplorer.py:961` now fails a test.
- **RESOLVED — test order.** `tests/conftest.py` restores `MAVLINK20`; the mavflightview tests also skip without wx.
- **RESOLVED — DO_JUMP past the last item.** `drawable_mission()` (`mavflightview.py:208`) uses the same bound as `AP_Mission::get_next_cmd`.
- **RESOLVED — "New mission" followed by zero items.** `MSG` is in `MISSION_TYPES` (`mavflightview.py:413`).

### New
- **ISSUE — `MAVProxy/modules/mavproxy_map3d/__init__.py:402`: the live 3D map re-flies a plane mission on the main thread whenever home moves.**
  - The cache key holds home's exact lat/lon. While disarmed, ArduPlane updates home every 5 s (`Plane.cpp:396-400`) and sends HOME_POSITION each time, so normal GPS wander misses the cache every time.
  - Measured: 0.05 s for the Kalaupapa mission, 0.2 s for a ~1.8 h survey, 5.4 s for a mission that reaches the 8 h cap. `missionpath` defaults to `flown`.
  - SITL GPS doesn't wander by default, so SITL testing won't show it.
  - Fix: ignore home moves under a few metres in the key, or compute off the main thread.
- **ISSUE — `MAVProxy/modules/lib/plane_track.py:336`: a mission crossing ±180° longitude runs for minutes.** `local()` subtracts longitudes without wrapping. A 223 m leg from 179.998° to −179.999° took 99 s and returned 1.87 million points; the same leg without the crossing takes 0.00 s. The 8 h cap is only checked between items, so it doesn't stop this. ArduPilot wraps the difference (`Location.cpp`, `diff_longitude`).
- **ISSUE — the flown path is drawn at the wrong altitude in two common cases.**
  - `mavproxy_map3d/__init__.py:335`: an item with no position, such as a LOITER_TO_ALT at 0,0, is passed with its altitude set to `None`. So it is flown at the current altitude. ArduPlane's `Location::sanitize()` replaces only the lat/lon and keeps a non-zero altitude.
  - `plane_track.py:764-767`: NAV_RETURN_TO_LAUNCH flies home at the current altitude. ArduPlane uses `get_RTL_altitude_cm()`, which is `RTL_ALTITUDE` above home (100 m by default) and only uses the current altitude when `RTL_ALTITUDE` < 0 (`altitude.cpp:99-105`). Changing `RTL_ALTITUDE` doesn't change the drawn path.
- **NOTE — smaller differences from ArduPlane:**
  - `plane_track.py:580`: a LOITER_TURNS of 2.5 is drawn as 2.5 turns, but AP_Mission stores the integer part, so ArduPlane flies 2.
  - `plane_track.py:451`: `ALT_SLOPE_MIN` <= 0 disables the glide slope in ArduPlane; the port keeps it.
  - A mission over 8 h silently falls back to the geometry drawing, after up to ~5 s of work.
- **NOTE — test gap:** changing `plane_track.py:559-561` to use param3 as the LOITER_TIME radius (the bug fixed in `mp_util`) still passes all tests.

Checked and correct:
- the L1 controller port: signs, units, bank limit, capture/circle switch;
- loiter radius and direction, turn counting, and lining up to leave;
- waypoint acceptance and pass-by;
- DO_JUMP/DO_JUMP_TAG, and termination for LOITER_UNLIM and repeat −1 (peak memory 2.8 MB);
- ArduPlane parameter defaults.

Tests: 169 passed in the three new files under both pymavlink 2.4.49 and master (Codex: 282 passed, 10 skipped for the whole suite). flake8 is clean. CI: 2 passing (flake8, Windows build); no workflow runs pytest.


</details>

tridge:

<details><summary>Previous review (2026-09-15)</summary>

Re-reviewed at head `918c407c8c` (you were last told `163018e2ed`); my earlier comment above is superseded. Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_15_1710/devcall_pr_reviews.html#prMAVProxy-1753

**All 8 findings from my previous comment are resolved, and 7 of the 8 fixes are shown tested by mutation. This push adds a MAVExplorer crash on older logs and a test that fails almost every run here, so the verdict is REQUEST CHANGES.** Both fixes are one line.

### Previous round
- **RESOLVED: map3d re-flying on the main thread as home wanders.**
  - The flight now runs on a worker thread, and a home move under `HOME_MOVE_CHANGE` (10 m) reuses the cached path (`mavproxy_map3d/__init__.py:414-419`, `:486-537`).
  - Setting `HOME_MOVE_CHANGE = 0` fails a test.
  - This holds for home-relative missions only; see the ISSUE below.
- **RESOLVED: antimeridian runaway.** `plane_track.py:351-354` and `terrain.py` both wrap the longitude difference. Removing either wrap fails a test, and a 300 m leg across 180° now takes 0.00 s.
- **RESOLVED: a positionless item flown at the current altitude.** `positionless_amsl()` (`plane_track.py:843-855`) keeps its altitude, as `Location::sanitize()` does. Dropping it fails a test.
- **RESOLVED: RTL flown at the current altitude.** `rtl_location()` (`plane_track.py:539-562`) follows `get_RTL_altitude_cm()` and the `AP_Rally` nearest-point/RALLY_LIMIT_KM/INCL_HOME logic, and matches the rally frame flags.
- **RESOLVED: LOITER_TURNS 2.5 drawn as 2.5** (`plane_track.py:628-632`). **RESOLVED: ALT_SLOPE_MIN <= 0** (`:470-472`). Undoing either fails a test.
- **RESOLVED by design change: the silent fallback after ~5 s for an 8 h mission.**
  - The cap is now 4 h, a mission is rejected up front by `least_flight_time()`, and the work is off the main thread.
  - The fallback to the geometry drawing is still silent while the style says "flown" (`mavproxy_map3d/elements.py:552`). It is cheap now.
- **RESOLVED: LOITER_TIME radius test gap** (`tests/test_plane_track.py:508`). Using param3 as the radius now fails it.

### New — please fix
- **BUG: `MAVProxy/tools/MAVExplorer.py:734` crashes the 3D map on any pre-4.5 log that has rally points.**
  - `rally[m.Seq] = (m.Lat, m.Lng, m.Alt, m.Flags)` needs `Flags`, which RALY only gained in 4.5: Copter-4.4.4 logs `"QBBLLh"` with `TimeUS,Tot,Seq,Lat,Lng,Alt`, Copter-4.5.0 logs `"QBBLLhB"` with `...,Flags`.
  - pymavlink's `DFMessage` raises AttributeError for a missing field, and nothing around the call at `:860` catches it. This affects every vehicle type.
  - Reproduced through `mission_from_log()` with a Flags-less RALY.
  - Fix: `getattr(m, 'Flags', 0)`, which is what `rally_amsl()` already assumes for flags 0, plus a test with an old-format RALY.
- **ISSUE: `tests/test_plane_track.py:932-936` (`test_a_path_for_a_mission_since_changed_is_not_drawn`) is racy.**
  - The second `send_mission()` goes in before the worker has taken the first. `fly_tracks()` then correctly merges them and flies only the newer mission, which the assertion at `:943` rejects.
  - It failed 9/10 in one pass and 10/10 when I re-ran it on Python 3.14. The cross-check reproduced it too.
  - Adding `self.wait_for(lambda: calls)` before changing the mission (the pattern `flying_slowly()` already uses) makes it pass 10/10.
  - No workflow runs pytest, so CI can't show this.
- **ISSUE: `mavproxy_map3d/__init__.py:442-451`: a mission with any AMSL-frame item still re-flies on every home update.**
  - The cache key stores every altitude as `amsl - home[2]` (`:445`, and `:450` for rally points). That is constant only for home-relative items, so an AMSL item's key changes whenever home's altitude wanders.
  - Measured with your `live_module` harness, 10 home moves of 1.5 m with ±0.3 m altitude: the relative mission flew once in 11 sends, and the same mission with one item switched to frame 0 flew 11 times.
  - With the flight slower than the home interval, most results are dropped as stale, so the path is rarely drawn.
  - Fix: key AMSL altitudes absolutely, and only relative ones against home.

### Notes (non-blocking)
- **`MAVExplorer.py:731-732` resets the rally table whenever a RALY with `Seq == 0` arrives.**
  - ArduPilot also logs a single RALY each time one point is set (`AP_Rally.cpp:131`), so editing only point 0 mid-log makes MAVExplorer forget points 1..n and pick RTL from the wrong set.
  - A full upload re-logs every point, so this is an edge case. Clearing on a change of `Tot`, or dropping entries with `Seq >= Tot`, avoids it.
- **The flown RTL path ends at the RTL point** (`plane_track.py:817-820`, `:834-837`), whereas ArduPlane then loiters there at `abs(RTL_RADIUS)` in the direction its sign gives (`ArduPlane/mode_rtl.cpp:98-103`). This isn't new in this push, but a circle there would match what the aircraft does.
- **`plane_track.py:937`:** `least_flight_time()` now runs outside the `try` at `:941-944`, so a NaN DO_JUMP param raises ValueError out of `mission_track()`. The map3d thread catches that; MAVExplorer's `fly()` doesn't. Moving it inside the try fixes it.
- **`mavproxy_map3d/__init__.py:513`:** `moved_home()` also shifts legs to AMSL items by home's altitude change. It is bounded by the 10 m tolerance, so it's cosmetic.
- **Test gaps:** removing the `least_flight_time()` pre-check still passes `test_a_mission_too_long_to_fly_is_not` (`tests/test_plane_track.py:407`); only the run time changes. The cross-check also found that allowing a DO_JUMP target equal to the item count passes the invalid-jump test.

### What was checked
- **Released pymavlink:** 22 of the 24 MAVLink names this PR's lines use exist in PyPI pymavlink 2.4.49. The other two are the literals at `mp_util.py:18-19`, so nothing can fail at import.
- **Degenerate-mission fuzz:** all of these finished within 0.42 s with sensible output: jump loops, zero radius, repeated points, LOITER_UNLIM, NaN params, near the pole, and across the antimeridian.
- **Tests, from the checkout under released pymavlink:**
  - Targeted: 146 passed, 1 failed (the racy test), 47 skipped for lack of wx.
  - Cross-check, full suite: 306 passed, 10 skipped, the same one failure.
- **Other checks:**
  - flake8 (the CI command) is clean.
  - vermin finds nothing new in the added lines.
  - `mp_util.py`, `mavflightview.py` and `conftest.py` are unchanged since last round.
- **CI:** flake8 passed. The Windows build was still running when checked.

**REQUEST CHANGES:** please fix the RALY `Flags` crash and the racy test. The AMSL cache key is worth doing in the same pass.


</details>

AP-Review:

Re-reviewed at head `fe8c38484e` (you were last told `918c407c8c`); my earlier comment above is superseded. Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_16_1759/devcall_pr_reviews.html#prMAVProxy-1753

(My previous comment above is superseded by this one. It's left un-collapsed because these reviews now post from the `AP-Review` account, which can't edit the earlier comments posted under `tridge`.)


**Every blocking finding from my previous comment is resolved, and all but one are locked in by a test that fails when the fix is reverted. Verdict REQUEST CHANGES → COMMENT.** I am also withdrawing part of one of my own findings. The new items below are accuracy problems in the drawn path rather than blockers.

### Previous round

- **RESOLVED: the RALY `Flags` crash on pre-4.5 logs.** `MAVExplorer.py:745` now reads `getattr(m, 'Flags', 0)`, with a test at `tests/test_plane_track.py:1571`. Reverting to `m.Flags` fails that test, and a real `DFMessage` built from the Copter-4.4.4 `QBBLLh` format now goes through `mission_from_log()` cleanly.
- **RESOLVED: the racy path test.** The `wait_for` is at `tests/test_plane_track.py:1139`. **25/25 passes** at this head; removing the wait gives **6/25**.
- **RESOLVED: the `Seq == 0` rally reset.** `MAVExplorer.py:739-744` prunes on a change of `Tot` and ignores `Seq >= Tot`. Reverting fails `test_mavexplorer_takes_rally_points_from_an_older_log`.
- **RESOLVED: RTL ending at the point instead of loitering.** `plane_track.py:843-852` appends a `NAV_LOITER_UNLIM` with `param3 = self.rtl_radius`, so the sign gives the direction and 0 falls back to `WP_LOITER_RAD`. The test at `:389` covers `RTL_RADIUS` 0 / +250 / −250 and fails when the appended loiter is removed. (See the new item below about the *approach* to that circle.)
- **RESOLVED: `moved_home()` shifting AMSL legs.** The `alt_moves` argument at `mavproxy_map3d/__init__.py:535`, applied at `:545`. Always-shifting fails a test.
- **RESOLVED: both test gaps.** `tests/test_plane_track.py:445` catches the removal of the `least_flight_time()` pre-check, and `:470` catches a DO_JUMP target equal to the item count. Both mutations now fail.
- **RESOLVED in code, still untested: `least_flight_time()` inside the `try`.** It is at `plane_track.py:980-984` and a NaN `DO_JUMP` now returns `None` instead of raising out of `mission_track()`. But moving the call back outside the `try` still passes the whole suite — there is no `nan` anywhere in the test file. One test would lock it in.
- **PARTIALLY RESOLVED, and I am withdrawing half of my own finding: the AMSL cache key.** `keyed_alt()` at `mavproxy_map3d/__init__.py:462-464` is the fix I asked for, and it works — an all-AMSL mission now flies **once in 11 sends** where it previously flew 11 times. The `key += (round(home[2], 2),)` at `:479-482` deliberately re-adds home altitude for a *mixed* mission, and **you are right that a mixed mission cannot be served from cache**: `moved_home()` shifts the whole path by one `dalt`, which is wrong when only some legs move with home. Re-flying is the honest answer and both halves are tested.

  **Withdrawn:** I claimed last round that the re-flights were "mostly dropped as stale, so the path is rarely drawn". I could not reproduce that at a realistic home rate. Measured flight times at this head: a 40-item survey takes 0.06 s, 240 items 0.34 s, 400 items 0.45 s — against a 5 s home interval. The staleness only appears if the home interval is forced down to 0.05 s. Please disregard that part.

### New — the drawn path can be silently wrong or silently absent

- **BUG: a `LOITER_TURNS` above ~40 turns silently produces no flown path at all.** `plane_track.py:653` sets `extra = MAX_LAPS * seconds_per_lap` — a flat 20-lap allowance that is not scaled by the requested `param1`. So a legitimately long loiter exhausts its own time limit and `mission_track()` returns `None`. Measured with the test fixture's parameters, a loiter followed by a waypoint:

  ```
  radius 80 m (WP_LOITER_RAD default): drawn up to 39 turns, NOT DRAWN from 40
  radius 200 m:                        drawn up to 32 turns, NOT DRAWN from 33
  ```

  The comment at `:69-76` describes `MAX_LAPS` as a stuck-detector — "the aircraft is taken to be stuck, circling a waypoint it cannot turn tightly enough to reach" — so I don't think a 40-turn loiter is an intended exclusion; it just falls under the same budget. Budgeting the requested turns separately from the stuck allowance would fix it. This lands badly in combination with the next item.

- **ISSUE: the fallback from "flown" to geometry is still silent, and the setting still reads `flown`.** `mavproxy_map3d/__init__.py:444-447` and `elements.py:576-583`. Measured on a copter with `missionpath='flown'`: tracks handed to the viewer `[None]`, setting still reads `flown`, nothing printed. Same for a mission declined for length, and now also for the turn-count case above. A user who explicitly types `map3d set missionpath flown` gets geometry with no indication that the thing they asked for could not be produced. One `print()` when the requested style cannot be honoured would cover all three.

- **ISSUE: the drawn RTL flies through the centre of the return circle before capturing it.** `plane_track.py:840` completes waypoint navigation to the RTL location and only then appends the loiter and calls `fly_loiter()`. ArduPlane calls `update_loiter()` during RTL navigation (`ArduPlane/mode_rtl.cpp:98`), so it captures the circle on approach. Measured at this head, a 3 km return with `RTL_RADIUS = 250`:

  ```
  closest approach to the RTL point, final half of the path: 0.53 m
  the path then settles onto the circle at 267 m
  ```

  So the drawing shows an overflight of the centre that the aircraft will not fly. (These are model measurements, not SITL.) The 267 m against a requested 250 m is the L1 flown-radius adjustment and looks right.

- **ISSUE: a positionless item is classified from its raw frame, so an all-terrain mission is re-flown on every home update.** `fixed_alt.append(frame in (0, 5))` appears twice with the same expression but different data: at `mavproxy_map3d/__init__.py:354` (positionless branch) `frame` is still the raw wploader frame, while at `:404` (positioned branch) it has already been rewritten to 0 by the terrain resolution at `:365-368`. Both my passes found this independently. Measured over 11 sends:

  ```
  AMSL waypoints + positionless DO_CHANGE_SPEED at frame 3   -> flown 11 times
  AMSL waypoints + positionless DO_CHANGE_SPEED at frame 0   -> flown  1 time
  terrain waypoints + positionless DO_CHANGE_SPEED, frame 10 -> flown 11 times
  ```

  The frame-10 case is the clean defect: `positionless_amsl(z, 10, home_amsl)` returns `None`, so that item contributes `None` to the key and nothing about it moves with home — yet it is counted as moving and forces the whole mission to be re-flown. Resolving the frame before `:354`, or treating an item whose `amsl` is `None` as neither fixed nor moving, would fix it. An altitude-independent command like `DO_CHANGE_SPEED` arguably shouldn't enter this classification at all.

- **ISSUE: for a genuinely mixed mission the cache key pins home altitude to 1 cm, while `same_home()` tolerates 10 m.** `mavproxy_map3d/__init__.py:482`. So even a correctly-classified mixed mission re-flies on essentially every home update. Measured over 10 home updates wobbling by a few cm: all-relative 1 re-fly, all-absolute 1 re-fly, **mixed 10 re-flies**. Replacing `round(home[2], 2)` with a quantisation at the granularity `same_home()` already tolerates gives 1 re-fly for the mixed case. Mixed is an ordinary shape here — `send_mission()` sets `frame = 0` for a terrain item once terrain resolves (`:366-369`), so a terrain-following plane mission is mixed by construction, and ArduPlane re-sends `HOME_POSITION` every 5 s while disarmed (`ArduPlane/Plane.cpp:397-404`). Cost is ~32 ms per fly-through on your own `tests/missions/plane-mission-geometry.txt`, so it is minor, but `reset_flown_track`'s own docstring notes a flight "can take a second or more".

- **ISSUE: the drawn rally marker disagrees with where the flown path returns to.** `send_rally()` (`mavproxy_map3d/__init__.py:677-686`) sends the raw `r.alt` and `elements.py:898-903` draws it at `home_amsl + alt` regardless of frame, while the new `rally_points()` at `:504` resolves the frame properly. With a rally point at alt=100, home 584 m, origin 500 m, terrain 300 m:

  ```
  frame           drawn marker AMSL    RTL destination AMSL
  absolute              684.0                 100.0
  above home            684.0                 684.0
  above origin          684.0                 600.0
  above terrain         684.0                 400.0
  ```

  Relatedly, the two new `self.send_rally()` calls in the `GPS_GLOBAL_ORIGIN`/`ORGN` handlers at `:816` and `:821` cannot change anything — `send_rally()`'s output is identical after the origin changes — so they are no-ops as written.

- **ISSUE: `MAVExplorer.py:987` re-arms `mission_to_fly` even when the flown path already came back `None`.** `m3d.mission_to_fly = (mission, fly) if track is None else None` doesn't check that `missionpath` was already flown, so the next unrelated `set` re-runs the whole computation (measured: `plane_mission_track` calls go 1 → 2 after `set showdirection true`). Adding `and mestate.settings.missionpath != 'flown'` covers it.

### Adjusted from the cross-check

- My cross-check reported the unresolved-terrain case as a bug: `plane_track.py:374` replaces an unknown AMSL with the aircraft's current altitude, and terrain-relative waypoints at 100 m AGL were measured drawing a path flat at 584 m AMSL. **I have downgraded it**, because `item_amsl()` at `:321-324` documents `None` as "until then nobody knows" and `:769-772` re-runs `send_mission()` once a deferred terrain lookup lands, so the wrong path is transient. The residual worth noting is the case where terrain *never* resolves (no tile, no network, terrain module off): the path then stays silently at the aircraft's altitude, because `location()` cannot distinguish "unknown" from "keep current altitude". A one-line note to the user, or suppressing the flown path while terrain is outstanding, would close it.

- My cross-check also reported that `test_a_loiter_is_left_lined_up_on_what_comes_next` (`tests/test_plane_track.py:152`) passes with the exit-alignment check disabled. **That part is true** — I reverted `lined_up()` to `return True` and that test still passed. But it is **not** a coverage gap: the same mutation fails `test_the_path_is_flown_in_the_order_flown` and `test_the_path_is_the_one_flown`. So the behaviour is covered; it is only that the test named for it asserts eventual convergence on the next leg rather than heading at the loiter exit. Worth a tighter assertion, nothing more.

### Notes (non-blocking)

- **`mavproxy_map3d/__init__.py` goes from 0 to 2 flake8 violations** — `:408 E128` and `:461 E306` — and `MAVExplorer.py` gains one `E126`. CI cannot see any of them: `scripts/run_flake8.py` only checks files carrying `AP_FLAKE8_CLEAN`, and that file doesn't carry it (it was clean before this PR, so the marker would be worth adding).
- **58 of the 156 new tests (37%) silently skip without wxPython**, which is in the `recommended` extra, not `install_requires` — and CI runs no tests at all (`.github/workflows/` is flake8 + a Windows build; `grep -rn pytest .github/workflows/` is empty). So nothing anywhere verifies more than a third of the new suite.
- `:425` has a stale comment — `track_wanted` is a 5-tuple now, not `(key, flown_from, home, items)`.
- `MAVExplorer.py:127-129` duplicates the style list and default as literals instead of using `map3d.MISSION_STYLES`, which the live module does use; they can drift.
- `mavproxy_map/__init__.py:166` binds `self._label_suffix_for_wp_command` to the shared `mp_util.MISSION_LABEL_SUFFIXES` dict. Nothing aliases it today, but a future per-instance edit would mutate the table the 3D map reads.

### What was checked

- **The PR fixes 8 real failures that master has under released PyPI pymavlink 2.4.49** — the `MAV_CMD_DO_ORBIT` `AttributeError` at `mavproxy_map/__init__.py:163` and the `test_mission_geometry` cases. That is exactly the class of bug CI cannot see, and it is worth calling out as a win rather than leaving it in the noise.
- **Released-pymavlink safety verified for real**, not by eye: every module-scope `mavutil.mavlink.*` reference across all ten changed source files enumerated by AST and checked against 2.4.49 — 11 references, 0 missing. The two that 2.4.49 genuinely lacks (`MAV_CMD_DO_ORBIT`, `MAV_CMD_NAV_ARC_WAYPOINT`) are now numeric literals 34 and 36 in `mp_util`, which I checked against `common.xml`. All five touched modules import cleanly under 2.4.49.
- **Suite: 352 passed, 10 skipped** on Python 3.10.21 / 3.11.16 / 3.12.14 / 3.13.15 / 3.14.4, under both released 2.4.49 and pymavlink git master — identical on all ten combinations. Five consecutive full runs identical. 13 thread- and timing-sensitive tests run 25× each in isolation: **325/325 passes**, and 25/25 again under loadavg 47 on a 32-core box. No flake found; I could not reproduce last round's Python-3.14 failure at this head.
- **The new labelling work is genuinely covered**: six mutations of `test_map3d_mission_labels.py`'s targets, six failures, no survivors. Rendered offscreen to confirm the labels are actually visible (non-black pixels 1459 → 1847 → 1459 on toggle, image byte-identical after toggling back, no actor leak). Edge cases are clean — unknown `MAV_CMD` falls through `.get()` rather than raising, NaN lat/lon/alt and `seq=None` produce no exception, 700 labels refresh in 0.065 s.
- Setting validation is correct on both sides: `set missionpath bogus` prints `Must be one of ['flown', 'geometry', 'plain']` and leaves the value unchanged; `PLAIN` is accepted case-insensitively. (`elements.set_mission_style()` itself ignores an unknown style silently, but the setting layer never lets one through.)
- Thread-safety of this push's changes: `track_request`/`track_result`/`track_thread` all touched under `track_lock`, `track_wanted` only on the main thread, all three assignments widened to 5-tuples consistently with the single `wanted[4]` read, and the stale-result guard `wanted[0] != key` exercised.
- Nothing in the arc-drawing code is new here — this push's only change to `mp_slipmap_util.py` and `mavproxy_map/__init__.py` is sourcing the two command numbers from `mp_util`; the arc geometry is already on master.
- CI snapshot: both `build` and `build (3.10)` passing.

**Not reviewed:** the ported `rally_alt_frame`/`rally_amsl` frame decoding against `AP_Rally::rally_location_to_location`, `mission_from_log`'s new `RALY`/`ORGN` handling against a real log with rally points (I have none), and `path_view()`'s span maths on an antimeridian-crossing log.

**COMMENT** — nothing blocks merge. The turn-count cap and the silent fallback are the pair I'd most want fixed, since together they mean a legitimate mission loses its flown path with no explanation.



</details>

AP-Review:

Re-reviewed at head `3e5a9b7062` (previously `fe8c38484e`); my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_17_0037/devcall_pr_reviews.html#prMAVProxy-1753

---

**Verdict: COMMENT** (was REQUEST CHANGES) — twelve of fifteen previous findings are fixed, and 24 of 26 mutations I tried are caught by the new tests, each with a named failing test. Nothing here blocks merge. Two new accuracy findings, one of which corrects something **I got wrong last round**.

### Previous round

Resolved: the `LOITER_TURNS` time budget (radius 80 m now draws at least 295 turns, was not drawn from 40); the silent flown→geometry fallback; the RTL flying through the circle centre; positionless-item classification; the mixed-mission cache key; the rally marker and the no-op `send_rally()` calls; `MAVExplorer` re-arming `mission_to_fly`; `least_flight_time()` inside the `try`; the flake8 regression (`map3d/__init__.py` now carries `AP_FLAKE8_CLEAN`); the stale `:425` comment; the duplicated style list; and the shared label dict.

Still open, all re-raised rather than new: **CI runs no tests at all** (repo-level, not yours — with wx blocked 88 tests skip against 13 normally); **terrain that never resolves** leaves the path silently at the aircraft's altitude (`plane_track.py:374`), which `say_unflown` does not cover because a path *is* produced; and `test_a_loiter_is_left_lined_up_on_what_comes_next` still asserts convergence rather than exit heading (cosmetic).

### ISSUE — `MAVProxy/modules/lib/plane_track.py:1052` — QRTL draws an ordinary glide slope, not ArduPilot's approach ramp

**I need to correct my previous comment here.** Last round I traced `Q_RTL_MODE` 1/2/3 and said the port was correct. It is not, and my reasoning was faulty: I inferred correctness from the fact that mutating "`Q_RTL_MODE` 3 returns at `RTL_ALTITUDE`" is caught by the suite. It is caught — by `tests/test_plane_track.py:695`, which asserts on `track[-1]`, the **final point only**. That pins the destination altitude and says nothing about the 3 km of return in between.

`:1034-1052` sets the destination to `home_amsl + q_rtl_alt` and then calls `fly_waypoint()`, which drives altitude only through `MissionWalk.update_target_altitude` (`:504-523`) — a port of fixed-wing `Mode::update_target_altitude`, i.e. a proportional glide slope. There is no port of `ModeQRTL::update_target_altitude` in the file.

ArduPilot does something else. `Q_RTL_MODE=3` enters QRTL immediately (`mode_rtl.cpp:21-25`); in fixed-wing cruise the motors are not `THROTTLE_UNLIMITED`, so `mode_qrtl.cpp:6-77` falls through to `do_RTL(home.alt + qrtl_alt*100)` and `poscontrol_init_approach()`. With `Q_TRANS_DECEL=2`, `transition_threshold()` is 181.5 m, so at 3 km the state is `QPOS_APPROACH` and `mode_qrtl.cpp:186-215` **overrides** the fixed-wing slope, holding `RTL_ALTITUDE` until `sink_dist` from home.

Executed with `RTL_ALTITUDE=100`, `Q_RTL_ALT=15`, `Q_RTL_MODE=3`, mission `[waypoint(3000,0), RTL]`:

| distance from home | ArduPilot target | model draws | error |
|---|---|---|---|
| 1200 m | 100.00 m | 50.23 m | −49.8 |
| 990.9 m | 100.00 m | 44.73 m | **−55.3** |
| 807 m | 100.00 m | 39.79 m | **−60.2** |
| 501 m | 61.50 m | 31.54 m | −30.0 |
| 195 m | 19.77 m | 23.31 m | +3.5 |

(`rtl_alt_delta=85`, `sink_dist=623.3 m`, so the real ramp reaches `RTL_ALTITUDE` at 783.3 m.) Not version-sensitive — the ramp has been there since `3bb840f794` (2021-06-01) and is still on master. Only `Q_RTL_MODE=3` is affected: mode 1 draws 100 m all the way, which matches RTL, and mode 2 goes through `fly_vtol_approach`. Accuracy only; the horizontal path is unaffected. The comment at `:1035-1036` is right about the destination and silent about the approach.

Your own fixture cannot see it either: `sitl-quadplane-rtl-q-rtl-mode-3.json` was recorded with `RTL_ALTITUDE=20`, `Q_RTL_ALT=15`, so `rtl_alt_delta=5 m` and the ramp is flat there by construction.

### ISSUE — `tests/test_plane_track.py:1017` — the SITL comparison is horizontal only

`distances()` projects to `(north, east)` and drops the third element; `arrivals()` uses `gps_distance`. Making every drawn point **200 m too high** leaves all **20** `TestFlownMission` assertions green (`20 passed in 1.78s`). The recordings already carry altitude (`record_flight.py:141` writes `round(m.Alt, 1)`), so a vertical assertion is nearly free — and it is the highest-value thing you could add, since this PR moves a lot of altitude logic and none of it is checked against a real flight. On the mode-3 recording drawn is 4–7 m high throughout (at 990 m: flown 70.1, drawn 74.2) and nothing looks.

### ISSUE — `tests/missions/*.json` — the fixtures cite commits and autotests that do not exist publicly

`gh api repos/ArduPilot/ardupilot/commits/cd2319308b` → 422 "No commit found"; same for `f2c18f532f`, and same against `peterbarker/ardupilot`. `Plane.MissionItemTypes`, `QuadPlane.VTOLMissionItemTypes` and `QuadPlane.KalaupapaCanyonRun` are in neither `arduplane.py` nor `quadplane.py` on master, GitHub code search returns 0 for all three, and no open ArduPilot PR mentions them. So the ground truth for the headline claim of `3e5a9b70` cannot be regenerated by anyone else, and "ArduPilot's autotest *now has* flights meant for this" is not true upstream yet. Either land the ArduPilot side first and cite resolvable SHAs, or say in the commit and in `source` that these came from a working branch. (Still better than the old canyon fixture, which cited no SHA at all.)

### Notes

- `map3d/__init__.py:909` — a non-finite `WIND.direction` raises out of `mavlink_packet`: `nan` → `ValueError`, `inf` → `OverflowError` (driven through the real handler). ArduPilot sends `wrap_360(degrees(atan2f(...)))`, NaN if the EKF wind estimate is NaN. `mavproxy_link.py:1139` catches it, so at default `moddebug 0` the message is silently dropped. One-line `math.isfinite` guard.
- `tests/test_map3d_mission_labels.py:204` — imports `map3d_ui` (→ `wx`) at line 204, before the fixture's `pytest.importorskip("wx")` at 205, so it **fails** rather than skips without wxPython. New in `ce2179a5`; it is the only one of your tests that fails in that configuration.
- `terrain.py:89` — the new whole-fetch deadline cannot interrupt a slow buffered read, because `response.read(64*1024)` blocks until its buffer or EOF and each `recv` resets the socket timeout. Measured: `timeout=0.1` with a 4096-byte body dripped at 64 B/50 ms → 3.163 s; `timeout=5.0` with 65536 bytes at 2048 B/400 ms → 12.410 s. **This is still an improvement on master**, which had no whole-fetch deadline at all — the guard is just partial. One caller, already wrapped in `try/except`, so it is a nit. (Also `:94-95` uses `%.0f`, so `timeout=0.1` prints "took longer than 0s".)
- `record_flight.py:117` — `if value:` drops a legitimately-zero `heading`/`approach`; a due-north approach would be recorded with no `approach` and replayed at `STILL_AIR_APPROACH`, a 180° error. `if value is not None` is the intent.
- `test_plane_track.py:1068` — `near = abs(params[1]) * 1.4` is 0 for a `LOITER_TO_ALT` with param2 0 (meaning `WP_LOITER_RAD`). Latent; no current fixture has one.

Suite at this head: 414 passed, 13 skipped, 20 subtests on Python 3.14.4 / pymavlink 2.4.49; five full runs identical under randomised ordering; the six timing-sensitive `TestDrawnTrack` tests 25/25 in isolation. 11 module-scope `mavutil.mavlink.*` references enumerated by AST across the 13 changed files, **0 missing under released 2.4.49**.


</details>

AP-Review:

<details><summary>Previous review (2026-09-18)</summary>

Re-reviewed at head `b669fe9444` (previously `3e5a9b7062`); my earlier AI comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_18_1127/devcall_pr_reviews.html#prMAVProxy-1753

**Both accuracy ISSUEs from my previous comment are properly fixed, and I locked each one in with the exact mutation that exposed it last round.** Four new commits, all of them responding to that comment. **Verdict: COMMENT** — everything new this round is a NOTE.

### Resolved since last round

- **QRTL drew an ordinary glide slope, not ArduPilot's approach ramp.** `plane_track.py:532-564` is a new `update_qrtl_target_altitude()`, dispatched at `:510-512` and armed at `:1101`. I diffed it line by line against `ArduPlane/mode_qrtl.cpp` on master — `radius`, `rtl_alt_delta`, `sink_time`, `sink_dist`, `rad_min`/`rad_max`, the approach-start latch, the `ALT_SLOPE_MIN` gate, the far-field interpolation and the don't-climb-back-up clamp all match, including the quirk that raw `WP_LOITER_RAD` is used with no `LOITER_RADIUS_DEFAULT` fallback; 10,000 numeric comparisons differed by < 0.00027 m. **Mutation-verified three ways:** removing the dispatch fails the new unit test *and* the mode-3 flight comparison; removing the clamp fails `test_qrtl_does_not_climb_back_up_to_its_slope`; neutering the `ALT_SLOPE_MIN` gate fails `test_qrtl_comes_home_on_its_own_altitude_profile`.
- **The SITL comparison was horizontal only.** `tests/test_plane_track.py:1122-1148` adds `heights()` (3D nearest-point-on-segment) and `:1249-1256` adds `test_the_path_is_flown_at_the_height_flown`, with p90/worst limits per fixture. **The same +200 m probe that left all 20 assertions green last round now fails 10 of 10 fixtures.** Margins are ~1.4× actual — median absolute height error 0.5–4.2 m across the ten recordings, p90 3.8–16.8 m against limits of 6–25 m. Tight enough to be worth something, loose enough not to be brittle. This is the highest-value thing in the push, and re-recording the mode-3 fixture at `RTL_ALTITUDE=60` (was 20) is what makes it actually discriminate the ramp — that was the specific weakness I raised.
- **The fixtures citing commits and autotests that did not exist publicly.** Mostly closed, and **I need to retract part of what I said**: the `source` strings now name ArduPilot PR 34424, which exists, is open, and whose diff really does add `def MissionItemTypes`, `def KalaupapaCanyonRun` and `def VTOLMissionItemTypes` — so my claim that these autotests were nowhere public is out of date. `cd2319308b`, `f2c18f532f` and `6ea2776f35` all resolve. Residual below.
- **A non-finite `WIND.direction` raised out of `mavlink_packet`.** `map3d/__init__.py:906` now guards with `math.isfinite`; reverting reproduces the NaN `ValueError` and fails the new `test_a_wind_with_no_estimate_changes_nothing` (which covers `nan`, `+inf`, `-inf`).
- **`test_map3d_mission_labels.py` failed rather than skipped without wxPython.** `:205` now calls `map3d_frame()` before importing `Map3DFrame`; it skips with wx blocked.
- **The terrain whole-fetch deadline couldn't interrupt a slow buffered read**, and `%.0f` printed "0s". `terrain.py:89` uses `read1` and `%g`. Re-measured against a real local dripping HTTP socket: 4096 B at 64 B/50 ms with `timeout=0.1` went from **3.166 s to 0.108 s**, and 65536 B at 2048 B/400 ms with `timeout=5.0` from **12.406 s to 5.203 s** — reproducing my previous numbers exactly. *Residual:* the deadline still isn't hard — slowly delivering a valid chunk header delays the timeout to 0.404 s with `timeout=0.1`, because `read1()` waits for that header. Still a large improvement; noting it so the guard isn't mistaken for absolute.
- **`record_flight.py` dropped a legitimately-zero heading/approach.** New `extras()` at `:51-59`; restoring `if value:` returns `{}` and fails `test_a_course_of_north_is_recorded`.
- **`near = abs(params[1]) * 1.4` was 0 for a default-radius `LOITER_TO_ALT`.** Fixed at `:1178-1182` with a named regression test — see N3 for the remaining hole one level up.

Still open and re-raised rather than new, all outside this PR: CI runs no tests at all (`grep -l pytest .github/workflows/` returns nothing, so `tests/` isn't even linted — and the 429 tests here are never executed); unresolved terrain still leaves the path silently at the aircraft's altitude (`plane_track.py:406-407` — a live probe with unresolved 100 m AGL waypoints produced 291 points all at 584 m AMSL with the style still `flown`, so `say_unflown` doesn't cover it); and `test_a_loiter_is_left_lined_up_on_what_comes_next` still asserts convergence rather than exit heading.

### New — all NOTEs

**N1 — `plane_track.py:551-558` — the far-field QRTL ramp is ArduPilot-master-only; no released firmware flies it.**

The gradual descent from the approach-entry altitude down to `RTL_ALTITUDE` was added in `bd9c0285fc` (2026-08-20) and refined by `3dc0c3621b` (2026-08-22). `git tag --contains` confirms neither is in any `Plane-4.7.x` tag — both are master-only. `Plane-4.7.1` still has the single `linear_interpolate(0, rtl_alt_delta, dist, rad_min, dist_rtl_alt_reached)`, which clamps beyond `dist_rtl_alt_reached`, i.e. holds `RTL_ALTITUDE` with no slope.

Measured with the new test's own parameters (entered at 300 m, `RTL_ALTITUDE=100`, `Q_RTL_ALT=15`, `RTL_RADIUS=250`, 3 km return):

| dist from home | drawn (master) | ArduPlane 4.7.1 | error |
|---|---|---|---|
| 2500 m | 248.8 | 148.7 | **+100.2** |
| 2000 m | 199.4 | 100.7 | **+98.7** |
| 1500 m | 147.8 | 100.0 | +47.8 |
| 1123 m | 110.7 | 100.0 | +10.7 |
| 900 m | 83.3 | 82.6 | +0.7 |

Below `reached` (1123 m here) the two agree. Modelling the released behaviour against your mode-3 recording gives a p90 height error of 25.9 m against a 6.0 m limit, so the fixture genuinely discriminates them — the choice is real rather than academic.

I don't think this blocks anything: tracking master is a defensible call, and the previous state (an ordinary glide slope) was wrong for *both*. But MAVProxy is mostly pointed at released firmware and the divergence is ~100 m. One line in the `update_qrtl_target_altitude()` docstring saying the far-field slope is master (post-`bd9c0285fc`) and that 4.7 and earlier snap to `RTL_ALTITUDE` would save the next reader comparing against a 4.7 flight.

**N2 — `tests/missions/sitl-quadplane-*.json` — the six QuadPlane recordings still cite a SHA that resolves nowhere.** All six cite `ardupilot 8d7ed31dfe`; it returns 422 "No commit found" from *both* `ArduPilot/ardupilot` and `peterbarker/ardupilot`, and it isn't one of PR 34424's three commits. And these are exactly the six fixtures whose **flight data was re-recorded** this round — comparing parsed JSON, `flown`, `home`, `items`, `heading` and `start` all changed in all six, with mode-3 additionally moving `RTL_ALTITUDE` 20 → 60. The four Plane fixtures changed `source` only. Separately, `grep RTL_ALTITUDE` over PR 34424's `Tools/autotest/quadplane.py` patch returns nothing, so the published `VTOLMissionItemTypes` wouldn't reproduce the 60 m recording. Net: the six fixtures whose data is new are the six nobody else can regenerate. Much better than last round; to close it, push the `RTL_ALTITUDE` change into PR 34424 and re-cite a commit from it. (The commit message for `2aa370fa` is already honest about `RTL_ALTITUDE 60`, so it's just the `source` strings catching up.)

**N3 — `tests/test_plane_track.py:1178-1181` — the new `WP_LOITER_RAD` fallback reintroduces the same hole one level up.** It stops at the parameter itself and doesn't apply ArduPlane's `|WP_LOITER_RAD| <= 1 → LOITER_RADIUS_DEFAULT` rule, which the model *does* apply at `plane_track.py:710`. Executed: `WP_LOITER_RAD=0` gives helper `near=0.0` against model `radius=60.0`, so a fixture with `WP_LOITER_RAD` 0 or ±1 and a default-radius `LOITER_TO_ALT` would report the item as never reached — the failure just fixed, moved up a level. Latent; no current fixture has it. Factoring the rule out of `:710` into a shared helper would stop the two drifting again.

**N4 — `tests/test_plane_track.py:1122-1148`** — `heights()` measures to the 3D-nearest point, which is right for a spiral (as its docstring says) but means a uniformly-too-high path can partly match a different part of itself. Clearly not fully absorbed — the +200 m mutation fails all ten, and the medians are too good for it to be hiding much. Recording it only so it isn't mistaken for a strict vertical-offset check; no change needed.

**N5** — `plane_track.py:1205-1207`: `linear_interpolate`'s polarity swap is unreachable from both call sites and deleting it leaves the suite green (429 passed). It's a faithful port of `AP_Math.cpp`'s, so keeping it is right — flagged only so it isn't mistaken for covered code. And `tests/test_plane_track.py:1161` has a 117-character docstring line from a botched rewrap (passes flake8 at `max-line-length=127`).

### Checked clean

The new `linear_interpolate()` is a faithful port of `AP_Math.cpp`'s, polarity swap and endpoint behaviour included. `plane_track.py:952` (`NAV_VTOL_LAND` flown in at the aircraft's altitude) is an exact port of `QuadPlane::do_vtol_land`'s `set_next_WP()` followed by `next_WP_loc.copy_alt_from(current_loc)` — *including* ArduPilot leaving the already-computed `offset_altitude` stale; neutering it fails two fixtures. The QPOS window needs no finding: `APPROACH < AIRBRAKE < POSITION1 < POSITION2`, so the ramp applies at every entry distance and the model's unconditional application is right. `self.qrtl_start` needs no reset because `fly_rtl()` runs once and returns. Suite at this head: **429 passed, 13 skipped, 20 subtests** on Python 3.14.4 / pymavlink 2.4.49, identical across three runs; 11 of 12 mutations tried are caught with a named failing test. With wx blocked the suite gives 89 skipped, 6 failures and 2 collection errors — my previous "88 skip" figure was incomplete, though the `test_map3d_mission_labels` fix itself does work.

Reviewed by: Claude + Codex (cross-checked) — a full read of the delta, the thread and the surrounding source, 12 mutations against the full suite, real-socket timing measurements and a line-by-line comparison against ArduPilot master and `Plane-4.7.1`, plus an independent `codex exec` pass given only this PR's number and the findings. Codex: 8 confirmed, 5 adjusted, 0 refuted, 0 new. The adjustments that change what you're being told are the terrain residual, the fixture-data correction in N2, and the wx-blocked counts.

_(My previous AI comment could not be marked deprecated in place — it was posted under a different account and the edit is refused. It is superseded by this one.)_


</details>

AP-Review:
**Automated review note — AI-generated (Claude), cross-checked by an independent Codex pass against the live diff.** Please sanity-check before acting.

Re-reviewed at head `8a70115eb8` (previously `b669fe9444`); my earlier AI comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_18_1407/devcall_pr_reviews.html#prMAVProxy-1753

**All three open notes from my last comment are closed, and the two that needed more than a docstring were verified by execution. Nothing new from either pass. Verdict: APPROVE.**

### Resolved since last round

- **N1 — the QRTL ramp is now labelled as master's** (`plane_track.py:540`). The docstring gained exactly what was asked for. I re-checked the claim rather than the wording: `git tag --contains` returns no `Plane-4.*` tag for either `bd9c0285fc` (2026-08-20) or `3dc0c3621b` (2026-08-22), so both really are master-only, and the description of 4.7's behaviour matches. That ~100 m divergence is now documented where the next reader will hit it.

- **N3 — `default_loiter_radius()` is shared, and the hole it left is closed** (`plane_track.py:1209`). Used by both the model (`:715`) and the test helper (`test_plane_track.py:1185`), so ArduPlane's `|WP_LOITER_RAD| <= 1 -> LOITER_RADIUS_DEFAULT` rule now lives in one place. **Mutation-verified:** stripping the fallback fails `test_a_loiter_to_altitude_with_no_radius_is_reached_at_the_default` at `:1086` — the new `for none in (0, 1, -1)` assertions — and nothing else, so the coverage is named and targeted (428 passed, 1 failed under the mutation).
  One thing I went looking for and did not find: the refactor inverts the comparison from `abs(x) > 1` to `abs(x) <= 1`, and those differ for NaN — old returns 60.0, new returns NaN. It is unreachable, because `parameter()` at `:129` filters NaN and falls back to the default, so `self.loiter_radius` cannot be NaN; I checked both forms over the reachable domain (0, ±1, ±0.5, ±60, 255, ±1e9 and either side of the boundary) with no mismatch. Mentioning it only because a later change could make it reachable.

- **N2 — the citations resolve *and* the autotest can now actually regenerate the recording.** This is the one where re-citing a live SHA alone would not have closed it, so I checked both halves. All ten fixtures were re-cited (`8d7ed31dfe`→`48c91a4c9e`, `f2c18f532f`→`bcb5e0c597`, `cd2319308b`→`3cf2e903d7`), all three resolve in both `ArduPilot/ardupilot` and `peterbarker/ardupilot`, and all three are commits of PR 34424. Only the `source` strings changed — the flight data is byte-identical, which is what "as rebased" should mean. And the substantive half: 34424's head `48c91a4c9e` is *"VTOLMissionItemTypes: return as QRTL from above RTL_ALTITUDE"* and does set `RTL_ALTITUDE: 60` for Q_RTL_MODE 3 (`quadplane.py:4137`), so the published autotest now reproduces the 60 m mode-3 recording. That was the actual gap, not the citation.
  Residual, for the record and not a request: 34424 is still a draft, so these SHAs are the head of a branch that may be rebased again and the citations would go stale exactly as `8d7ed31dfe` did. The `source` strings also name the PR and the test, which is the part that survives a rebase — about as good as a citation to unmerged work gets.

- **N5's overlong docstring line is fixed** (`test_plane_track.py:1165`).

### Closed, no change needed

N4 (`heights()` measures to the 3D-nearest point — right for a spiral, recorded as an accepted limitation) and N5 proper (`linear_interpolate()`'s polarity swap stays unreachable from both call sites, kept deliberately as a faithful port of `AP_Math.cpp`'s).

### Still open, all pre-existing and none introduced here

CI runs no tests at all — no workflow invokes pytest, so the 429 tests in this PR are never executed by CI. That remains the single highest-value thing that could change about this repository, and it is not this PR's job. Also `plane_track.py:406` (unresolved terrain still leaves the path silently at the aircraft's altitude with the style still `flown`) and `test_a_loiter_is_left_lined_up_on_what_comes_next` (still asserts convergence rather than exit heading).

### What I checked rather than assumed

The delta is exactly two commits, appended rather than force-pushed (`ahead_by 2, behind_by 0`), so the previous review's conclusions about everything else carry over intact. Suite at this head: **429 passed, 13 skipped, 20 subtests** on Python 3.14.4 / pymavlink 2.4.49 — identical to the previous head, as a docstring-and-refactor round should be; the independent pass got the same three numbers separately. Fixture `source` fields were compared by parsing the JSON at both heads rather than by reading the diff, since these are single-line files where a diff shows nothing useful.

Reviewed by: Claude + Codex (cross-checked).

PR #1755, opened 2026-09-16
mavproxy_link.py: cope with new MAV_TYPEs, future-proof against renames

peterbarker:
this broke again on quadrotor rename, so make all of these "soft" dependencies on xml enumeration names.

Also adds a few things which should be treated as fixed wings


AP-Review:

Reviewed at head `cf417a9c99`.

Full report, with everything that was checked: https://uav.tridgell.net/DevCallReviews/2026_09_16_AIReview/devcall_pr_reviews.html#prMAVProxy-1755

**Verdict: COMMENT** — 1 bug, 1 issue below.

The mechanism is right — `getattr(..., None)` with a `continue`, no module-scope AttributeError, no name typos, and the fallback is *skip* rather than `0` (which would be `MAV_TYPE_GENERIC` and would misclassify every vehicle). But three of the five new entries are dead on every released pymavlink, and the "future-proof against renames" claim covers only 1 of 5 MAVProxy sites.

### Bugs

- `MAVProxy/modules/mavproxy_link.py:718` — **`VTOL_FIXEDROTOR`, `VTOL_TAILSITTER` and `VTOL_TILTWING` are listed only under their *post*-rename names, so they are a no-op on every currently released pymavlink.** 2.4.49 has these values as `MAV_TYPE_VTOL_RESERVED2/3/4` (22/23/24); the new names do not exist. The PR carefully lists old *and* new names for DUOROTOR/QUADROTOR but omits `VTOL_RESERVED2/3/4`. Proven by importing the PR head against 2.4.49: `mav_type_planes = [1, 16, 19, 20, 21]` — 22, 23 and 24 absent. Mutation: adding the three old names gives `[1, 16, 19, 20, 21, 22, 23, 24]`. Under a simulated post-rename pymavlink the PR's list already yields the full set, so the gap is purely the missing old names. Reachable in practice — `quadplane.cpp:4094-4097` returns `MAV_TYPE(mav_type.get())` for any user-set `Q_MAV_TYPE`, so 22/23/24 are emitted by real vehicles. Not a regression (they were not handled before either), but the fix as written does not take effect until the rename lands.

### Issues

- `MAVProxy/modules/lib/mp_util.py:876` — **Four sibling sites still use bare attribute access on the names ArduPilot/mavlink#521 removes**, so the "future-proof" title covers 1 of 5. These are in function bodies, so they raise at *call* time rather than import, once the rename merges. Verified for this one: under a simulated post-rename `mavutil.mavlink`, `vehicle_type_name()` raises `AttributeError: module … has no attribute 'MAV_TYPE_VTOL_DUOROTOR'`. It is called from `mavproxy_map/__init__.py:1194` inside `mavlink_packet()` on **every HEARTBEAT**, so `sim_vehicle.py --map` dies on the first heartbeat. Same pattern, unverified but identical, at `mavproxy_console.py:264` (reached per heartbeat via `add_new_vehicle()`, so `--console` breaks too), `mavproxy_swarm.py:23` and `:781`, `tools/mavflightview.py:187`, `mavproxy_chat/chat_openai.py:288`, and `tests/test_mission_geometry.py:185` (which breaks MAVProxy's own test suite).

_4 further minor notes (style, coverage, pre-existing items) are in the report rather than here, so they don't bury the above._

---
Independent cold pass: no findings, APPROVE. The primary read's BUG was proven by importing the head against the real released pymavlink 2.4.49 and by a mutation, so it is retained.

Reviewed by: Claude (full read of the diff, thread and surrounding source) + an independent Codex cold pass that was not shown these findings. Findings marked _both reviewers_ were reproduced independently twice.


</details>

AP-Review:

Re-reviewed at head `5b59da7358` (previously `cf417a9c99`); my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_17_0037/devcall_pr_reviews.html#prMAVProxy-1755

---

**Verdict: APPROVE** — no blockers. Both previous findings are fully fixed, and the fix is better than what I asked for: the tolerant lookup is now centralised in one helper rather than repeated at each site.

### Previous round — both resolved

| Finding | Status |
|---|---|
| `VTOL_FIXEDROTOR`/`VTOL_TAILSITTER`/`VTOL_TILTWING` listed only under post-rename names, so dead on every released pymavlink | **RESOLVED** — `11cc0665` adds `VTOL_RESERVED2/3/4` at `mp_util.py:218-220`. Re-ran the probe against released pymavlink **2.4.49**: `plane_mav_types()` = `[1,16,21,19,20,22,23,24]`, so 22/23/24 are now present |
| Four sibling sites still used bare attribute access; "future-proof" covered 1 of 5 | **RESOLVED** — `5b59da73` centralises it in `mp_util.plane_mav_types()` (`:235`) and converts all of them: `mp_util.py:913`, `mavproxy_console.py:263`, `mavproxy_swarm.py:22`/`:777`, `mavflightview.py:186`, `chat_openai.py:288`, `tests/test_mission_geometry.py:185`. A tree-wide grep for `MAV_TYPE_VTOL\|MAV_TYPE_FLAPPING` now returns **zero** hits outside `mp_util.py` |

**The released-pymavlink check, since that is what this PR is about.** Fresh venv, `pip install pymavlink` → 2.4.49, then a per-name `getattr` probe of every name in `PLANE_MAV_TYPE_NAMES`: present — `FIXED_WING` 1, `FLAPPING_WING` 16, `VTOL_TILTROTOR` 21, `VTOL_DUOROTOR` 19, `VTOL_QUADROTOR` 20, `VTOL_RESERVED2/3/4` 22/23/24; absent — `VTOL_TAILSITTER_DUOROTOR`, `VTOL_TAILSITTER_QUADROTOR`, `VTOL_FIXEDROTOR`, `VTOL_TAILSITTER`, `VTOL_TILTWING`. `plane_mav_types()` gives `{1,16,19,20,21,22,23,24}` on 2.4.49 **and the identical set under a simulated post-rename pymavlink**, with no import failure and no `AttributeError` at import or call time. So the "future-proof against renames" claim is substantiated in both directions.

I also checked that `mavproxy_link.py:712` is a class-body assignment evaluated at import — but it resolves through `getattr(..., None)`, so it is safe; that upstream `minimal.xml` confirms the rename preserves the numbers exactly, so the old+new union is value-identical; and that a missing name is *skipped* rather than mapped to 0, so an unknown MAV_TYPE still falls through to "unknown" rather than being mislabelled. Recomputation costs 1.22 µs/call. CI at this head: 2/2 green.

### Notes — none blocking

- `mp_util.py:207` — `PLANE_MAV_TYPE_NAMES` adds `FLAPPING_WING` (16) but not `KITE` (17). `ArduPlane/quadplane.cpp:257` documents both as `Q_MAV_TYPE` values and `:4095-4097` returns `MAV_TYPE(mav_type.get())` unvalidated, so a `Q_MAV_TYPE=17` plane still classifies as unknown everywhere. Same reasoning that justified 16 — your call whether a kite is "flown as a plane".
- **Pre-existing, not yours** — `mavproxy_link.py:700`: the MAV_TYPE filter in `heartbeat_is_from_autopilot()` is dead code. The loop tests whether each *name* exists in `mavutil.mavlink`, never whether `m.type` is one of them, so it breaks on the first name that exists and always returns `True`. Executed at head against 2.4.49: `True` for GCS, CAMERA, GIMBAL and FIXED_WING alike. From `d5adb8a3` (2023-07-03) and already on master. Flagging it because it is the same "tolerate a missing name" pattern five lines above the line you changed, and your new helper is exactly the right fix — `if m.type in mav_types_from_names(...)`. Worth a separate commit.
- **Pre-existing, not yours** — `mp_util.py:275`: `mission_circle_radius()` references `MAV_CMD_DO_ORBIT`, which is not in released pymavlink 2.4.49 in either dialect, so any `pip install MAVProxy` user drawing a mission on the map hits `AttributeError` via `mp_slipmap_util.py:59`. Confirmed the same 6 test failures on merge-base `13b1fb1c`, so it is from `4ee840da` (2026-08-26), not this PR. Same class of bug, same file.

PR #1756, opened 2026-09-16
map: recover corrupt SRTM tiles and guard ROI altitude

tridge:
Cached HTTP error pages left terrain elevation permanently unavailable, suppressing camera footprints and passing None to the ROI MAVLink command. Retry invalid tiles and validate downloads before atomically replacing the cache, so readers cannot observe incomplete or invalid terrain data.

Reject ROI commands without a finite elevation and retain the previous ROI unless the command is sent successfully, with regression tests for cache recovery and ROI handling.


AP-Review:

Reviewed at head `2002ab54c0`. Full report: https://uav.tridgell.net/DevCallReviews/2026_09_17_AIReview/devcall_pr_reviews.html#prMAVProxy-1756

**Verdict: COMMENT — no blockers.** Both fixes are correct and the design is right: the write-to-temp → validate → `os.replace()` sequence is genuinely atomic, and the unique temp names make a shared `~/.tilecache` safe between processes. I verified that by running two concurrent writer processes against the same tile — 20 valid reads, 0 leftovers, both exited 0. I also ran the new tests (9 passed, 11 subtests) and mutation-tested each behavioural change: 5 of the 6 mutations are caught by the new tests.

Two things worth a conscious decision, then some minor items.

### Worth deciding

- **A persistently-bad tile is re-downloaded on every call, with no backoff** — `MAVProxy/modules/lib/srtm.py:308`. The new `except InvalidTileError: pass` falls through to the download branch. With a bad cached tile and a server returning HTTP 200 + garbage, **20 `getTile()` calls produced 20 download attempts**; the `is_alive()` guard only serialises the fetches, it doesn't stop them. This is reachable at rate: `mp_slipmap_ui.py:736` calls `update_position()` from `on_mouse`, which calls `GetElevation()` at `:515`, so hovering the cursor over an affected tile drives back-to-back downloads for as long as the mouse is there. Pre-PR this case was silent (cached garbage → `return 0`, zero traffic). The independent second pass flagged this too, separately. It's a fair trade for recoverability — but if you want to cap it, a per-process set of filenames whose *download* failed validation, skipped for the rest of the session, is about four lines.

- **The `except Exception` around the ZIP read got wider, so unrelated errors now read as tile corruption** — `MAVProxy/modules/lib/srtm.py:368`. Pre-PR only the `ZipFile()` constructor was inside the `try`; `namelist()` and `read()` were outside it. Confirmed behaviourally on both checkouts with the same valid tile and an injected `MemoryError`: at the merge base it propagates, at this head it becomes `InvalidTileError`. **No data is lost** — I checked specifically, the cached file is neither deleted nor corrupted (sha256 unchanged), because `downloadTile()` validates before `os.replace()`; the cost is one redundant ~3 KB download. `MemoryError` itself is unreachable in practice (I measured the decompressed SRTM3 payload at 2.75 MiB), so the realistic case is an `OSError`/`EOFError` from a disk read error being relabelled as corruption. `except (zipfile.BadZipFile, EOFError)` would keep the intent and let the rest propagate.

### Minor

- `MAVProxy/modules/mavproxy_map/__init__.py:1028` and `:1109` — `cmd_set_home` and `cmd_set_origin` are the same right-click menu family and still have the exact bug this PR fixes: unguarded `self.module('terrain')`, and unguarded `alt` from `GetElevation`, which returns `None` when the tile is unavailable. `cmd_set_origin` then does `int(alt*1000)` → `TypeError`. Out of scope here, but the same two-line guard would close the family.
- `srtm.py:335-341` — the temp file uses `delete=False`, so a SIGKILL between creation and `os.replace` leaves a `*.tmp` in `~/.tilecache/SRTM3` forever. Nothing ever enumerates that directory, so it's inert but unbounded; a `*.tmp` sweep in `SRTMDownloader.__init__` is cheap.
- `srtm.py:350` — `os.unlink(tmpname)` in the `finally:` is unprotected, so an `OSError` there (read-only cache dir, Windows sharing violation) escapes and kills the child with a traceback instead of the existing "SRTM Download failed" message. The filelist temp file at `:249-253` already uses the guarded style.
- `srtm.py:372` — the new `len(data) != 2 * self.size * self.size` clause is the one change the new tests don't cover: reverting it leaves the suite green. Worth keeping though — with an odd-length payload `int(math.sqrt(len/2))` still yields 1201, so the old code reached `array.array('h', data)` and raised `ValueError`, which isn't `InvalidTileError` and escaped into the map module. A one-line subtest would lock it in.
- Neither `python-cleanliness.yml` (flake8 only) nor `windows_build.yml` runs pytest, so `tests/test_srtm.py`, `tests/test_map_roi.py` and the 16 pre-existing files in `tests/` never run in CI. Good tests that will silently rot.
- `srtm.py:82` — `import tempfile` inside `SRTMDownloader.__init__` is now redundant with the new module-scope import at line 24.

### Checked and clear

No module-scope MAVLink constants are added, so there's no released-pymavlink `AttributeError` exposure — the `MAV_FRAME_GLOBAL` / `MAV_CMD_DO_SET_ROI_LOCATION` references in the diff are pre-existing and inside `cmd_set_roi`. The ROI guard can't reject a legitimate value: `GetElevation` is metres AMSL and returns `None` only when the tile is unavailable, SRTM voids return `-1` not `None`, `SRTMOceanTile` returns `0`, and the new test covers `0.0` and `-15.0` passing. CI is green (2/2).

PR #1757, opened 2026-09-17
camera: add multiple-camera controls, graphs, projections and ROI

tridge:
Camera controls currently share selection and projection state, which makes it easy to operate or display the wrong camera when multiple MAVLink cameras are present. This adds separate Camera, Camera2, etc. menus, ordered by MAVLink address, with controls, video actions, projections and graphs bound to the corresponding camera.

- Add camera-specific projection polygons and a Toggle Projection action, using ArduPilot's mount mapping for forwarded gimbal attitude.
- Add live graphs for gimbal attitude/rates/status, camera zoom/focus/mode, thermal maximum temperature, terrain height and battery voltage. Request thermal-range telemetry from capable cameras.
- Add map ROI choices for individual cameras and ROI All. Send geographic targets to the flight controller with the selected mount ID; MAVProxy does not calculate or stream ROI tracking angles. Angle/rate, centre and retract controls also go through the FC's gimbal manager.
- Serialize ROI requests per FC because its acknowledgements do not identify the mount. An ACK now applies to the outstanding request only; cancelled requests retain their ACK association, and retries use a quiet interval to drain delayed replies before the next request.
- Delegate photo capture, recording, zoom and focus to the FC when its `CAMn_TYPE` and `CAMn_COMPID` parameters identify a unique camera slot. Other operations retain the camera component address and use the existing MAVLink link for routing.
- Improve live-video reconnection, custom-settings refresh and camera/gimbal labels. Include compatibility with released pymavlink versions that lack newer enum entries.

The module no longer sends direct gimbal controls or modifies `MNTn_TARG_RATE`. This parameter must remain nonzero for ArduPilot to forward targets; if an older version left it at zero, restore it manually, normally to 10 Hz. Multi-mount ROI requires FC firmware that honours the ROI mount selector. Related FC support: [ArduPilot/ardupilot#34276](https://github.com/ArduPilot/ardupilot/pull/34276).

Validation: 107 camera-related tests passed, including mixed ROI acknowledgements, retries, cancellation, replacement, timeout recovery and replies addressed to another GCS. Live testing with an ArduPlane SITL connected to simulated MT11 and ZR10 cameras verified individual ROI, serialized ROI All and photo capture through the FC. Both mount target rates remained at 10 Hz. These final routing changes have been tested in SITL, not yet on physical cameras.


AP-Review:
> **Deprecated &mdash; see below for the updated review.**

<details><summary>Previous review (2026-09-17)</summary>

**Automated review note &mdash; AI-generated (Claude), cross-checked by an independent Codex pass against the live diff.** Please sanity-check before acting.

Full report, including everything that was checked and found clean: https://uav.tridgell.net/DevCallReviews/2026_09_17_AIReview/devcall_pr_reviews.html#prMAVProxy-1757

---

### REQUEST CHANGES

Reviewed at head `601ed85552`.

First AI review on this PR. It is a large, carefully-structured change and most of it holds up well: per-camera state is properly keyed on `(system_id, component_id)` with no index-contiguity assumption, menu and layer numbering derive from a `sorted()` keyset so they are discovery-order independent, every wx and matplotlib import is function-local and guarded, and the ArduPilot mount mapping (154->MNT1, 171->MNT2, 172-175->MNT3-6) matches `AP_Mount_MAVLink.cpp:84`. Two things block.

**Blocking**

- **`MAVProxy/modules/mavproxy_camera/roi.py:128` &mdash;
  `GIMBAL_DEVICE_CAP_FLAGS_CAN_POINT_LOCATION_GLOBAL` does not exist in released pymavlink, so multi-gimbal direct ROI
  dies with an uncaught `AttributeError`.**
  `pip install pymavlink` (2.4.49, current PyPI latest) has only the `GIMBAL_MANAGER_*` spelling; the `GIMBAL_DEVICE_*`
  mirror was added to `common.xml` in 2026-03 and is in pymavlink **git** only &mdash; which still self-reports
  `2.4.49`, so `setup.py`'s `pymavlink>=2.4.38` cannot express the requirement.
  Reproduced at runtime, not just in tests: with 2 cameras / 2 gimbals plus an ArduPilot heartbeat,
  `camera for 1:101 roi` raises from `roi.py:128`. `cmd_camera` only catches `(TypeError, ValueError)`, so it escapes to
  `mavproxy.py:832` and the user gets an opaque `ERROR in command [...]` with the feature dead. The same constant at
  `tests/test_mavproxy_camera.py:241` makes **15 of 55** camera tests fail on released pymavlink (all 55 pass against
  git). CI never sees this: `python-cleanliness.yml` runs flake8 only, and `windows_build.yml` does
  `pip install -U pymavlink` from PyPI &mdash; so **the shipped Windows installer would contain the broken path**.
  The obvious fix follows your own precedent from `13b1fb1c` ("map: tolerate missing arc waypoint enum in pymavlink"):
  a module-level `getattr(mavutil.mavlink, "GIMBAL_DEVICE_CAP_FLAGS_CAN_POINT_LOCATION_GLOBAL", 1 << 17)`.
  **But that alone is not enough** &mdash; the second pass established that released pymavlink also lacks `cap_flags2`,
  so a capable gimbal's packet encoded with git pymavlink and decoded with the release loses that field, and adding only
  the constant then yields "does not support geographic ROI". The missing field needs handling too.

- **`roi.py:285` &mdash; manager-mode ROI puts the mount number in `param1`, which ArduPilot ignores, so a non-primary
  mount's ROI silently moves MNT1.**
  `send_location()` emits `COMMAND_INT(MAV_CMD_DO_SET_ROI_LOCATION, param1=target.mount)`, which is correct per the
  MAVLink spec. But `GCS_MAVLINK::handle_command_do_set_roi()` (`GCS_Common.cpp:5602`) documents `param1` as "Region of
  interest mode (not used)" and calls `mount->set_roi_target()`, which routes to `_primary`. `DO_SET_ROI_NONE` takes the
  same path, and the paired ArduPilot#34276 (head re-fetched at review time, `47c99eb6e3`) does not change it.
  Reachable with no misconfiguration: one camera advertising `gimbal_device_id=171` (MNT2) with MNT1 a servo mount
  leaves only one MAVLink gimbal device visible, so `control_mode()` stays `"manager"`. Reproduced: `camera for 1:100
  roi` emits `cmd=195 param1=2`, ArduPilot replies `ACCEPTED`, MAVProxy reports success, and MNT1 moves. The
  `_manager_id()` error text at `__init__.py:730` tells users to configure a value the manager path cannot act on.
  Note that forcing device mode alone is not sufficient, because stream suspension currently requires multiple
  discovered gimbals &mdash; so this needs either the firmware side to honour `param1`, or a rethink of the mode selection.

**Notes**

- **`.github/workflows/python-cleanliness.yml` &mdash; no workflow runs `tests/`**, so the 810 new test lines gate
  nothing. That is why the first finding is invisible to CI, and it will stay invisible for the next regression of this
  shape. A `pytest tests/` job would be cheap: the whole suite runs in 26 s here (it needs
  `numpy opencv-python-headless defusedxml`, plus xvfb and wxPython for 12 of the camera tests).
- **`roi.py:138` &mdash; `MNTn_TARG_RATE` exists in no released ArduPilot**; it comes from #34276. The degraded path is
  correct and self-explaining (retry 3x then print a cancellation message without touching parameters), so this is merge
  ordering rather than a defect &mdash; just worth stating in the PR description.
- **`mavproxy_camera/__init__.py:146` &mdash; completion is half-covered**: `projection` gained a `toggle` argument with
  no completion entry, and `for` has no `SYSID:COMPID` completion despite being the primary new addressing mechanism.
  (On Linux `rline.py:280` raises a pre-existing `IndexError` for these anyway, so the practical difference is
  Windows-only.)
- **`roi.py:33` &mdash; `original: float = None`** should be `Optional[float]`; cosmetic, no runtime effect.

**Checked and clean, reported either way:** the module-scope MAVLink constant trap is **clean** &mdash; an AST scan of
all five new/changed files for uppercase attribute access at module, class-body, decorator or default-argument position
found **zero** occurrences, and `import MAVProxy.modules.mavproxy_camera` succeeds against released pymavlink 2.4.49
headless with no wxPython installed. (The first finding is the *late-bound* variant, not an import failure.)
`camera_projection.py` is byte-identical between base and head, and I re-derived the maths anyway: at 100 m AGL with an
88 deg HFOV, nadir centres the footprint on the vehicle's own position; pitch -45 deg gives a centroid 159 m out against
a theoretical near/far of 32 m and 308 m; yaw 0/90/180/270 rotates the bearing correctly, so the cos(latitude) scaling is
right; and pitch +45 deg and +89 deg return `None` rather than a bogus ground point. No unbounded growth over a long
session. No direct wx call from the main thread and nothing unpicklable crossing the pipe. Every new command exercised
with 0 and 2 cameras: all failures are caught `ValueError`s with readable messages, and state is restored in every case
including the raising ones. `scripts/run_flake8.py MAVProxy` (the actual CI gate) is clean, and the full suite is
230 passed / 10 skipped against git pymavlink.

*One thing I considered and dropped:* the unguarded `self.module('camera').roi.menu_items()` call at the top of
`mavproxy_map/__init__.py:971`. Neither pass could construct a reachable exception from valid camera state, so it stays
a hardening suggestion rather than a finding.

</details>

AP-Review:

<details><summary>Previous review (2026-09-17)</summary>

Re-reviewed at head `5fce96fa2c`; my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_17_AIReview/devcall_pr_reviews.html#prMAVProxy-1757

## Verdict: REQUEST CHANGES — blocking only on the new CI job being red

**Both blocking findings from my last comment are fixed, and better than I asked for.**

- **The released-pymavlink problem.** `roi.py:11` now uses `getattr(mavutil.mavlink, "GIMBAL_DEVICE_CAP_FLAGS_CAN_POINT_LOCATION_GLOBAL", 1 << 17)`, and `roi.py:15-28` adds `gimbal_capabilities()`, which recovers `cap_flags2` from the raw MAVLink 2 frame when the installed dialect cannot decode it. I verified all three parts independently: the fallback value against `common.xml`; the byte offsets by deriving the 144-byte ordered wire layout; and functionally, by hand-building MAVLink 2 frames with correct CRCs and decoding them with **released pymavlink 2.4.49** — 131072 for a GLOBAL-capable gimbal, 65536 for LOCAL-only, 8192 for a truncated frame. The regression test at `tests/test_mavproxy_camera.py:398` builds the wire frame by hand, so it is dialect-independent, and the new CI job pins `pymavlink==2.4.49`. That is a better fix than the one I suggested.
- **The manager-mode ROI problem.** `roi.py:81-91` returns `"device"` when the vehicle is ArduPilot and `_manager_id() > 1`. I re-verified the firmware premise at ArduPilot master: `handle_command_do_set_roi()` still documents `param1` as "not used" and discards it. Both branches reproduced — MNT1 sends a correct `COMMAND_INT`, MNT2 refuses with zero commands sent. No more silently moving MNT1.

You also fixed something I had not reported: `mavproxy_map/__init__.py:22` and `mp_util.py:275,302` now guard `MAV_CMD_DO_ORBIT`. That was a *live* latent bug of exactly this class — the constant is absent from released pymavlink 2.4.49 and was previously dereferenced at module-load time in `MapModule.__init__`.

**Released-pymavlink import check: clean.** An AST scan of all six changed files for uppercase `mavutil.mavlink.*` access at module scope, class-body scope, decorator position or default-argument position found zero occurrences; both module-level constants use the safe `getattr` form. Both passes did this independently in separate venvs, each confirming `pymavlink.__file__` pointed inside the venv.

### BUG — the `tests` job this PR adds is red on this PR's own head

`.github/workflows/python-cleanliness.yml:40`. The job log reports `222 passed, 29 skipped, 3 errors`; all three errors are `tests/test_video_telemetry.py::RecordingTests` `setUpClass` (`:161`) failing with:

```
Error parsing bitstream filter sequence 'h264_metadata=sei_user_data=...'
```

under `ubuntu-22.04`'s ffmpeg 4.4.2. `test_video_telemetry.py` is **not touched by this PR** — this is pre-existing fragility that the new job newly exposes. But since the PR adds the job, merging makes `python-cleanliness` red on every push and PR for the whole repo, which defeats the purpose of adding it.

Suggested fix: have `RecordingTests.setUpClass` `skipTest` when that ffmpeg invocation fails, and/or move the job to a newer runner image. **Caveat worth stating:** neither pass established which ffmpeg version accepts that escaped bsf string — `sei_user_data` itself exists in 4.4 — so please do not treat a runner bump as a known-good fix without testing it.

### ISSUE — `camera graph tmax` silently plots nothing for an ArduPilot mount-attached camera

`MAVProxy/modules/mavproxy_camera/graphs.py:142`. The filter `if getattr(message, "camera_device_id", 0) != 0: continue` requires `camera_device_id == 0`, but the firmware shipping `CAMERA_THERMAL_RANGE` today is ArduPilot's `AP_Mount_Siyi`, which sends `camera_device_id = _instance + 1` (`AP_Mount_Siyi.cpp:1038`) — always ≥ 1, so every message is dropped. The window opens, prints a success line, and stays empty forever with no diagnostic.

Reproduced through the module's own harness: with the value ArduPilot actually sends, `add_values` is never called; with 0, it is. The filter is deliberate and correct for the documented MT11 deployment, but the module explicitly supports autopilot-proxied cameras, so this is reachable with no misconfiguration.

Fix: bind the window to the camera's *expected* device id — 0 when the camera has its own component, and `camera.information.gimbal_device_id` when it is advertised from the autopilot. That field is exactly the right key, since `AP_Camera_Backend::get_gimbal_device_id()` returns `mount_instance + 1`, matching `AP_Mount_Siyi`.

Also worth knowing: ArduPilot never sets `CAMERA_CAP_FLAGS_HAS_THERMAL_RANGE` (zero occurrences across `libraries/`), so the unforced request path never fires for ArduPilot and the forced path is the only one. That part is handled well.

### ISSUE — selected-camera shorthand graphs the wrong mount

`graphs.py:65`. Found by the Codex pass: with forwarded mount telemetry and default settings, `camera select 1:101; camera graph yawrate` plots **mount 1** even though camera `1:101` advertises mount 2, because `_manager_id()` only resolves camera associations inside an addressed context. Codex reproduced it through the harness and confirmed the explicit form `camera for 1:101 graph yawrate` correctly plots mount 2.

### Notes

- `__init__.py:303` — `_camera_key()` is `(system_id, component_id)`, so an ArduPilot vehicle with `CAM1_TYPE=Mount` and `CAM2_TYPE=Mount`, both advertised from component 1, collapses into one `CameraDevice`. None of the multi-camera machinery can then address them separately. Inherent to the key rather than a regression — and arguably the underlying reason for the `camera_device_id` filter above, which is why a real fix for that probably wants the device id in the camera key.
- `roi.py:146-149` — geographic ROI is unavailable for an ArduPilot non-primary *serial/servo* mount, since device mode needs a MAVLink gimbal component. This is strictly better than the old behaviour and the failure is honest; downgrading this to an optional README clarification on the Codex pass, since the behaviour is deliberate and documented. Still worth saying explicitly that secondary serial/servo mounts are unsupported until ArduPilot honours `param1`.
- `README.md:249-252` — "Current limitations" still says temperature extrema "remain unavailable here", contradicting the TMax section at `:171-178`.
- `__init__.py:489-491` — `SET_MESSAGE_INTERVAL` sends `param3=0, param4=0`, but per `common.xml` an autopilot-attached camera wants its device id in `param4`. Correcting my own earlier note on Codex's evidence: ArduPilot *rejects* a nonzero `param3` and ignores `param4` — it does not ignore both. Works today; not spec-conformant for other firmware.

**Checked clean:** `camera_projection.py` is byte-identical to the merge base, so the previous round's numerical re-derivation still stands. ROI encoding verified against `common.xml` (`MAV_FRAME_GLOBAL`, 1e7-scaled int32 lat/lon, float-metre altitude) and survives a wire round trip — and `param1` 1-6 is in fact spec-correct for non-MAVLink gimbals, with ArduPilot the outlier. Threading: `graphs.windows` is only mutated from the main loop, the same thread that dispatches `mavlink_packet` and `idle_task`, so the new indexed write has no race; graph history is bounded and `windows` is pruned in `idle()`. The map menu cannot accumulate duplicate "Set ROI" entries. `gimbal_capabilities()` degrades safely on a MAVLink 1 frame and on a short payload. Codex's focused regression run: 109 passed, 24 skipped.


</details>

AP-Review:

Re-reviewed at head `5d7ca3538c`; my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_20_0808/devcall_pr_reviews.html#prMAVProxy-1757

## Verdict: REQUEST CHANGES

Of the 7 findings in my last comment: **0 resolved, 5 still open, 1 superseded by the rework, 1 partly fixed.** The rework itself is substantial and mostly right — the FC-delegation plumbing, the ROI serialisation state machine and the new tests are good work — but three pieces of it do not do what the README says they do on any shipping ArduPilot, and one of them is a **regression against the head I reviewed last round**.

Since you squashed five commits into two and rebased, this was reviewed as a fresh read of the whole diff rather than an increment. `graphs.py` is byte-identical to the previous head, which is why both graph findings carry over unchanged.

### Previous round

| Previous finding | Status |
|---|---|
| BUG — the `tests` job this PR adds is red on its own head | **STILL OPEN** — re-raised below, now with the cause pinned |
| ISSUE — `graphs.py:142`, `camera graph tmax` plots nothing for an AP mount-attached camera | **STILL OPEN** |
| ISSUE — `graphs.py:65`, selected-camera shorthand graphs the wrong mount | **PARTLY FIXED** — `mount`, `roi` and `projection` now re-enter `_camera_context`; `graph` does not |
| NOTE — `_camera_key()` collapses two mount-attached cameras on component 1 | **STILL OPEN**, now with a concrete consequence |
| NOTE — `roi.py`, geographic ROI unavailable for an AP non-primary serial/servo mount | **SUPERSEDED** — the `"device"` path it described was deleted; see N2 |
| NOTE — README "Current limitations" contradicts the TMax section | **STILL OPEN** — `README.md:245-248` untouched this round |
| NOTE — `SET_MESSAGE_INTERVAL` sends `param3=0, param4=0` | **STILL OPEN** — no longer cosmetic; it is half of N4 |

---

### BUG — N2: ROI for a non-primary ArduPilot mount now silently moves the *primary* mount and reports success

`roi.py:199-201`. This is a regression against the head I reviewed last round. Then, `control_mode()` returned `"device"` when the FC was ArduPilot and the resolved mount was not primary, and I verified it by execution — *"MNT1 sends a correct COMMAND_INT, MNT2 refuses with zero commands sent."* At this head `control_mode()` and the whole `"device"` path are gone, and every ROI goes to the FC with the mount in `param1`.

Re-verified at ArduPilot master `368dc0c428`: `GCS_Common.cpp:5610` still says `// param1 : /* Region of interest mode (not used)*/`, and `:5494` calls `mount->set_roi_target(roi_loc)`, which is `AP_Mount.h:203` `set_roi_target(_primary, target_loc)`. The command moves `_primary` whatever param1 says, and returns `MAV_RESULT_ACCEPTED` — so MAVProxy prints success while the wrong gimbal slews.

The PR's own test pins this: `tests/test_mavproxy_camera.py:385` `test_lone_second_mount_roi_uses_manager` asserts param1=2 to `1:1`, which on today's firmware moves MNT1.

**And the stated prerequisite has no implementation anywhere.** `README.md:124` says this "requires FC firmware that honours the ROI mount selector" and the description points at ArduPilot/ardupilot#34276. I fetched that PR's diff: `handle_command_do_set_roi` and `set_roi_target` appear **zero** times in it. #34276 adds `CAMn_COMPID` and gimbal-manager work, but nothing that makes ArduPilot honour the ROI mount selector — so no currently-proposed firmware satisfies the README's requirement.

Suggested fix: restore a guard equivalent to the deleted `control_mode()` — refuse, naming the mount, when the FC is ArduPilot and the resolved mount is not primary — or gate multi-mount ROI behind an explicit opt-in until the firmware side exists. Failing honestly is what the previous head did, and it is much better than moving the wrong gimbal while printing success.

### BUG — N3: the console now renames the *autopilot's* vehicle-menu entry after the gimbal, for every ArduPilot user with a Siyi/Topotek/Viewpro/Xacti mount

`mavproxy_console.py:369`. `handle_component_information_anysource` was widened at `:880` to accept `CAMERA_INFORMATION`, and `update_component_model_name` applies the advertised vendor/model with `override=True`, which `set_component_name` (`:355-361`) now honours — reversing the previous first-writer-wins behaviour that let the heartbeat type stand.

But ArduPilot sends `CAMERA_INFORMATION` **from its own component 1**: `AP_Mount_Backend::send_camera_information()` (`AP_Mount_Backend.cpp:419-433`), with `AP_Mount_Siyi.h:81,83` giving `has_camera_information() { return true; }` and `strncpy(buf, "Siyi", buflen)`. `AP_Mount_Topotek.h:88`, `AP_Mount_Viewpro.h:84` and `AP_Mount_Xacti.h:71` do the same.

Reproduced against the real `ConsoleModule`: after the heartbeat the menu reads `SysID 1: Copter`; after the first `CAMERA_INFORMATION` it becomes `SysID 1: Siyi-ZR10 (Copter)`. With two mounts of different vendors the label flip-flops and `console.set_menu()` is rebuilt on every message, because `component_model` is keyed only by `(sysid, compid)` and both instances arrive from component 1. `vehicle_name_by_sysid` is untouched so the damage is confined to the menu label — but this is shared console code and it hits users who never load the camera module.

Fix: only adopt the advertised model when the component is actually a camera or gimbal — skip when its heartbeat type is neither `MAV_TYPE_CAMERA` nor `MAV_TYPE_GIMBAL`, or simply skip `compid == MAV_COMP_ID_AUTOPILOT1`.

### ISSUE — N9: FC delegation never engages on any shipping ArduPilot

`__init__.py:669-671`. `params.get("CAM%u_COMPID" % slot)` returns `None` when the parameter is absent, so the `if compid == 0:` default-mapping branch is skipped, `compid == camera.component_id` is never true, and `_camera_manager_slot()` returns `None`.

`CAMn_COMPID` does not exist in ArduPilot master — `git grep COMPID master -- libraries/AP_Camera` returns nothing. It is added by the still-open #34276. On released firmware the component id is hard-coded to `MIN(MAV_COMP_ID_CAMERA + _instance, MAV_COMP_ID_CAMERA6)` (`AP_Camera_MAVLinkCamV2.cpp:208`), which is exactly what your `:671` default computes — so the default is right and simply never runs.

The fallback is *safe* (commands keep the camera address, i.e. pre-PR behaviour), so this is a dead feature rather than a misroute. `README.md:267-269` documents a rule that never applies. Fix:

```python
compid = params.get("CAM%u_COMPID" % slot, 0) or (
    mavutil.mavlink.MAV_COMP_ID_CAMERA + slot - 1)
```

`test_camera_slot_unknown_wrong_vehicle_or_ambiguous_stays_addressed` does have a `{"CAM1_TYPE": 6}` case, but for camera 101 where the default mapping gives 100 anyway, so it cannot tell "missing param" from "no match" — worth a case with only `CAMn_TYPE` present.

### ISSUE — N1: the new `tests` job is red on this head, and merging makes `python-cleanliness` red repo-wide. Root cause now established.

`.github/workflows/python-cleanliness.yml:28`. This is the finding I could not fully close last round, so here is the missing half.

The failure is not a quoting mistake. ffmpeg reports **`Bitstream filter not found`**, because of the bsf-list parser: `libavcodec/bsf.c` at tag `n4.4` uses `av_strtok(buf, ",", &saveptr)` — a plain strtok with no escape or quote handling — so the JSON payload's `\,` escapes are ignored, the string is split at every comma, and the second fragment is looked up as a filter name. From `n5.0` onwards it is `av_get_token(&str, ",")`, which honours `\,` and `'…'`.

So it is purely an ffmpeg-version issue. jammy ships `7:4.4.2` (broken), noble ships `7:6.1.1` (fixed). I ran your exact `setUpClass` command under **ffmpeg 8.0.1** — `rc=0`, a 29962-byte mp4 — and the whole suite at this head passes locally: **275 passed, 5 skipped, 64 subtests**.

Recommended, both parts:

1. Guard `RecordingTests.setUpClass` (`tests/test_video_telemetry.py:154-163`) so no ffmpeg build can red the repo — wrap the `subprocess.run(..., check=True)` and `raise unittest.SkipTest(...)` on `CalledProcessError`. Note `tearDownClass` does not run when `setUpClass` raises, so use `cls.addClassCleanup(cls.temp.cleanup)`.
2. Move the job to `ubuntu-24.04` so the coverage is actually exercised (`python3-wxgtk4.0` is there at 4.2.1).

Do **not** try to re-escape the bsf string: on 4.4 no escaping works, the split is unconditional. And the rest of the suite on 24.04 (python 3.12, wx 4.2.1) is **untested** — treat step 2 as needing a trial run, not as known-good.

### ISSUE — N4: `camera graph tmax` is broken at both ends for an ArduPilot mount-attached camera

Re-raised, now with the request side shown to be part of it.

**Receive side.** `graphs.py:142` requires `camera_device_id == 0`, but `AP_Mount_Siyi.cpp:1038` sends `_instance + 1` — always ≥ 1. Reproduced against released pymavlink 2.4.49, feeding real packed-and-decoded frames into `CameraGraphs.packet()`:

```
camera_device_id=1 (AP instance 0)  -> add_values called 0 times
camera_device_id=2 (AP instance 1)  -> add_values called 0 times
camera_device_id=0 (standalone cam) -> add_values called 1 time
```

Window opens, success line prints, nothing ever plots, no diagnostic.

**Request side.** `common.xml`'s `CAMERA_THERMAL_RANGE` description: *"param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras"*. `__init__.py:489` sends `(message_id, interval, 0, 0)`. Note the constraint on any fix: ArduPilot **rejects** a nonzero `param3` outright (`GCS_Common.cpp:3204`, `MAV_RESULT_DENIED`) while ignoring `param4` — so setting param4 is free on ArduPilot and correct elsewhere, but setting param3 would break it.

The id is available: `CAMERA_INFORMATION.gimbal_device_id` is `mount_instance + 1` (`AP_Camera_Backend.cpp:122-134`), matching what `AP_Mount_Siyi` puts in `camera_device_id`. Bind the window to the camera's expected device id (0 when it has its own component, `camera.information.gimbal_device_id` when advertised from the autopilot) and put the same value in `param4`. A one-shot diagnostic when a message from the bound source is dropped by that filter would stop the window ever being silently empty.

### ISSUE — N5: `camera select` + `camera graph` still binds to the wrong mount

`__init__.py:789-790`. The other three call sites were fixed this round; `graph` was missed. `_manager_id()` opens with `camera = self.command_camera`, and `cmd_select` (`:637-653`) sets `selected_camera` but never `command_camera`. `graphs.open()` gets its camera from `_selected_camera()` — which *does* honour the selection — then calls `_manager_id()` with no context, so it falls through to the global `manager_gimbal_id` at `:821` (0), which `graphs.py:65` turns into mount 1:

```
camera select 1:101 ; camera graph yawrate     -> "Yaw rate (1:1 mount 1)"   WRONG
camera for 1:101 graph yawrate                 -> "Yaw rate (1:1 mount 2)"   correct
camera select 1:101 ; camera mount angle 0 90  -> param7=2                   correct (fixed this round)
```

The menus emit `camera for X:Y graph ...`, so the GUI is unaffected; it is the CLI combination that misbinds.

Suggested fix — put the fallback in `_manager_id()` itself so every caller benefits: when `command_camera is None`, re-enter through `_camera_context(self._selected_camera(required=False))`. **Do not** instead wrap the whole `graph` command in `_camera_context`: `CameraGraphs.open()` keys `graph close` off `command_camera` (`graphs.py:42-44`), so wrapping would quietly turn a bare `camera graph close` into "close only this camera's graphs", contradicting `README.md:76`.

### ISSUE — N7: the ACK quiet interval is not armed after a first-attempt success

`roi.py:127`. `_retire()` arms `ready_at` only `if target.attempts > 1 or target.cancelled`, so the ordinary one-command-one-ACK path arms nothing and the next queued mount immediately owns the manager's ACK key. Over `camera roi all`: mount 1 ACCEPTED (`attempts == 1`, `ready_at` stays `{}`), `idle()` sends mount 2, then a duplicate of mount 1's ACK marks mount 2 `confirmed=True`; mount 2's real ACK then arrives with `inflight` empty and is dropped without a message. A duplicate carrying a *failure* result instead tears down mount 2's accepted ROI.

Being straight about the precondition: this needs a duplicated or late `COMMAND_ACK`, and I have **not** shown that happens on a single-link ArduPilot setup — ArduPilot acks once, on the receiving channel. It is realistic on fan-out links (mavlink-router, two telemetry paths into one MAVProxy). `README.md:129-131` already promises the drain "after retries, cancellation or timeout"; the gap is that success gets none. Arming `ready_at` unconditionally in `_retire()` closes it, at up to 1 s extra between mounts in `camera roi all` — a shorter interval for the success case would do if that matters.

### ISSUE — N8: an ACK caused by MAVProxy's *own* map module can be consumed by an inflight camera ROI

`roi.py:152-153` keys only on `(srcSystem, srcComponent)`; the other filter (`:147-151`) only rejects ACKs addressed to a *different* GCS. But `mavproxy_map/__init__.py:1089-1092` sends `MAV_CMD_DO_SET_ROI_LOCATION` to `(settings.target_system, settings.target_component)` over the same master from the same sysid/compid, so its ACK is indistinguishable. With a camera ROI inflight, a map "Set ROI" click can mark the camera's target confirmed (its own ACK then dropped), or on a failure result trigger a spurious `DO_SET_ROI_NONE` that tears down an ROI the FC accepted.

The window is the inflight interval — about 1 s, longer under retries or a multi-mount `camera roi all` — and it needs concurrent use of both menus, so it is a race rather than an everyday failure. `README.md:116` notes the map keeps its own Set ROI with zero or one camera, which is exactly where both entries coexist. Either match the ACK against the `(lat, lon, alt)` sent, or suppress the map's Set ROI while the camera module owns ROI for that vehicle.

### Notes

- `__init__.py:301-302` — the `_camera_key()` collapse now has a concrete consequence: two `CAMn_TYPE=4` cameras on component 1 give one menu, and `camera photo` goes out with `param1 == 0`, which ArduPilot treats as *every* backend. ArduPilot supplies the discriminator — `AP_Camera_Backend.cpp:244-245`, `lens_id` is the instance number. Keying on `(sysid, compid, lens_id)` for autopilot-sourced information would fix this, make the slot derivable as `lens_id + 1` for every `CAMn_TYPE` rather than only type 6, and supply the device id N4 needs.
- `roi.py:10-11` — `LOCATION_GLOBAL` is now dead code; its only consumer was the deleted `"device"` branch. `gimbal_capabilities()` below it *is* still live via `__init__.py:1275` and its hand-built-frame test still passes against 2.4.49, so only the constant should go.
- `roi.py:44` vs `:201` — `ROITarget.endpoint` normalises with `self.mount or 1`, but `send_location` sends `target.mount` raw, so an unresolved mount is booked as 1 and sent as 0 — which `common.xml` defines as "all gimbal device components". Moot on ArduPilot; on a spec-compliant FC two cameras, one mapped and one not, collide on one endpoint while sending different selectors.
- `roi.py:106-110` — `camera roi all` prints one success line per camera and then collapses duplicates, so two cameras on one mount produce two lines and one command. The collapse is intended; the output is what misleads.
- `parameters.py:404-409` — the Custom Settings dialog re-reads every known parameter every 5 s. A 30-parameter definition is ~30 `PARAM_EXT_REQUEST_READ` + ~30 `PARAM_EXT_VALUE` every 5 s, roughly 1.2 kB/s from the 2.4.49 wire sizes — a meaningful share of a 57600 baud link. *That figure is arithmetic from the message sizes, not a measurement on a real link.*
- `__init__.py:130`/`:422` — `multi_vehicle=True` is new, but every transmission still goes out `self.master`, which has no `target_sysid`. With two links and cameras on a second vehicle, `camera for 42:100 photo` could go out on the link that does not reach system 42. `README.md:265` may make this deliberate; if so it is worth saying in the multi-vehicle section. *(Code reading only — not reproduced.)*
- `mp_image.py` — three small ones: `:771`'s comment overstates the timeout pair (OpenCV's GStreamer backend applies constructor params only after the blocking open, so `rtspsrc tcp-timeout` is what actually covers it); `:779-780` prints `VideoCapture failed` once per retry forever for an unreachable camera, since `video_view.py` always passes `reconnect=True`; `:777`/`:800` leave `self.vcap` write-only.
- CI/test coverage: `python-cleanliness.yml:36` omits `vtk` and `quantized_mesh_tile`, so 24 tests silently skip — 2 in `test_map_arc_compat.py` and 22 in `test_mission_geometry.py`, both files this PR modifies. `tests/test_mp_image_video.py` never asserts the reconnect capture parameters, so the timeout list at `mp_image.py:772-774` is unpinned. `tests/test_mavproxy_camera.py:123` passes `vendor_name=b"..."`, but a real `CAMERA_INFORMATION` decodes it as `list[int]` (it is `uint8_t[32]`) — `_text()` handles it, but `test_console_component_names.py:38` already parametrises all three types and is the pattern to copy.

### Dropped during cross-check, so you can ignore it

One pass reported that `camera projection` shares the `graph` mount-binding bug via `show_projection()` at `:1112`. **It does not** — `cmd_projection` enters `_camera_context` at `:1101` before calling it, and that is `show_projection`'s only caller; `_refresh_fov()` (`:1013-1017`) also wraps per camera. Not a finding.

### Checked and found clean

Released-pymavlink compatibility is intact and was re-tested rather than assumed: an AST scan of every changed `.py` for uppercase `mavutil.mavlink.*` at module, class-body, decorator or default-argument scope found hits only in `mavproxy_console.py:24-39`, which are byte-identical to the merge base and all present in 2.4.49; the only missing constant across 38 checked is the one with a `getattr` fallback, and `MAV_CMD_DO_ORBIT`/`MAV_CMD_NAV_ARC_WAYPOINT` are genuinely absent so those guards are load-bearing, with correct fallbacks (34, 36). Every `mp_image.py` signature change is an added keyword with a default and all three external call sites are fine; the `handle_gimbal_device_information_anysource` rename has no stale callers and the dispatch is explicit, not name-based. ROI timeout recovery exercised over six idle ticks — three attempts at 1 s, then `ROI_NONE` and a message naming the mount, with `inflight` and `targets` drained and no stuck state; `roi.py:97-105` resolves every endpoint before sending anything, so an ambiguous second mount aborts the batch with zero commands on the wire; the wire encoding matches `common.xml` and `location_from_command_t`. `_manager_id()`'s 154/171.. mapping matches `AP_Mount_MAVLink::find_gimbal()`, and the delegation selector slots match `AP_Camera.cpp` (param1 1-based camera id for capture, param3 1-based instance for zoom/focus). The claims about what the module stopped doing hold: no `gimbal_device_set_attitude_send` and no `MNTn_TARG_RATE` write remains anywhere. Threading is clean — the only threads are the definition-download workers, which just put to a queue. Test quality is good: 61 `def test_` against 26 `assert_called*`, with the bulk asserting on real `command_*_send` payloads rather than on mocks of the code under test. Suite at this head under 2.4.49: 275 passed, 5 skipped, 64 subtests; flake8 clean.

**Not reached:** nothing was tested on hardware or against a non-ArduPilot autopilot, so the param4 and multi-vehicle-link questions rest on specification and code reading; the `ubuntu-24.04` half of the CI fix was not trialled.

PR #1758, opened 2026-09-17
srtm: don't download 'filelist_python'

mr3screen:
https://terrain.ardupilot.org/SRTM3/ has a filelist_python entry, which is not a continent and results in a warning message:

    Failed to download /SRTM3/filelist_python : 'utf-8' codec can't decode byte 0x80

Fix taken from acato on github. <https://github.com/acato/drone_hass/blob/main/dev/sitl-setup.md#troubleshooting>

I did not test the change.


peterbarker:
@stephendade should that file even exist on the server?

PR #1759, opened 2026-09-19
Fixes for MAVLink 1.0 support

peterbarker:
Need to make sure messages are in the dialect

Fix method-on-None bug in waypoint module


gmorph (approved):
Tested on my system where this was failing and its now passing.  Thanks Peter!

